Issue #5, opened 2018-02-16
Replicate `astro airflow pause` from open


ryw:
Pauses containers, keeps data

andscoop:
Have also proposed proper naming of this in #18

PR #21, opened 2018-02-21
add astro airflow pause

### LINKED ISSUE #18: Proper naming of Pause, Kill, Stop Commands
I propose that we use two commands for these behaviors

`astro airflow stop` 
A non-destructive wrapper around [docker stop](https://docs.docker.com/engine/reference/commandline/stop/) which maintains mounted volumes. Similar to how some may view a pause command

`astro airflow kill`
A destructive tear down of containers. How `astro airflow stop` currently works at time of issue creation.

andscoop:
Closes #18 and #5

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -21,8 +21,10 @@ import (
 )
 
 const (
-	componentName   = "airflow"
-	deployTagPrefix = "cli-"
+	componentName     = "airflow"
+	deployTagPrefix   = "cli-"
+	dockerStateUp     = "Up"
+	dockerStateExited = "Exited"
 )
 
 // ComposeConfig is input data to docker compose yaml template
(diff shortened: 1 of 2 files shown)


airflow/docker.go
```diff
@@ -85,8 +85,8 @@ func generateConfig(projectName, airflowHome string) string {
 	return buff.String()
 }
 
-// createProjectFromContext creates project with yaml config as context
-func createProjectFromContext(projectName, airflowHome string) (project.APIProject, error) {
+// fetchDockerProject creates project with yaml config as context
+func fetchDockerProject(projectName, airflowHome string) (project.APIProject, error) {
```
andscoop: still not sure about the naming of this func

airflow/docker.go
```diff
@@ -107,30 +107,50 @@ func Start(airflowHome string) error {
...
+	// Fetch project containers
+	psInfo, err := project.Ps(context.Background())
 	if err != nil {
-		return errors.Wrap(err, "Error building, (re)creating or starting project containers")
+		return errors.Wrap(err, "Error checking docker-compose status")
+	}
+
+	if len(psInfo) > 0 {
+		// Ensure project is not alreasdy running
+		for _, info := range psInfo {
```
andscoop: Prevents a user from running start and having a hard error on an already running project.

airflow/docker.go
```diff
@@ -107,30 +107,50 @@ func Start(airflowHome string) error {
...
+			if info["State"][0:2] == "Up" {
```
schnie: Wonder if there's a constant for "Up" in `libcompose`. If not, oh well.

airflow/docker.go
schnie: Yea, how about just `createProject`?

andscoop:
@schnie made requested changes This should be good to go

PR #44, opened 2018-03-06
Feature/36 42 config improvements

### LINKED ISSUE #36: Add --global / -g flags to astro config set/get
Current astro config commands only set configs in the project directory, or "fail" silently if not in a project directory. We need to add a global option to set configs in the users home directory.

### LINKED ISSUE #42: Rename config.json to astro.json to avoid old CLI conflicts
Proposal to rename `./astro/config.json` to `./astro/astro.json` in order to avoid conflicts between configurations in the old CLI and the new CLI. This could make dev and internal work easier while we continue to support two CLIs.

andscoop:
- closes #36  adding -g flag
- closes #42  changing json config to yaml
- refactors airflow checkForProject to use PreRun Hooks
- moves default config setting to cfg struct

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,14 +30,14 @@ var (
 
 	// CFG Houses configuration meta
 	CFG = cfgs{
-		PostgresUser:      initCfg("postgres.user", true, true),
-		PostgresPassword:  initCfg("postgres.password", true, true),
-		PostgresHost:      initCfg("postgres.host", true, true),
-		PostgresPort:      initCfg("postgres.port", true, true),
-		RegistryAuthority: initCfg("docker.registry.authority", true, true),
-		RegistryUser:      initCfg("docker.registry.user", true, true),
-		RegistryPassword:  initCfg("docker.registry.password", true, true),
-		ProjectName:       initCfg("project.name", true, true),
+		PostgresUser:      initCfg("postgres.user", true, true, true, "postgres"),
+		PostgresPassword:  initCfg("postgres.password", true, true, true, "postgres"),
+		PostgresHost:      initCfg("postgres.host", true, true, true, "postgres"),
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -52,8 +52,8 @@ func InitConfig() {
 	initProject()
 }
 
-func initCfg(path string, gettable bool, settable bool) cfg {
-	cfg := cfg{path, gettable, settable}
+func initCfg(path string, gettable bool, settable bool, setD bool, d string) cfg {
```
schnie: @andscoop this method signature is starting to feel a little off. Same for the init block of callers above it.

What if we added a `NewCfg` method to `meta.go` that acts as a "constructor" and returns a new `cfg` object, with default values. Seems like a lot of this the invocations of this method use the same defaults. Maybe we could cut out some boilerplate, add some default values, and let callers of `NewCfg` directly override specific values by name. Something like

```
PostgresUser := NewCfg(cfg{Name: "postgres.user", Default: "postgres"})
```

Really comes down to moving the `initCfg` function to live with the `cfg` type definition and using the `NewXXX` pattern. Rather than having a seemingly arbitrary list of parameters to pass every time, we can set default values, that are overridden specifically by name, rather than just appearing in a list.

schnie: Maybe we should rename `meta.go` to `types.go` as well. Then we could have multiple `Cfg` types that maybe share some group of defaults. For example, `GettableCfg`, `SetableCfg`. Not sure if multiple types is really needed yet.

Also, do we need a `setD` param here? Could we get away with just checking if the `Cfg` has a `Default` value set when created and call `setDefault` if it's not `nil` or an empty string?

andscoop:
@schnie pushed change that
- removes gettable and settable props as we discussed offline about them being pointless for now. all configs should be gettable and settable - otherwise we will hardcode them
- renames meta.go to types.go

PR #46, opened 2018-03-08
add example dag on project init

### LINKED ISSUE #4: init command should create a “hello world” DAG in CLI

andscoop:
- closes #4


schnie:
@andscoop what you think about breaking these files out into their own?

andscoop:
We can - I thought above moving them to `include/`

schnie:
Sounds good

andscoop:
@schnie I broke the files out into `/airflow/include`. I originally planned to put them into `/include/` but this would have potentially conflicted with the `/include/` dir we create on `astro airflow init` if a developer is working from root astro directory.

PR #48, opened 2018-03-15
check for non-200 status code on registry request

andscoop:
This pr
 - throws an error on all non-200 status codes
 - adds temporary fix for silent 401 errors when attempting to deploy to a registry with unset auth configs.

The temporary fix can be removed as soon as houston is handling auth to registry

### DIFF
--- a/docker/registry.go
+++ b/docker/registry.go
@@ -34,7 +34,19 @@ func ListRepositoryTags(repository string) ([]string, error) {
 
 	// Make the request
 	resp, reqErr := client.Do(req)
-	if reqErr != nil {
+	// TODO Remove config suggestion and 401 check once houston is handling auth
+	if resp.StatusCode == 401 {
+
+		fmt.Println(`Failed to authenticate to registry
+
+You can set your registry auth using
+	astro config set docker.registry.user [registry_user] -g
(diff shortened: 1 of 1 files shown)


andscoop:
Updated pr to remove non 200 status code checks.

- It will not be needed when houston handles this portion
- it was having an unintended consequence of obscuring cert errors

docker/registry.go
```diff
@@ -34,10 +34,23 @@ func ListRepositoryTags(repository string) ([]string, error) {
...
 	if reqErr != nil {
 		return []string{}, errors.Wrap(reqErr, "Error requesting repositories")
 	}
 
+	// TODO Remove config suggestion and 401 check once houston is handling auth
+	if resp.StatusCode == 401 {
+
+		fmt.Println(`Failed to authenticate to registry
+	
+	You can set your registry auth using
```
schnie: @andscoop what's up with this formatting?

andscoop: @schnie equivalent to 

`fmt.Println("Failed to authenticate to registry\n\nYou can set your registry auth using \n    astro config set docker.registry.user [registry_user] -g \n    astro config set docker.registry.password [registry_password] -g")`

I can change it, wrote it the other way to be more readable in code.

PR #53, opened 2018-04-17
Capture and pass auth to docker login

andscoop:
This commit allows for us to capture credentials in the same command we auth to the private registry. It paves the way for a single auth command to login to houston and the registry

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -37,7 +37,7 @@ func ExecLogin(registry, username, password string) error {
 	}
 
 	loginCmd := fmt.Sprintf("echo '%s' | docker login %s -u %s --password-stdin", password, registry, username)
-	cmd := exec.Command("bash", "-c", loginCmd)
+	cmd := exec.Command("/usr/bin/env", "sh", "-c", loginCmd)
 
 	cmd.Stdout = os.Stdout
 	cmd.Stdin = os.Stdin
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


docker/docker.go
```diff
@@ -26,3 +27,21 @@ func Exec(args ...string) {
...
+// ExecLogin executes a docker login
+// Is a workaround for submitting password to stdin without prompting user again
+func ExecLogin(registry, username, password string) error {
+	_, lookErr := exec.LookPath(Docker)
+	if lookErr != nil {
+		panic(lookErr)
+	}
+
+	loginCmd := fmt.Sprintf("echo '%s' | docker login %s -u %s --password-stdin", password, registry, username)
+	cmd := exec.Command("bash", "-c", loginCmd)
```
schnie: @andscoop any way we can avoid calling `bash` directly? How do we exec things out elsewhere? If possible, I'd like to just use whatever the default shell is on the system. Can we just call the `loginCmd` directly?

cwurtz: @schnie We can work on getting it to use the docker go api in the coming weeks, but there isn't an exposed `Login(username, password)` function that will do everything, we'd have do everything the current docker login cobra command does and fill in any gaps of unexposed go function. 

We also can't just directly exec `docker login` because it won't accept a password via the `--password` flag anymore without issuing a warning of `WARNING! Using --password via the CLI is insecure. Use --password-stdin.`. 

As far as running in the default shell, @andscoop, you should be able to do `exec.Command("/usr/bin/env", "sh", "-c", loginCmd)`

schnie:
Cool, LGTM

PR #54, opened 2018-04-23
Feature/base64 encode credentials

### LINKED ISSUE #23: Remove need for registry user/password in config file
We really shouldn't need to have the user/password in the global config file for the registry. Would be great if we can use the creds created after we `astro auth login`.

andscoop:
- base64 encodes credentials so that we don't have them sitting in human readable text
- Is a good interim step until we can fully resolve and close #23 
- breaking change due to changing auth configurations

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -177,9 +177,9 @@ func saveConfig(v *viper.Viper, file string) error {
 	return nil
 }
 
-// FetchDecodedAuth fetches auth string from config, decodes and
+// GetDecodedAuth fetches auth string from config, decodes and
 // returns username password
-func FetchDecodedAuth() (string, string, error) {
+func GetDecodedAuth() (string, string, error) {
 	encodedAuth := CFG.RegistryAuth.GetString()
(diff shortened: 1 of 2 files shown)


schnie:
Looks good. Whacha think @cwurtz?

cwurtz:
I'm fine with it in the interim.  We should probably check if the unencrypted values exist in the config, (maybe attempt an auth with them to verify they work) and then encoded them and delete them from the config.  Should maintain backwards compatibility a little better

andscoop:
CJ and I chatted quickly offline yesterday. The logic required to check for existing username and password is probably more code than we would like to drag around to support the erasing of credentials for the current users of the CLI. 

If we merge as is, users will be prompted to re-authenticate and that will store all future creds in the base64 encoded auth config setting. If this is okay with everyone, we can merge as is.

schnie:
Let's just roll with it. If they just have to re-authenticate then that's not too big of a deal.

PR #56, opened 2018-04-25
Check for uncommited changes before deploy

### LINKED ISSUE #13: Check for uncommitted code before deploying
Add check to `astro airflow deploy` to check for uncommitted code if user is using git. Gracefully do nothing if not in a git project.

If user is using git and they have uncommitted changes, stop deployment and show warning. Warning should specify they can `astro airflow deploy -f` to force deployment.

Otherwise, if they are using git and are clean, then allow the deploy to go right through.

andscoop:
- closes #13 
- checks if git repo has been init'd
- then checks if there are uncommitted changes
- adds -f flag to `astro airflow deploy` to force a deploy if there are uncommitted changes

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -12,6 +12,7 @@ import (
 var (
 	projectRoot string
 	projectName string
+	forceDeploy bool
 
 	airflowRootCmd = &cobra.Command{
 		Use:   "airflow",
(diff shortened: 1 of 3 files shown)


cmd/airflow.go
```diff
@@ -91,6 +92,7 @@ func init() {
 
 	// Airflow deploy
 	airflowRootCmd.AddCommand(airflowDeployCmd)
+	airflowDeployCmd.Flags().BoolVarP(&forceDeploy, "force-deploy", "f", false, "Force deploy if uncommited changes")
```
schnie: @andscoop what do you think about just using `force` for the long option name, instead of `force-deploy`? Feels more standard.

andscoop:
Made changes. Merging so we can get this in before big merge of houston integration

PR #70, opened 2018-05-21
Make registry authentication explicit

### LINKED ISSUE #61: `astro auth login -d [domain]` Should Override Registry Config if Exists
Ensure that docker.registry.authority config is overwritten when cmd `astro auth login -d [domain]` is run and docker.registry.authority already exists.

andscoop:
- prompts user with registry location that they successfully authenticate to
- closes #61

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -41,6 +41,8 @@ func Login() {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


andscoop:
I think this is good enough to resolve this issue. We can re-evaluate if we end up further customer confusion down the road.

PR #80, opened 2018-06-28
Feature/50 add command to upgrade cli

### LINKED ISSUE #50: Add Command to Update or Upgrade CLI to Latest
Add a command that allows user to update or upgrade the CLI from within the cli itself

`astro update`
or
`astro upgrade`

andscoop:
- closes #50

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -5,38 +5,47 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
+	"time"
 
 	"github.com/astronomerio/astro-cli/pkg/httputil"
 	"github.com/pkg/errors"
 )
 
+var (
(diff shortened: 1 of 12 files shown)


pkg/github/github.go
```diff
@@ -0,0 +1,99 @@
...
+	"io/ioutil"
+	"strings"
+	"time"
+
+	"github.com/astronomerio/astro-cli/pkg/httputil"
+	"github.com/pkg/errors"
+)
+
+var (
+	http = httputil.NewHTTPClient()
```
andscoop: @cwurtz How do you feel about having these at the API level instead of moving them into the cmd level like we have it with houston? This allows us to treat the `github` package as a complete wrapper to the API, not having to worry about client instantiation

PR #106, opened 2018-08-08
Prevent dir walking when searching for a project

### LINKED ISSUE #103: Astro init looks at parent directory
**Observation:** astro airflow deploy is traversing parent level directories for the first .astro file it finds. This results in child astro projects not working as expected.

**Steps to reproduce:**
Step 1: astro airflow init a parent level directory
Step 2: Create a new directory within the parent directory
Step 3: astro airflow init within the new child directory
Step 4: Run astro airflow deploy in your child directory

**Expected outcome:**
astro airflow init does not traverse parent directories, astro airflow deploy does not traverse parent directories

**Actual outcome:**
astro airflow init and astro airflow deploy traverse parent directories and push their contents rather than the current project directory.

**Story time:**
At some point I may have done an astro airflow init in my Documents folder. Later I created new project directories within the documents folder. When running an astro airflow init in the new project directory I was told that the existing project was being reinitialized. 

Running an astro airflow deploy pushed the entire contents of my Documents directory, rather than the current project folder I was in.

andscoop:
This ended up being more re-work than expected as `projectRoot` was always looking in the curr dir and walking upwards towards usr root dir if config not found. All walking of dirs has been removed in favor of commands that need a config being run directly out of a project root.

- closes #103
- adds prompt on `astro init` if user is in non-empty directory
- removes the upward walking of a dir if user is not in an immediate project dir (defined by existence of config)

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -1,16 +1,18 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/astronomerio/astro-cli/pkg/input"
+	"github.com/iancoleman/strcase"
+	"github.com/pkg/errors"
+
 	"github.com/astronomerio/astro-cli/messages"
 
-	"github.com/iancoleman/strcase"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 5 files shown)


schnie:
Makes sense, sounds good to me.

PR #113, opened 2018-08-22
Add command to output local airflow logs

### LINKED ISSUE #88: Output Scheduler Logs for Local Development
As a user, having scheduler logs when debugging a custom plugin or workflow would be really helpful. Often times, the error message exposed in the UI doesn't contain the full traceback.

Something like `astro airflow start --logs` or something would be a great.

I know a few customers have bought this up for cloud CLI, so just wanted to make sure that feedback was captured here.

andscoop:
- closes #88


benjamingregory:
Tested. With the exception of not signaling that logging does not work if an airflow project is not running locally, it works as I would expected.

I believe @andscoop is making a fix for ^^ now.

schnie:
@benjamingregory @andscoop what happens if you aren't running the dev cluster locally yet? Error message?

andscoop:
@schnie Added prompt for user to start project to view logs. Before it was null output

andscoop:
Fixed with #113

PR #134, opened 2018-10-02
Disable airflow examples to match production environment

schnie:


andscoop:
LGTM

schnie:
Matches here: https://github.com/astronomerio/helm.astronomer.io/blob/master/charts/airflow/templates/_helpers.yaml#L19-L20

This was one of the few (maybe only) settings we changed from the default cfg. Since we removed it, we need to patch it in both places.

PR #143, opened 2018-11-12
Feature/dynamic padding

benjamingregory:
~Currently only works when returning multiple rows~

### DIFF
--- a/deployment/deployment.go
+++ b/deployment/deployment.go
@@ -10,8 +10,8 @@ import (
 
 var (
 	tab = printutil.Table{
-		Padding: []int{30, 50, 50},
-		Header:  []string{"NAME", "RELEASE NAME", "DEPLOYMENT ID"},
+		Padding: []int{30, 30, 10, 50},
+		Header:  []string{"NAME", "RELEASE NAME", "CHART", "DEPLOYMENT ID"},
 	}
 )
 
@@ -32,7 +32,7 @@ func Create(label, ws string) error {
 	if err != nil {
 		return err
 	}
-	tab.AddRow([]string{d.Label, d.ReleaseName, d.Id}, false)
+	tab.AddRow([]string{d.Label, d.ReleaseName, d.Version, d.Id}, false)
 	tab.SuccessMsg = "\n Successfully created deployment. Deployment can be accessed at the following URLs \n" +
 		fmt.Sprintf("\n Airflow Dashboard: https://%s-airflow.%s", d.ReleaseName, c.Domain) +
 		fmt.Sprintf("\n Flower Dashboard: https://%s-flower.%s", d.ReleaseName, c.Domain)
@@ -87,14 +87,17 @@ func List(ws string, all bool) error {
 
 	deployments = r.Data.GetDeployments
 
+	rows := [][]string{}
 	// Build rows
 	for _, d := range deployments {
 		if all {
 			ws = d.Workspace.Uuid
 		}
-
-		tab.AddRow([]string{d.Label, d.ReleaseName, d.Id}, false)
+		row := []string{d.Label, d.ReleaseName, d.Version, d.Id}
+		rows = append(rows, row)
 	}
+
+	tab.AddRows(rows, false)
 
 	tab.Print()
 
@@ -115,7 +118,7 @@ func Update(uuid string, args map[string]string) error {
 
 	d := r.Data.UpdateDeployment
 
-	tab.AddRow([]string{d.Label, d.ReleaseName, d.Id}, false)
+	tab.AddRow([]string{d.Label, d.ReleaseName, d.Version, d.Id}, false)
 	tab.SuccessMsg = "\n Successfully updated deployment"
 	tab.Print()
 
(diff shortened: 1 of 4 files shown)


benjamingregory:
Realized the changes need to still be made to `astro list cluster`

deployment/deployment.go
```diff
@@ -87,15 +88,19 @@ func List(ws string, all bool) error {
 
 	deployments = r.Data.GetDeployments
 
+	rows := []printutil.TempRow{}
```
andscoop: Do we need to instantiate out the row slice outside of the `printutil` pkg?

deployment/deployment.go
```diff
@@ -87,15 +88,19 @@ func List(ws string, all bool) error {
...
+
 	// Build rows
 	for _, d := range deployments {
 		if all {
 			ws = d.Workspace.Uuid
 		}
-
-		tab.AddRow([]string{d.Label, d.ReleaseName, d.Version, d.Id}, false)
+		row := printutil.TempRow{[]string{d.Label, d.ReleaseName, d.Version, d.Id}, false}
```
andscoop: Let's keep this part simple and limit it to a single call for adding a row to the table vs two calls for adding a row and a third call to add all rows.

pkg/printutil/printutil.go
```diff
@@ -136,3 +170,40 @@ func strSliceToInterSlice(ss []string) []interface{} {
...
+
+func (t *Table) coalesceHeaderAndRowPadding(val int, i int) {
+	if val < t.headerPadding[i] {
+		t.altPadding[i] = t.headerPadding[i]
+	} else {
+		t.headerPadding[i] = t.altPadding[i]
+	}
+}
+
+func (t *Table) dynamicPadding(rows []TempRow) {
```
andscoop: if you make []TempRow an attribute on the Table struct you can avoid having to pass an arg in here

pkg/printutil/printutil.go
```diff
@@ -136,3 +170,40 @@ func strSliceToInterSlice(ss []string) []interface{} {
...
+	for _, row := range rows {
+		for i, col := range row.Values {
+			colLength := len(col) + 5
+			if len(t.altPadding) != len(row.Values) {
+				t.altPadding = append(t.altPadding, colLength)
+			} else {
+				if t.altPadding[i] < colLength {
```
andscoop: What is case covering?

pkg/printutil/printutil.go
```diff
@@ -35,6 +35,18 @@ type Table struct {
 	// Function which will eval whether to apply color to a col
 	// ColorColCond    func(t *Table) (bool, error)
 	// ColorColCode [2]string
+
+	altPadding []int
+
+	headerPadding []int
```
andscoop: Are we able to re-use altPadding here? Is there a usecase when the header wouldn't be the same to the rest of the columns if we're left aligning?

pkg/printutil/printutil.go
andscoop: Remove nested if statement in favor of another top-level `if` or an `else if`, depending on exact needs

pkg/printutil/printutil.go
```diff
@@ -35,6 +35,18 @@ type Table struct {
...
+
+	DynamicPadding bool
+}
+
+// comment
+type TempRow struct {
```
andscoop: It may make sense to remove this TempRow struct and instead combine the attributes you need into the Row struct. This will allow you to avoid passing around multiple row types and instead modify rows in place.

deployment/deployment.go
andscoop: Do we need to instantiate out the row slice outside of the printutil pkg?

deployment/deployment.go
```diff
@@ -87,15 +88,19 @@ func List(ws string, all bool) error {
...
+		rows = append(rows, row)
```
andscoop: Let's keep this part simple and limit it to a single call for adding a row to the table vs two calls for adding a row and a third call to add all rows.

 In general reduce in the number of ways a developer can interface with an object.

pkg/printutil/printutil.go
benjamingregory: If the altPadding slice is smaller than the number of incoming Values slice, then that means that values still need to be added to fill it up. 

If the altPadding slice length is equivalent to the Values slice length, then it should compare the length of the incoming value being evaluated to see if it longer than the value already stored in it's place in altPadding. If it is, replace it's value with the length of the incoming value.

This helps ensure as it iterates through each row that the length of the longest value for each column is kept or replaced as new values are introduced.

benjamingregory: I added ^^ as a comment to the code itself.

pkg/printutil/printutil.go
benjamingregory: Removed headerPadding and associated methods in favor of altPadding.

pkg/printutil/printutil.go
benjamingregory: Fixed.

andscoop:
Looks good, thanks for making the changes

PR #144, opened 2018-11-13
Change ColorRowCode to work with Terminal

benjamingregory:


andscoop:
@benjamingregory Looks good - like that it's readable on white and black now across both apps. Can you apply the same fix for `astro cl ls`? It appears to suffer from the same bug

Issue #140, opened 2018-10-12
Add support for .env file for secrets


vparekh94:
As a user, I might use private github repos to store python packages in my org. It would be nice to have the ability to pass those secrets in through some sort of .env file and be able to use them to auth+pip install from my git repo. 

@schnie I think this is a good summary of our conversation - anything to add?

schnie:
Yep, looks good. We should probably add support for runtime env vars, as well as build time env vars.

mplovepop:
I also would like this!

benjamingregory:
We're working to get this included in the next CLI release but here is a workaround for now:
In you astronomer project (after `astro airflow init`)...
1) Put the following in a file labeled Dockerfile.build
```
FROM astronomerinc/ap-airflow:0.6.2 AS stage1
LABEL maintainer="Astronomer <humans@astronomer.io>"
ARG BUILD_NUMBER=-1
LABEL io.astronomer.docker=true
LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
LABEL io.astronomer.docker.airflow.onbuild=true
# Install alpine packages
COPY packages.txt .
RUN cat packages.txt | xargs apk add --no-cache

FROM stage1 AS stage2
RUN mkdir -p /root/.ssh
ARG PRIVATE_RSA_KEY=""
RUN echo "${PRIVATE_RSA_KEY}" >> /root/.ssh/id_rsa
RUN chmod 600 /root/.ssh/id_rsa
RUN apk update && apk add openssh
RUN ssh-keyscan -H github.com >> /root/.ssh/known_hosts
# Install python packages
COPY requirements.txt .
RUN pip install --no-cache-dir -q -r requirements.txt

FROM stage1 AS stage3
# Copy requirements directory
COPY --from=stage2 /usr/lib/python3.6/site-packages/ /usr/lib/python3.6/site-packages/
```
2) Build the image with `docker build -f Dockerfile.build --build-arg PRIVATE_RSA_KEY="$(cat ~/.ssh/id_rsa)" -t custom-ap-airflow .` If your private key is not held at `~/.ssh/id_rsa`, change the reference to the relevant path.
3) Replace the contents of Dockerfile from `FROM astronomerinc/ap-airflow:latest-onbuild` to `FROM custom-ap-airflow`
4) Issue `astro airflow start`

paolaperaza:
Thanks, @benjamingregory ! Posted this on our forum too, for reference.

[...]

PR #148, opened 2018-11-21
Add support for secrets in .env file

benjamingregory:
Resolves https://github.com/astronomer/astro-cli/issues/140


schnie:
@benjamingregory this is useful for setting environment variables at _runtime_, but I think this issue was created from the need to be able to `pip install` a package from a private github repo. Specifically I think there is a need to be able to pass an ssh key into a container temporarily as its being built, so the repo can be cloned during the `docker build` phase.

I think @vparekh94 was working with a customer and may have a better concrete use case.

PR #155, opened 2018-11-29
Add deploy label and dynamicPadding

### LINKED ISSUE #139: Add Label to Astro Airflow Deploy
Not having label in the `astro airflow deploy` selection prompt forces a user to run an `astro deployment list` if they are used to working off of instance name.

benjamingregory:
Resolves #139


andscoop:
Looks good!

Issue #159, opened 2018-12-07
Allow for custom .env reference when building local image


benjamingregory:
When building and testing new dags, you often need distinct environments between testing, staging, production.

In the current system having to replace your docker file for testing and before deploying is not convenient and can lead to human errors. It would be really convenient to be able to specify to Dockerfile source file like `astro airflow start FILE_PATH` or `astro airflow start -p FILE_PATH` the same usage could apply to the deploy command as well.

benjamingregory:
@Tristan-Soriano @emirot -- Could you elaborate on your original request?

My understanding was that the need for multiple Dockerfiles was to use alternative env vars but want to make sure nothing is lost in translation.

Tristan-Soriano:
I find it interesting to have a dev Dockerfile and a prod Dockerfile.
For examples on my local dev environment:
 - I have a fernet key
-  I want to be able to test email alerts
-  I want to modify airflow.cfg using ENV variables
-  I want a way to set up all my dev connection without dealing with it in the UI or having extra script to run once "astro airflow start" is done

schnie:
Hey @Tristan-Soriano thanks for the input! Our latest release includes support for setting environment variables through the UI for deployments running out on kubernetes. All environment variables submitted through our UI are stored in a kubernetes secret and injected into all airflow containers. 

We also automatically generate and store a fernet key in our helm charts for deployments on kubernetes - https://github.com/astronomer/helm.astronomer.io/blob/master/charts/airflow/templates/_helpers.yaml#L29-L33.

I think that solves some of the issues for deployments running in the cloud, but not locally via `astro airflow start`.

@benjamingregory and I were talking about adding support for a `.env` file that could live inside the project directory and could be `.gitignore`d  and `.dockerignore`d by default. We could source the env vars in that file into the locally running containers. Would something like that work for you?

As for the automatic connection setup, we may need to think about that one some more.

benjamingregory:
@Tristan-Soriano We're going to use this issue to focus on the request to make the passing of ENV variables. The dev connection feature would fall outside of this so I've put it in an issue here: https://github.com/astronomer/astro-cli/issues/160

We could add a flag on `astro airflow start` to specify which .env to use. e.g. `astro airflow start --env local.env` vs `astro airflow start --env prod.env`.

Would that solve your use case for that issue?

Tristan-Soriano:
@benjamingregory I believe so, thanks

PR #165, opened 2019-01-09
Allow for custom env files

benjamingregory:
Resolves https://github.com/astronomer/astro-cli/issues/159

- Creates .env on init

Example behavior:

Using .env file
```
astro airflow start
```

Using custom.env file
```
astro airflow start --env custom.env
```


benjamingregory:
Notably, this does not bundle .env files with a deploy, only on start when building locally

schnie:
LGTM

Issue #160, opened 2018-12-10
Set up connections by default with `astro airflow init`


benjamingregory:
Request by @Tristan-Soriano: 

I want a way to set up all my dev connection without dealing with it in the UI or having extra script to run once "astro airflow start" is done

benjamingregory:
Proposed YAML structure
```
airflow:
  connections:
      - conn_id: my_s3
        conn_type: s3
        conn_extra: {"key":"XXX", "secret": "XXX"}
      - conn_id: my_postgres
        conn_type: Postgres
        conn_host: 12.2.2.1
        conn_login: user
        conn_password: password
        conn_port: 5342
  pools:
      - pool_name: my_new_pool
        pool_slot: 4
        pool_description: for_the_haters
  variables:
      - variable_name: new_variable
        variable_value: {"Never Gonna": "Give You Up"}
      - variable_name: second_variable
        variable_value: {"Never Gonna": "Let You Down"}
```

benjamingregory:
PR here: https://github.com/astronomer/astro-cli/pull/168

PR #168, opened 2019-01-10
Feature/connections

benjamingregory:
Resolves https://github.com/astronomer/astro-cli/issues/160

Creates settings.yaml file on init:

```
airflow:
    connections:
      - conn_id:
        conn_type:
        conn_host:
        conn_login:
        conn_password:
        conn_port:
    pools:
      - pool_name:
        pool_slot:
        pool_description:
    variables:
      - variable_name:
        variable_value:
```

Example behavior: 
On `astro airflow start`, connections, pools, and variables are created as defined.

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -284,19 +284,19 @@ func Start(airflowHome string, envFile string) error {
 				if len(pool.PoolName) == 0 && pool.PoolSlot > 0 {
 					fmt.Print("Skipping Pool Creation: No Pool Name Specified.")
 				} else {
-				airflowCommand := fmt.Sprintf("airflow pool -s \"%s\" \"%s\" \"%s\"", pool.PoolName, pool.PoolSlot, pool.PoolDescription)
-				cmd := exec.Command("docker", "exec", "-it", info["Id"], "bash", "-c", airflowCommand)
-				cmd.Stdin = os.Stdin
-				cmd.Stderr = os.Stderr
-				if cmdErr := cmd.Run(); cmdErr != nil {
-					return errors.Wrapf(cmdErr, "Error adding %s Pool", pool.PoolName)
-				} else {
-					fmt.Printf("Added Pool: %s\n", pool.PoolName)
+					airflowCommand := fmt.Sprintf("airflow pool -s \"%s\" \"%v\" \"%s\"", pool.PoolName, pool.PoolSlot, pool.PoolDescription)
+					cmd := exec.Command("docker", "exec", "-it", info["Id"], "bash", "-c", airflowCommand)
+					cmd.Stdin = os.Stdin
+					cmd.Stderr = os.Stderr
+					if cmdErr := cmd.Run(); cmdErr != nil {
+						return errors.Wrapf(cmdErr, "Error adding %s Pool", pool.PoolName)
+					} else {
+						fmt.Printf("Added Pool: %s\n", pool.PoolName)
+					}
 				}
 			}
 		}
 	}
-
 	fmt.Printf(messages.COMPOSE_LINK_WEBSERVER+"\n", config.CFG.WebserverPort.GetString())
 	fmt.Printf(messages.COMPOSE_LINK_POSTGRES+"\n", config.CFG.PostgresPort.GetString())
 	return nil
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -248,9 +248,7 @@ func Start(airflowHome string, envFile string) error {
 		if strings.Contains(info["Name"], "scheduler") {
 			for _, conn := range settings.Airflow.Connections {
 
-				if len(conn.ConnID) == 0 {
-					fmt.Print("Skipping Connection: Conn ID must be specified.")
-				} else if len(conn.ConnType) == 0 && len(conn.ConnUri) == 0 {
+				if len(conn.ConnID) > 0 && len(conn.ConnType) == 0 && len(conn.ConnUri) == 0 {
 					fmt.Printf("Skipping %s: ConnType or ConnUri must be specified.", conn.ConnID)
 				} else {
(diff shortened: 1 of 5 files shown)


airflow/docker.go
```diff
@@ -54,6 +58,32 @@ type ComposeConfig struct {
 	AirflowWebserverPort string
 }
 
+// SettingsConfig is input data to generate connections, pools, and variables
```
schnie: @benjamingregory could maybe break out any nested types into top-level code and just refer to those types as you create new types, like:

```
type SettingsConfig struct {
   Airflow: AirflowConfig
}
```
where the `AirflowConfig` is defined as a top-level type.

Could help avoid some pyramid of doom indentation.

airflow/docker.go
```diff
@@ -106,6 +136,25 @@ func generateConfig(projectName, airflowHome string, envFile string) (string, er
...
+	err = yaml.Unmarshal(buff, &settings)
+
+	if err != nil {
+		return settings, errors.Wrap(err, "failed to generate settings")
+	}
+
+	return settings, nil
+}
+
 func checkServiceState(serviceState, expectedState string) bool {
```
schnie: @benjamingregory this function could probably be broken out into some more modular chunks to help avoid the deep indentation.

schnie (commented):
@benjamingregory since this is essentially just another config file, I wonder if it'd be useful to just utilize `cobra`, like we do for the main configuration files.

Could possibly trim down the code and provide some nice helpers for accessing configurations.

schnie:
@benjamingregory looking pretty good. How do you execute these commands again? I can't seem to find the entrypoint from `cobra`?

benjamingregory:
@schnie -- They're built on start - https://github.com/astronomer/astro-cli/blob/a97af751d9c323b60313dfc1c27f44bb6565d249/airflow/docker.go#L199

There's not a way right now to just create connections without also starting the containers.

Do you think there should be a separate entry point?

benjamingregory:
@schnie -- I'm confident it works but I also added a note in the settings.yaml config file stating that the feature is in beta and to open an issue in this repo for any issues.

schnie:
Acted on all ten. Two were the same shape as the PR's own thesis and I'd missed them, and two of my "fixed" mechanisms turned out to be doing nothing until this review made me test them.

## The consolidation was half done

**The comment rejecting the helper was still above the line that calls it.** Right, and it's the worst of the ten: a live instruction to copy rather than share, in the file that shares. Gone.

**The root module kept two more byte-identical copies.** Also right, and it undercuts the premise I opened with — "three times, once per module" was four places, not three, and `cmd/local` already has the require + replace pair so it cost nothing to fix. Both now use `instancestest.Preamble`, and the blast radius says so: dropping `[tool.astro]` from the one definition now fails **41 tests in `pkg/instances`, 23 in `pkg/awsauth`, 2 in `pkg/googleauth` and 25 in `cmd/local`**.

I left `cmd/local/env_test.go:19` alone deliberately — it looks like the same string but has no `workspace`, and workspace-level env resolution is exactly what its neighbours test, so pointing it at a preamble that adds one would change what it covers rather than dedupe it.

## `link`: one is better than two

Your framing beat mine — three copies means two is strictly an improvement, so "two definitions rather than one" was not the argument I thought it was. But your fix keeps a copy, and there's a version that keeps none: the cycle only forbids *naming `instances.Set`* here, not sharing the assertion. `OneLink` is generic over the instance type and takes the names for its message, so it stays on this side of the cycle and all three packages route through it:

```go
func link(t *testing.T, body string) instances.Instance {
	t.Helper()
	set := instances.Build(instancestest.Manifest(t, body))
	return instancestest.OneLink(t, set.All(), set.Names())
}
```

Nine lines to three, every one a call into shared code, and no fourth package. Changing `len(all) != 1` to `!= 2` fails 28 / 23 / 2 tests, so it is genuinely the one definition and not just present.

## Two fixes that were doing nothing

Both of these I'd have reported as fixed. Neither was, and reverting them is how I know.

**archlint.** The widening is one string, as you said, but I wanted to see the gap before trusting it. Importing `oauth2/google` from `instancestest` fails the widened rule with `the pkg/instances module reaches golang.org/x/oauth2/google/internal/impersonate` and **passes the rule this PR shipped**. Real gap, closed.

**The lint exemption.** My first attempt was `^instancestest/`, on my own reasoning that golangci-lint matches paths relative to the module dir a `lint-submodules` run is in. It doesn't — it reports and matches relative to the config — so the anchor meant the pattern never fired and my "0 issues" was the same 0 issues as before. A planted `_ = probeErr()` was rejected either way. Unanchored now: the probe passes with the exemption and `errcheck` fires without it, and the comment explains the actual matching rather than my guess about it.

That also caught a `misspell` in a comment I'd just written, which is the one finding this review didn't have to tell me about.

## Two behaviors, now guarded rather than described

You were right that promoting these to shared API used by three modules is what makes them worth fixing, and a fix without a test is a comment.

- **`Env` closed over the caller's map.** `maps.Clone` at construction. The test mutates the map after the call and asserts the lookup is unchanged; without the clone it fails with `AF_ID = ""/false after the caller deleted it`.
- **`Header` conflated a no-scheme value with sending nothing.** Refused now. To test a refusal that used to be a `t.Fatalf`, `Header`'s answer moved into an unexported `header` returning an error — so this package's own tests can assert it instead of being killed by it. The refusal deliberately does not echo the credential.

## Also

**`PreambleFor(airflow)`**, with `Preamble = PreambleFor("3.1")`, so the last inline copy is gone. Worth noting your prediction was exactly right: the copy's second difference was that it dropped `workspace`, the comment mentioned only the version, and adding `workspace` back changes no outcome in any of the local-mint tests — I checked before parameterizing, so the difference was unstated *and* not load-bearing.

**The doc and both `go.mod` comments** now say what the doors actually do with `pkg/manifest` — name an `AuthMethod` to key themselves into a `Providers` map — and register `pkg/instances/instancestest`, per the doc's own "New exceptions get added here, not assumed".

**The two `Fatalf`s that re-invoked the source.** Both now use the `if got := ...` form the rest of the file uses. Worth spelling out why it mattered here specifically: the mint tests assert on `got.method`/`got.username` recorded by the stub server, so the second request overwrote the evidence for the assertion that follows.

## Verification

`go build ./...` and `go test ./...` on the root, `scripts/test-submodules.sh` green, `make lint-submodules` 0 issues across all seven, root `golangci-lint` 0 issues, `gofumpt` clean. Each new guard confirmed by reverting it: four mutations caught, two config guards shown to pass a violation before the fix and fail it after.

Issue #177, opened 2019-02-25
Switch wording from UUID to ID


schnie:
We have some legacy wording in the CLI, as of 0.8. We should update all labels from UUID to just read ID.

PR #184, opened 2019-02-27
Switch wording from UUID to ID

andriisoldatenko:
Fix https://github.com/astronomer/astro-cli/issues/177

### DIFF
--- a/houston/queries.go
+++ b/houston/queries.go
@@ -39,7 +39,7 @@ var (
 
 	DeploymentDeleteRequest = `
 	mutation DeleteDeployment($deploymentId: Uuid!) {
-		deleteDeployment(deploymentId: $deploymentId) {
+		deleteDeployment(deploymentUuid: $deploymentId) {
 			id
 			type
 			label
--- a/houston/queries.go
+++ b/houston/queries.go
@@ -19,15 +19,15 @@ var (
 		$label: String!
 		$type: String = "airflow"
 		$version: String = "0.7.5"
-		$workspaceUuid: Uuid!
+		$workspaceId: Uuid!
 	) {
 		createDeployment(
 			label: $label
 			type: $type
 			version: $version
-			workspaceUuid: $workspaceUuid
+			workspaceUuid: $workspaceId
 		) {
-			uuid
+			id
 			type
 			label
 			releaseName
@@ -38,16 +38,16 @@ var (
 	}`
 
 	DeploymentDeleteRequest = `
-	mutation DeleteDeployment($deploymentUuid: Uuid!) {
-		deleteDeployment(deploymentUuid: $deploymentUuid) {
-			uuid
+	mutation DeleteDeployment($deploymentId: Uuid!) {
+		deleteDeployment(deploymentId: $deploymentId) {
+			id
 			type
 			label
       description
 			releaseName
 			version
       workspace {
-        uuid
+        id
       }
 			createdAt
 			updatedAt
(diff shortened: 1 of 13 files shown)


houston/queries.go
```diff
@@ -19,15 +19,15 @@ var (
...
 		$version: String = "0.7.5"
-		$workspaceUuid: Uuid!
+		$workspaceId: Id!
 	) {
 		createDeployment(
 			label: $label
 			type: $type
 			version: $version
-			workspaceUuid: $workspaceUuid
+			workspaceId: $workspaceId
```
schnie: @andriisoldatenko I think we may have to keep the parameter names, as they are sent to houston, with the old UUID naming. Internally, we should use the `id` terminology, as well is in any labels that show up while using the CLI. 

I would think that this would fail due to `workspaceId` being sent on this line, instead of `workspaceUuid`, as  seen here: https://github.com/astronomer/houston-api/blob/master/src/schema.graphql#L290.

I think it should work if you just change this line (and similar lines in other mutations to `workspaceUuid: $workspaceId`, keeping it internally named `workspaceId`, just sending `workspaceUuid` to Houston. We can patch up Orbit/Houston/CLI to fully switch to `...Id` format next, all at once.

houston/queries.go
```diff
@@ -209,9 +209,9 @@ var (
 	}`
 
 	UserGetAllRequest = `
-	query GetUsers($userUuid: Uuid, $username: String, $email: String) {
-		users(userUuid: $userUuid, username: $username, email: $email) {
-			uuid
+	query GetUsers($userId: Id, $username: String, $email: String) {
```
andriisoldatenko: @schnie unused and invalid query: 
{"errors":[{"message":"Cannot query field \"createdAt\" on type \"Email\".","locations":[{"line":8,"column":5}],"extensions":
can I remove it?

schnie: @andriisoldatenko ah, yea doesn't look like we've re-implemented it in Houston2. So there's no code in the CLI that calls that query?

If not, yea we can remove it.

andriisoldatenko:
```
➜  astro-cli git:(switch-wording-from-uuid-to-id) grep --exclude-dir=vendor -r UserGetAllRequest .
./houston/queries.go:	//UserGetAllRequest = `
```

andriisoldatenko:
@schnie please take another look. I checked main functions.

schnie:
@andriisoldatenko I think it looks good. You were able to try these changes against a live platform install and they worked?

andriisoldatenko:
@schnie sorry for delay, you can see short test results:
- Test #1 
```
go run main.go auth login datarouter.ai
Successfully authenticated to registry.datarouter.ai
```
- Test #2 
./astro deployment list
 NAME     DEPLOYMENT NAME     ASTRO     DEPLOYMENT ID

- Test #3
./astro deployment create andrii
 NAME       DEPLOYMENT NAME               ASTRO     DEPLOYMENT ID
 andrii     quasarian-atmosphere-2855     0.7.5     cjsrfgbh6006f0b52grm9n5nx

 Successfully created deployment. Deployment can be accessed at the following URLs

 Airflow Dashboard: https://quasarian-atmosphere-2855-airflow.datarouter.ai
 Flower Dashboard: https://quasarian-atmosphere-2855-flower.datarouter.ai

andriisoldatenko:
./astro cluster switch datarouter.ai
 datarouter.ai                       cjsrf9icd005r0b527ymm6w0c

 Switched cluster

./astro cluster list
 NAME
 alpha.datarouter.ai
 datarouter.ai
 localhost
 astronomer.io

andriisoldatenko:
./astro user create
Email: test@test.com
Password:
Re-enter Password:
Successfully created user test@test.com. Check your email for a verification.

schnie:
@andriisoldatenko looks like I may have a bug on the Houston side for deleting deployments. I'll look at that and merge this in today.

Issue #178, opened 2019-02-25
Fix workspace switch workflow


schnie:
Right now when a user logs into a new cluster, we don't update the active workspace, which leads to a confusing UX. Listing deployments or doing deploys appears broken, until the user realizes they need to switch to an active workspace in the active cluster. We need to fix this workflow to help users.

Maybe we could auto switch the workspace to the first workspace or default workspace in the new cluster for them, or something like that.

andriisoldatenko:
As confirmed in slack:
- attempt to use the last used workspace, unless it's been deleted, then show the menu to select.

PR #188, opened 2019-03-05
Fix workspace switch workflow

andriisoldatenko:
Fix https://github.com/astronomer/astro-cli/issues/178

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -2,12 +2,13 @@ package auth
 
 import (
 	"fmt"
-
 	"github.com/astronomer/astro-cli/cluster"
+	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/docker"
 	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/messages"
 	"github.com/astronomer/astro-cli/pkg/input"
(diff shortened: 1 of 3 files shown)


auth/auth.go
```diff
@@ -152,7 +152,7 @@ func Login(domain string, oAuthOnly bool) error {
 
 	workspaces := wsResp.Data.GetWorkspaces
 
-	if len(workspaces) == 1 && len(c.Workspace) == 0 {
```
andriisoldatenko:
```
✗ go run main.go auth login localhost
 localhost                           cjsp2qct2000x0838cmbll6t3

 Switched cluster
Default "Andrii Soldatenko's Workspace" (cjsp2qct2000x0838cmbll6t3) workspace found, setting default workspace.
```

schnie:
@andriisoldatenko what happens when you switch a cluster or re-login to a cluster now if they belong to multiple workspaces? Do you think we should show a workspace selection menu?

schnie:
@andriisoldatenko do you think we could have it so we keep track of the workspace you were last using, per cluster and use that when you switch between them? If we can query for fresh data, I'd say we use the last-used workspace, and if it is now missing, them prompt them with a menu / list to choose from.

schnie:
Works great!

PR #189, opened 2019-03-06
Cleanup code

andriisoldatenko:
Fix https://github.com/astronomer/astro-cli/issues/183


auth/auth.go
```diff
@@ -28,27 +28,6 @@ func basicAuth(username string) (string, error) {
 	return resp.Data.CreateToken.Token.Value, nil
 }
 
-func getWorkspaceByLabel(label string) *houston.Workspace {
```
andriisoldatenko: unused function.

PR #225, opened 2019-05-27
Rethink login to registry

andriisoldatenko:
i have created test:
```
func TestLoginFailed(t *testing.T) {
	err := ExecLogin("https://quay.io/v1", "", "")
	if err == nil {
		t.Error(err)
	}
}
```
and results we can check in OSX:
```
cat ~/.docker/config.json
{
	"auths": {
		"https://quay.io/v1": {},
		"quay.io": {}
	},
	"HttpHeaders": {
		"User-Agent": "Docker-Client/18.09.2 (darwin)"
	},
	"credsStore": "osxkeychain"
}%
```
and windows:

```
{
	"auths": {
		"https://index.docker.io/v1/": {},
		"quay.io": {},
		"registry.datarouter.ai": {}
	},
	"HttpHeaders": {
		"User-Agent": "Docker-Client/18.09.2 (windows)"
	},
	"credsStore": "wincred"
}
```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,6 +12,7 @@ pipeline:
   # Test the code
   test:
     image: asoldatenko/dnd-golang
+    privileged: true
     commands:
       - make test
 
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,7 +11,7 @@ pipeline:
 
   # Test the code
   test:
-    image: golang
+    image: docker:dind
     commands:
       - make test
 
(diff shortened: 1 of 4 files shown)


andriisoldatenko:
@schnie also i need something like this to start writing some tests 
https://stackoverflow.com/questions/46366271/using-dind-on-drone-io can you enable it?

schnie:
@andriisoldatenko looks good! You should be able to do something similar to what we do here: https://github.com/astronomer/astronomer/blob/master/.drone.yml#L45. Would that work? The drone settings should allow for it.

andriisoldatenko:
@schnie nice catch thanks! fixed.

PR #226, opened 2019-05-28
Rename password to token

andriisoldatenko:


schnie:
@andriisoldatenko this works for windows/osx and linux?

Issue #197, opened 2019-03-15
Refactor subcommands


schnie:
Recap of a slack convo:

- Move `astro airflow deploy` to `astro deploy` (maybe keep an alias for a bit).
- Figure out how to properly separate commands that act on a deployment running in kubernetes, vs commands that act on the local instance (`astro airflow start`, etc).

schnie:
Desired | Current
-- | --
astro deploy | astro airflow deploy
astro dev * | astro airflow *
astro deployment logs | astro logs

PR #229, opened 2019-05-29
deploy subcommand moved (copied for now) to the root

rvillablanca:
This should resolve the first point of #197.

At first I only added `airflowDeployCmd` as root command but when the help of astro-cli was rendered, it is not showed correctly, so I had to copy the deploy command. The `astro airflow deploy` is still there for backward compatibility.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/astronomer/astro-cli/airflow"
+	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/messages"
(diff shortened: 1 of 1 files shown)


schnie:
@rvillablanca thanks for the PR! @andriisoldatenko what do you think?

andriisoldatenko:
@schnie lgtm, need practical tests.

andriisoldatenko:
@rvillablanca can you compile and run some practical tests?

andriisoldatenko:
let close and review https://github.com/astronomer/astro-cli/pull/233 instead.

rvillablanca:
Sorry for the delay @andriisoldatenko. I don't have much experience running airflow so I don't know how I can run that practical tests. I'll close this PR in favor of #233

PR #231, opened 2019-06-07
Support "docker-compose.override.yml" files for local customization

ashb:
This lets people mount extra files in, or run extra side-car services
that they might want for local development


ashb:
I tested this locally with the following docker-compose.override.yml file:

```yaml
version: '2'
services:
  scheduler:
    volumes:
      - .:/usr/local/airflow
```

andriisoldatenko:
@schnie please merge it when you have free min

Issue #244, opened 2019-07-02
Dead link in comment within `airflow_settings.yaml`


paolaperaza:
The `airflow_settings.yaml` file automatically generated when you `astro airflow init` via the CLI by default links out to a docs page that no longer exists/doesn't live at that URL.

**Current link:** https://github.com/astronomer/docs/blob/master/docs/cli-airflow-configuration.md 

**Link should be:** https://github.com/astronomer/docs/blob/master/v0.9/cli-airflow-configuration.md

```
$ ls
Dockerfile		airflow_settings.yaml	dags			include			packages.txt		plugins			requirements.txt
Paolas-MacBook-Pro:astro-trial paola$ more airflow_settings.yaml
# This feature is in Beta.
# Please report any bugs to https://github.com/astronomer/astro-cli/issues
# NOTE: If putting a dict for any field, please wrap in single quotes.
# More details you can find https://github.com/astronomer/docs/blob/master/docs/cli-airflow-configuration.md 
airflow:
  connections:
    - conn_id:
      conn_type:
      conn_host:
      conn_login:
      conn_password:
      conn_port:
      conn_extra:
  pools:
    - pool_name:
      pool_slot:
      pool_description:
  variables:
    - variable_name:
      variable_value:
airflow_settings.yaml (END)
```

Would we still consider this feature in beta, @schnie @andriisoldatenko? Not a huge deal here, but figured I'd create an issue to draw attention to the docs link and in case we wanted to otherwise adjust that comment at all.

PR #245, opened 2019-07-08
fix dead link in comment airflow_settings.yaml

andriisoldatenko:
fix https://github.com/astronomer/astro-cli/issues/244


paolaperaza:
Thanks @schnie @andriisoldatenko !

PR #253, opened 2019-07-15
add astro dev * aliased to astro airflow *

andriisoldatenko:
https://github.com/astronomer/astro-cli/issues/197

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -33,13 +33,21 @@ var (
 
 	RunExample = `
 # Create default admin user.
-astro airflow run create_user -r Admin -u admin -e admin@example.com -f admin -l user -p admin
+astro dev run create_user -r Admin -u admin -e admin@example.com -f admin -l user -p admin
 `
 
 	airflowRootCmd = &cobra.Command{
+		Use:     "airflow",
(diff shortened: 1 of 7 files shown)


ryw:
Should we update README as well?

ryw:
Should we structure it so `airflow` is aliased to `dev` since `dev` is the future? Also, can we output a deprecation warning when ppl use `airflow` command?

cmd/airflow.go
```diff
@@ -33,13 +33,22 @@ var (
...
-astro airflow run create_user -r Admin -u admin -e admin@example.com -f admin -l user -p admin
+astro dev run create_user -r Admin -u admin -e admin@example.com -f admin -l user -p admin
 `
 
 	airflowRootCmd = &cobra.Command{
-		Use:   "airflow",
-		Short: "Manage airflow projects and deployments",
-		Long:  "Airflow projects are a single top-level directory which represents a single production Airflow deployment",
+		Use:     "airflow",
+		Aliases: []string{"a", "dev"},
```
schnie: @andriisoldatenko maybe remove the `dev` alias here.

Issue #254, opened 2019-07-16
Update to use new AuthConfig/AuthProvider schema


schnie:
We made some recent changes to the schema in Houston around AuthConfig. Looks like we need to upgrade the CLI to use it.

https://github.com/astronomer/houston-api/blob/master/src/schema.graphql#L52-L63

schnie:
@andriisoldatenko looks like we have to update the code to look at the list of providers instead of the top-level `googleEnabled`, `githubEnabled` and `auth0Enabled` flags.

andriisoldatenko:
cc @ashb

PR #255, opened 2019-07-16
update authConfig schema

andriisoldatenko:
fix https://github.com/astronomer/astro-cli/issues/254


andriisoldatenko:
hm.... 🤔 
cc @schnie

schnie:
We do release a matching version of everything so we could make the switch. Maybe at 1.0 we start supporting backward compatibility. I just hate to start accruing backward compatible code before we have something we feel is even stable to begin with.

andriisoldatenko:
cc @dimberman

PR #257, opened 2019-07-22
change error message to something more meaningful

andriisoldatenko:


user/user.go
```diff
@@ -28,7 +28,7 @@ func Create(email string) error {
 
 	resp, err := req.Do()
 	if err != nil {
-		return err
+		return errors.New("user creation is disabled")
```
schnie:
```suggestion
		return errors.New("User creation is disabled")
```

PR #259, opened 2019-07-26
Fix user queries

andriisoldatenko:
fix https://github.com/astronomer/houston-api/issues/77

### DIFF
--- a/houston/queries.go
+++ b/houston/queries.go
@@ -271,7 +271,7 @@ var (
 			id
 			label
 			description
-			systemUsers {
+			workspaceUsers {
 				id
 				username
 				roleBindings {
@@ -306,7 +306,7 @@ var (
 		  id
 		  label
 		  description
-		  systemUsers {
+		  workspaceUsers {
 			id
 			username
 		  }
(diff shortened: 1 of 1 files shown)


andriisoldatenko:
pls merge only together with https://github.com/astronomer/houston-api/pull/95

houston/queries.go
```diff
@@ -306,7 +306,7 @@ var (
 		  id
 		  label
 		  description
-		  users {
+		  systemUsers {
```
schnie: @andriisoldatenko I think we'll want the `workspaceUsers` query here. `systemUsers` is intended to be called via system admins and requires those permissions. `workspaceUsers` should be the equivalent of the old one.

andriisoldatenko: @schnie fixed, do you know where i can test it?
cc / @petedejoy

ashb:
I don't think this change is needed. The `workspace` type still has a `users` property on it, and that doesn't change even with Pete's PR95. It's just the top level `query { users(...) {}}` that changes, but I don't think we call that from the CLI.

schnie:
Oh, you're totally right @ashb. I'll close this out.

Issue #260, opened 2019-07-29
Remove .astro from .gitignore


andrewhharmon:
`.astro` is added to `.gitignore` by default. This causes any other users to get an error unless they run `astro airflow init`. There isn't anything sensitive in `.astro` so it shouldn't be in `.gitignore` by default

andrewhharmon:
merged. Closing issue

PR #261, opened 2019-07-29
remove .astro from .gitignore template #260

andrewhharmon:
Remove `.astro` from `.gitignore` so users don't have to run `astro airflow init` after cloning a repo


schnie:
Thanks @andrewhharmon!

PR #262, opened 2019-07-29
Fixing ignored filename

gonzalodiaz:
The file `airflow_settings.yaml` was intended to be ignored using `airflow_setttings.yaml`. This PR fixes that.


paolaperaza:
Many thanks, @gonzalodiaz ! @andriisoldatenko A current prospect caught this typo for us - can you take a look?

PR #268, opened 2019-08-13
Use Z flag on volumes in docker compose for airflow

nvander1:
Addresses #263. 

@andscoop @andriisoldatenko @benjamingregory @schnie

Here is the first draft of a solution to the issue. Do you think it would be more appropriate to enable this via a command line switch?

Thanks again to @vparekh94 for noticing the "." in the permissions on `ls -la`.


andriisoldatenko:
@nvander1 pls resolve conflicts

PR #271, opened 2019-08-15
Use Z flag on volumes in docker compose for airflow

andriisoldatenko:
Addresses #263.

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -59,6 +59,7 @@ type ComposeConfig struct {
 	AirflowHome          string
 	AirflowUser          string
 	AirflowWebserverPort string
+	MountLabel           string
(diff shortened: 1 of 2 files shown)


andriisoldatenko:
copy from https://github.com/astronomer/astro-cli/pull/268

andriisoldatenko:
tested on OSX seems works
```
      {
        "Type": "volume",
        "Name": "ttt973993_airflow_logs",
        "Source": "/var/lib/docker/volumes/ttt973993_airflow_logs/_data",
        "Destination": "/usr/local/airflow/logs",
        "Driver": "local",
        "Mode": "z",
        "RW": true,
        "Propagation": ""
      }
```

Issue #258, opened 2019-07-22
Bash completion


dstandish:
here is a start:

```
_astro_complete()
{
    local cur prev

    cur=${COMP_WORDS[COMP_CWORD]}
    prev=${COMP_WORDS[COMP_CWORD-1]}

	astro_commands='airflow auth cluster config deployment help service-account upgrade user version workspace'
	airflow_commands='deploy init kill logs ps start stop'
	auth_commands='login logout'
	cluster_commands='list switch'
	config_commands='get set'
	deployment_commands='create get delete'
	service_account_commands='create get delete'
	workspace_commands='create delete list switch update user'

    case ${COMP_CWORD} in
        1)
            COMPREPLY=($(compgen -W "$astro_commands" -- ${cur}))
            ;;
        2)
            case ${prev} in
                airflow)
                    COMPREPLY=($(compgen -W "$airflow_commands" -- ${cur}))
                    ;;
                auth)
                    COMPREPLY=($(compgen -W "$auth_commands" -- ${cur}))
                    ;;
                cluster)
                    COMPREPLY=($(compgen -W "$cluster_commands" -- ${cur}))
                    ;;
                config)
                    COMPREPLY=($(compgen -W "$config_commands" -- ${cur}))
                    ;;
                deployment)
                    COMPREPLY=($(compgen -W "$deployment_commands" -- ${cur}))
                    ;;
                'service-account')
                    COMPREPLY=($(compgen -W "$service_account_commands" -- ${cur}))
                    ;;
                workspace)
                    COMPREPLY=($(compgen -W "$workspace_commands" -- ${cur}))
                    ;;
            esac
            ;;
        *)
            COMPREPLY=()
            ;;
    esac
}

complete -F _astro_complete astro

```

ryw:
@andriisoldatenko have you looked at this?

PR #277, opened 2019-08-21
Add ability to generate bash completion to astro cli

andriisoldatenko:
TODO:
- [x] add tests based on shellcheck
- [x] add examples
- [x] check with zsh/bash

fix https://github.com/astronomer/astro-cli/issues/258

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -22,16 +22,22 @@ Can be sourced as such
 
 Bash users can as well save it to the file and copy it to:
 	/etc/bash_completion.d/
+
 Correct arguments for SHELL are: "bash" and "zsh".
 Notes:
+
 1) zsh completions requires zsh 5.2 or newer.
 	
 2) macOS users have to install bash-completion framework to utilize
 completion features. This can be done using homebrew:
 	brew install bash-completion
+
 Once installed, you must load bash_completion by adding following
 line to your .profile or .bashrc/.zshrc:
 	source $(brew --prefix)/etc/bash_completion
+
+3) For oh-my-zsh users make sure you have added to your ~/.zshrc
+	autoload -Uz compinit && compinit -C
 `
 
 var (
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


schnie:
@andriisoldatenko more to do here or ready to merge?

andriisoldatenko:
@schnie code is OK. need some help with practical QA.

cmd/completion.go
```diff
@@ -0,0 +1,267 @@
...
+
+var (
+	completionShells = map[string]func(out io.Writer, cmd *cobra.Command) error{
+		"bash": runCompletionBash,
+		"zsh":  runCompletionZsh,
+	}
+)
+
+func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
```
ashb: We should be able to use `ExactArgs(1)`  validator here to have cobra handle this for us

```go

	cmd := &cobra.Command{
		Use:   "completion SHELL",

		Short: "Generate autocompletions script for the specified shell (bash or zsh)",
		Long:  completionDesc,
		RunE: func(cmd *cobra.Command, args []string) error {
			return runCompletion(os.Stdout, cmd, args)
		},
		Args: cobra.ExactValidArgs(1),
		ValidArgs: shells,
	} 
```

cmd/completion.go
```diff
@@ -0,0 +1,267 @@
...
+
+func runCompletionBash(_ io.Writer, cmd *cobra.Command) error {
+	var buf bytes.Buffer
+	err := cmd.Root().GenBashCompletion(&buf)
+	if err != nil {
+		return fmt.Errorf("error while generating bash completion: %v", err)
+	}
+	// remove the command "completion" from auto-completion
+	code := buf.String()
+	code = strings.Replace(code, `commands+=("completion")`, "", -1)
```
ashb: Instead of needing to this we can set a `Hidden` property on the completion command object.

(Either globally which would also hide it from the help, or just when actually running the completion command)

cmd/completion.go
```diff
@@ -0,0 +1,267 @@
...
+
+	fmt.Print(code)
+	return nil
+}
+
+func runCompletionZsh(_ io.Writer, cmd *cobra.Command) error {
+	var buf bytes.Buffer
+
+	// zshHead is the header required to declare zsh completion
+	zshHead := "#compdef astro\n"
```
ashb: Can we not use `cmd.Root().GenZshCompletion(w)` here?

cmd/completion.go
andriisoldatenko: @ashb it's wrong, how can you run ./astro ---help?

cmd/completion.go
andriisoldatenko: fixed.

cmd/completion.go
andriisoldatenko: not sure if it works or not ... i checked
helm/kubectl/doctl ... let me try.

andriisoldatenko: fixed.

cmd/completion.go
andriisoldatenko:
```
astro is a command line interface for working with the Astronomer Platform.

Usage:
  astro [command]

Available Commands:
  auth            Manage astronomer identity
  cluster         Manage Astronomer EE clusters
  completion      Generate autocompletions script for the specified shell (bash or zsh)
  config          Manage astro project configurations
```

cmd/completion.go
```diff
@@ -0,0 +1,89 @@
...
+	for s := range completionShells {
+		shells = append(shells, s)
+	}
+
+	cmd := &cobra.Command{
+		Use:   "completion SHELL",
+		Short: "Generate autocompletions script for the specified shell (bash or zsh)",
+		Long:  completionLong,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCompletion(os.Stdout, cmd, args)
```
ashb:
```suggestion
			cmd.Hidden = true
			return runCompletion(os.Stdout, cmd, args)
```

This will hide `astro completion` from the completion options, but leave it shown in `astro --help`

cmd/completion.go
```diff
@@ -0,0 +1,89 @@
...
+		"bash": runCompletionBash,
+		"zsh":  runCompletionZsh,
+	}
+)
+
+func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
+	return completionShells[args[0]](out, cmd)
+}
+
+func runCompletionBash(_ io.Writer, cmd *cobra.Command) error {
```
ashb: We can use the io.Writer directly now we don't need to do any filtering.
```
func runCompletionBash(w io.Writer, cmd *cobra.Command) error {
	err := cmd.Root().GenBashCompletion(w)
	if err != nil {
		return fmt.Errorf("error while generating bash completion: %v", err)
	}

	return nil
}
```

ashb:
Tested this on OSX with bash and bash-completion installed:

```
source <(astro completion bash)
```

and 

```
source <(astro completion zsh)
```

on ZSH with `compinit` in my .zshrc (not using oh-my-zsh, but should work there too as I copied the completion set up form there)

:+1:

andriisoldatenko:
@schnie i tested using bash and Ubuntu/Debian also works.

cmd/completion.go
andriisoldatenko:
```
/ Hidden defines, if this command is hidden and should NOT show up in the list of available commands.
    Hidden bool

!!! @ashb 
```

andriisoldatenko:
oh-my-zsh note:
```
autoload -Uz compinit && compinit -C
source <(go run main.go completion zsh)
compdef _astro astro
```

cmd/completion.go
```diff
@@ -0,0 +1,82 @@
...
+This command can generate shell autocompletions. e.g.
+
+	$ astro completion bash
+
+Can be sourced as such
+
+	$ source <(astro completion bash)
+
+Bash users can as well save it to the file and copy it to:
+	/etc/bash_completion.d/
```
schnie:
```suggestion
	/etc/bash_completion.d/

```

schnie: Just inserting an extra line here.

cmd/completion.go
```diff
@@ -0,0 +1,82 @@
...
+Correct arguments for SHELL are: "bash" and "zsh".
+Notes:
+1) zsh completions requires zsh 5.2 or newer.
+	
+2) macOS users have to install bash-completion framework to utilize
+completion features. This can be done using homebrew:
+	brew install bash-completion
```
schnie:
```suggestion
	brew install bash-completion

```

schnie: Adding another separating newline here.

schnie:
@andriisoldatenko want to add the oh-my-zsh note to the description string of `astro completion`?

ashb:
> ```compdef _astro astro```

I didn't have to run that command on zsh for me.

cmd/completion.go
andriisoldatenko: fixed.

Issue #278, opened 2019-08-26
`astro serviceaccount create` now fails with Role required error from API


ashb:
```
$ astro sa cr -d cjzshq8sr00gx0b03bjg7bd4l --label wh-data-us-sdp-us-east-1-nonprod-service-account-cicd
Error: API error (400): {"errors":[{"message":"Field \"createServiceAccount\" argument \"role\" of type \"Role!\" is required but not provided.","locations":[{"line":8,"column":4}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"},"level":"ERROR","timestamp":"2019-08-26T14:30:28"}]}
```

Since we changed the houston API to now require the `Role` argument to the "createServiceAccount" mutation the CLI is now unable to create service accounts.  If the role is not specified it should default to the Workspace/Deployment Editor role as appropriate.

ashb:
Was thinking this was a quick fix in a 0.10.1, but we could just do the larger one, yes.

andriisoldatenko:
@schnie are you OK if i quick fix it in 0.10.1?

schnie:
Yea lets do it

PR #280, opened 2019-09-04
astro serviceaccount create now fails with Role required

andriisoldatenko:
fix https://github.com/astronomer/astro-cli/issues/278


cmd/serviceaccount.go
```diff
@@ -102,6 +103,15 @@ func getValidEntity() (string, string, error) {
 	return entityType, id, nil
 }
 
+func getValidRole() (string, error) {
+	var fullRole string
+	if len(workspaceId) > 0 {
+
+		fullRole = "WORKSPACE_ADMIN"
```
schnie: @andriisoldatenko to match Orbit, I think we want to use `WORKSPACE_VIEWER` as the default role, if the entity type is `WORKSPACE` and `DEPLOYMENT_VIEWER` if the entity type is `DEPLOYMENT`.

andriisoldatenko: @schnie 
Ashb suggested to have flag `--role=admin` and generate `WORKSPACE_ADMIN `
if `--role=viewer`  generate WORKSPACE_VIEWER? right?

schnie: That works for me

schnie: So we should essentially concatenate `entityType` and `role`?

andriisoldatenko: yes

andriisoldatenko: if user skip flag role, should it be viewer as default value?

andriisoldatenko:
practical QA:
```
astro sa cr -d ck05r3bor07h40d02y2hw4n4v --label abc -r viewer
DEPLOYMENT_VIEWER
 NAME     CATEGORY     ID                            APIKEY
 abc      default      ck05r3z1b07hq0d02b4tphx0o     ****************************

 Service account successfully created.
```

cmd/serviceaccount.go
schnie: yep

houston/queries.go
```diff
@@ -106,12 +106,14 @@ var (
...
 			$category: String
 			$entityType: EntityType!
+			$role: Role!
 		) {
 			createServiceAccount(
 				entityUuid: $entityId
 				label: $label
 				category: $category
 				entityType: $entityType
+                role: $role
```
ashb: Tabs vs spaces/inconsistent indenting here.

andriisoldatenko: fixed

andriisoldatenko:
@schnie @ashb PTAL

andriisoldatenko:
@ashb 100% agree, i would like to refactor during astronomer/astronomer#449

Issue #133, opened 2018-09-28
New release on a new branch causes entire commit history to be uploaded to changelog


andscoop:
Currently if a new release is cut on a `release-x` branch, the entire commit history is uploaded. This is probably an issue with the goreleaser script that we easily can fix.

andscoop:
Created a ticket with go-releaser to see if we will be able to resolved this without custom work on goreleaser.

ryw:
@andscoop can you link that issue here?

andscoop:
https://github.com/goreleaser/goreleaser/issues/817

ryw:
@schnie is this still an issue?

schnie:
Yes it is for us. We may need to update `goreleaser` or something though. Any idea @andriisoldatenko?

andriisoldatenko:
@ryw @schnie yes it's exists. Let me jump on it in next release and fix it.

andriisoldatenko:
@ryw @schnie done.
![image](https://user-images.githubusercontent.com/1374633/64454222-1afe1400-d0f3-11e9-82a6-24a02c6460e0.png)

schnie:
Thanks @andriisoldatenko!

PR #284, opened 2019-09-06
Fix fetch all tags

andriisoldatenko:


andriisoldatenko:
fixed https://github.com/astronomer/astro-cli/issues/133

PR #285, opened 2019-09-06
[WIP] migrate to go modules

andriisoldatenko:


schnie:
@andriisoldatenko what exactly are we doing here? This looks pretty big so I'm thinking we may hold off on merging until we v0.10.1 out the door.

andriisoldatenko:
it's just migration to go modules, it's not ready. 
will be ready in 1.0.0

Issue #247, opened 2019-07-10
`astro airflow deploy <non-existent deployment name>` returns "Error parsing reference..."


paolaperaza:
**Problem**

If a user runs `astro airflow deploy` followed by a non-existent release name (e.g. `dev` in this example ), the CLI returned the following error:

```
$ astro airflow deploy dev -f
Deploying: dev
dev/airflow
Building image...
Sending build context to Docker daemon  11.78kB
Step 1/1 : FROM astronomerinc/ap-airflow:0.7.5-1.10.1-onbuild
# Executing 5 build triggers
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> fac83ad0b2d8
Successfully built fac83ad0b2d8
Successfully tagged dev/airflow:latest
Error parsing reference: "registry.astronomer.cloud/dev/airflow:" is not a valid repository/tag: invalid reference format
Error: command 'docker tag dev/airflow:latest registry.astronomer.cloud/dev/airflow:' failed: failed to execute cmd: exit status 1
```

_Just_ running `astro airflow deploy` (no deployment name) gives this error:

```
$ astro deployment list
 NAME     RELEASE NAME     ASTRO     DEPLOYMENT ID     
$ astro airflow deploy -f
Error: No airflow deployments found
```

**Potential Solution**

Is there a way for us to better handle the initial error here, perhaps similarly to running `astro airflow deploy` without specifying a deployment name? Or the latter perhaps something like `no deployment specified` and the former something like `no deployment "dev" found`?

Deferring to you @schnie or @andriisoldatenko but figured I'd document in case that's something we want to change.

schnie:
@paolaperaza in your second example, do you actually have deployments on the cloud or is it wrong about not having any airflow deployments?

paolaperaza:
@schnie Yea should've clarified, but in this case I did not have _any_ deployments in that Workspace, so `No airflow deployments found` might be the right error there (and it's not just that one isn't specified since you'd otherwise be prompted to pick, but that none exist).

schnie:
Cool, yep we can probably catch the `docker push` error and output a cleaner, more meaningful message there instead. I'll need @andriisoldatenko to verify that though.

andriisoldatenko:
@schnie i think @sjmiller609 fix this issue ^

PR #287, opened 2019-09-10
if no deployments are found, we can't deploy

sjmiller609:
```
// Deploy pushes a new docker image
func Deploy(path, name, wsId string, prompt bool) error {

        ... omitted

	// Get Deployments from workspace ID
	deReq := houston.Request{
		Query:     houston.DeploymentsGetRequest,
		Variables: map[string]interface{}{"workspaceId": currentWorkspace.Id},
	}

	deResp, err := deReq.Do()
	if err != nil {
		return err
	}

	deployments := deResp.Data.GetDeployments

        ... omitted

	// NOTE: this part is working - when I do "astro deploy" (without specifying a deployment name) it does report error
	// Prompt user for deployment if no deployment passed in
	if len(name) == 0 || prompt {
		if len(deployments) == 0 {
			return errors.New(messages.HOUSTON_NO_DEPLOYMENTS_ERROR)
		}

		... omitted
	}

	// NOTE: Since we have no deployments, this will not run
        // that is why the tag is blank.
	nextTag := ""
	for _, deployment := range deployments {
		if deployment.ReleaseName == name {
			nextTag = deployment.DeploymentInfo.Next
		}
	}
        ... omitted
}
```
Now instead of this error:
```

Stevens-MacBook-Pro:airflow_dev steven$ astro deploy test
Deploying: test
test/airflow
Building image...
Sending build context to Docker daemon  13.31kB
Step 1/1 : FROM astronomerinc/ap-airflow:0.9.7-1.10.3-onbuild
# Executing 5 build triggers
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> Using cache
 ---> 625af0ec11fa
Successfully built 625af0ec11fa
Successfully tagged test/airflow:latest
Error parsing reference: "registry.staging.astronomer.io/test/airflow:" is not a valid repository/tag: invalid reference format
Error: command 'docker tag test/airflow:latest registry.staging.astronomer.io/test/airflow:' failed: failed to execute cmd: exit status 1
```

we get this error:
```
Stevens-MacBook-Pro:airflow_dev steven$ astro deploy test
Error: No airflow deployments found
```

listing deployments remains to be solved.


airflow/docker.go
```diff
@@ -532,11 +532,12 @@ func Deploy(path, name, wsId string, prompt bool) error {
 		name = config.CFG.ProjectDeployment.GetProjectString()
 	}
 
+	if len(deployments) == 0 {
+		return errors.New(messages.HOUSTON_NO_DEPLOYMENTS_ERROR)
+	}
+
```
sjmiller609: this doesn't fix our current problem, but it fixes the client side error message to be correct in the case of houston not returning a list of deployments. i.e. if a user had a workspace with no deployments, then tried to deploy using a name, ```astro deploy this-deployment-was-not-created``` they would find the same issue. this is an informative error message instead.

.gitignore
```diff
@@ -9,4 +9,5 @@ requirements.txt
 # IDEs
 .idea/
 .vscode/
-.DS_Store
\ No newline at end of file
+.DS_Store
+explain.md
```
andriisoldatenko: what is explain.md?

andriisoldatenko:
https://github.com/astronomer/astro-cli/issues ?

.gitignore
sjmiller609: I can remove it. just a file I have locally

sjmiller609:
oh what do you know? https://github.com/astronomer/astro-cli/issues/247

PR #290, opened 2019-09-11
Fix error message when deployment name

andriisoldatenko:


airflow/docker.go
```diff
@@ -532,6 +542,10 @@ func Deploy(path, name, wsId string, prompt bool) error {
 		name = config.CFG.ProjectDeployment.GetProjectString()
 	}
 
+	if len(name) != 0 && !deploymentNameExists(name, deployments) {
+		return errors.New(messages.HOUSTON_DEPLOYMENT_NAME_ERROR)
```
sjmiller609: why does there need to be a deployment to run locally?

andriisoldatenko:
fix https://github.com/astronomer/astro-cli/pull/287
fix https://github.com/astronomer/astro-cli/issues/247

schnie:
LGTM

messages/messages.go
```diff
@@ -52,6 +52,7 @@ var (
 	HOUSTON_DEPLOYMENT_HEADER        = "Authenticated to %s \n\n"
 	HOUSTON_DEPLOYING_PROMPT         = "Deploying: %s\n"
 	HOUSTON_NO_DEPLOYMENTS_ERROR     = "No airflow deployments found"
+	HOUSTON_DEPLOYMENT_NAME_ERROR    = "Please specify correct deployment name"
```
sjmiller609: Maybe something like
"Please specify a deployment that exists at link/deployments" or
"Please specify one of the following deployments:
deploy1
deploy2
...
" if there are some or otherwise "Please create a deployment in Astronomer before deploying"?

PR #297, opened 2019-09-19
Bump airflow version astronomer/issues#377

ianstanton:
* Bump airflow version to `1.10.5`
* I did not know if I needed to bump `acceptableAirflowVersions` in `cmd/airflow.go` but saw it bumped in a previous PR https://github.com/astronomer/astro-cli/commit/70374e90a1304024cb5f86757b3976b445777120#diff-036da75ff930f37d0e7765ce446cf663. It doesn't looked like it was bumped with `1.10.4`


ianstanton:
Related to https://github.com/astronomer/issues/issues/377

PR #298, opened 2019-09-24
Add integration tests

andriisoldatenko:
Summary:
- refactored Global XxxCmd variables to constructors `newXxxCmd(*args)`
- created small utils to mock filesystem calls
- add fake config
- add Test Houston Client to mock responses from Houston
- add simple e2e tests


cmd/airflow.go
```diff
@@ -121,56 +183,7 @@ astro dev run create_user -r Admin -u admin -e admin@example.com -f admin -l use
 		Example:            RunExample,
 		DisableFlagParsing: true,
 	}
-)
-
-func init() {
```
ashb: Not that I object, but what is the reason for changing how the cli command objects are built?

cmd/root_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRootCommand(t *testing.T) {
+	fs := afero.NewMemMapFs()
+	configYaml, err := ioutil.ReadFile("./testdata/config.yaml")
+	err = afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
+	config.InitConfig(fs)
+	expectedOut := `astro is a command line interface for working with the Astronomer Platform.
```
ashb: This test seems like it's going to be overly fragile - almost any change to the command would need a matching change in the test file -- i.e. I don't think having this test here gives us much value.

cmd/workspace_test.go
```diff
@@ -0,0 +1,63 @@
...
+	testClient.HTTPClient.Transport = RoundTripFunc(fn)
+	return testClient
+}
+
+func TestWorkspaceList(t *testing.T) {
+	fs := afero.NewMemMapFs()
+	configYaml, err := ioutil.ReadFile("./testdata/config.yaml")
+	err = afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
+	config.InitConfig(fs)
+	expectedOut := " NAME           ID                            \n\x1b[1;32m airflow        ck05r3bor07h40d02y2hw4n4v     \x1b[0m\n airflow123     XXXXXXXXXXXXXXX               \n"
```
ashb: COuld you format this over multiple lines too so it's a bit easier to read?

```suggestion
	expectedOut := ` NAME           ID                            
\x1b[1;32m airflow        ck05r3bor07h40d02y2hw4n4v     \x1b[0m
 airflow123     XXXXXXXXXXXXXXX               
 `
```

(I forget if there's a way in Go stdlib to "de-dent" things like python's textwrap.dedent)

houston/houston.go
```diff
@@ -36,6 +34,18 @@ type Request struct {
...
+		Headers: map[string]string{
+			"Accept": "application/json",
+		},
+	}
+
+	return api.Do(doOpts)
+}
+
 // Do (request) is a wrapper to more easily pass variables to a client.Do request
 func (r *Request) Do() (*Response, error) {
```
ashb: Do we need a "V2" here - (i.e. do we still use `Do()` anywhere?) If we want to keep both then `DoWithClient` is probably a better name than `DoV2`


And `Do` could be implemented as `return DoWithClient(NewHoustonClient(httputil.NewHTTPClient()))`

cmd/airflow.go
andriisoldatenko: better to avoid `init` and `global vars` as much as we can: two reasons better testing and no magic behavior,

cmd/root_test.go
andriisoldatenko: yes and no, only when you add new command or remove you must update this test.

cmd/workspace_test.go
andriisoldatenko: best i can do:
```
	expectedOut := " NAME           ID                            \n" +
		"\x1b[1;32m airflow        ck05r3bor07h40d02y2hw4n4v     \x1b[0m\n " +
		"airflow123     XXXXXXXXXXXXXXX               \n"
```

houston/houston.go
andriisoldatenko: done.

cmd/completion_test.go
```diff
@@ -0,0 +1,27 @@
...
+	expectedOut := `Error: accepts 1 arg(s), received 0
+Usage:
+  astro completion SHELL [flags]
+
+Flags:
+  -h, --help   help for completion
+
+`
+	_, output, err := executeCommandC(nil, fs, "completion")
+	assert.Equal(t, expectedOut, output, err)
```
ashb: When called with not enough arguments like this the exit code on the shell is 1 - how do we test that here?

cmd/version_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVersionRootCommand(t *testing.T) {
+	fs := afero.NewMemMapFs()
+	configYaml, err := ioutil.ReadFile("./testdata/config.yaml")
+	err = afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
+	config.InitConfig(fs)
+	// Locally we are not support version, that's why we see err
+	expectedOut := "Error: Astronomer CLI version is not valid\n"
```
ashb: Shouldn't this message be on stderr, not stdout?

ashb: Random thought - rather than passing `client *houston.Client, out io.Writer, error io.Writer` (as it would have to become to have error messages captured) to all the functions should we instead use [context](https://golang.org/pkg/context/) (I'm not too familiar with using contexts - i know it is "normally" used for http request serving for instance, so I'm not sure if it makes sense here)

andriisoldatenko: @ashb thanks for review!
IMO context here is a little bit overkill, it's for another set of problems. I checked helm/docker/k8s/doctl and this approach has been inspired by common usage. I prefer 1 step in time, instead of global vars and inits, now we have aka cmd factory to build commands and can test and control client/output. Yes some parts of old code still looks strange, but it's point to improve in future, create extra struct type to hide this details:
```
type SomeDetails struct {
    client Client
    out os.Writer
    ...
}
someD := NewSomeDetails()
// to reduce number of args
newWorkspaceCmd(someD) 
```

cmd/completion_test.go
andriisoldatenko: it's fake execution, not sure that we can capture it here.

schnie:
@andriisoldatenko I pulled this branch down and ran `make integration_test` and got this:

```
Error reading config in home dir: open /home/schnie/.astro/config.yaml: file does not exist
Error reading config in home dir: open /home/schnie/.astro/config.yaml: file does not exist
Error reading config in home dir: open /home/schnie/.astro/config.yaml: file does not exist
--- FAIL: TestExampleTestSuite (0.00s)
    --- FAIL: TestExampleTestSuite/TestCreateUser (0.00s)
        e2e_test.go:29: 
                Error Trace:    e2e_test.go:29
                Error:          "Error: User creation is disabled
                                " does not contain "Successfully created user"
                Test:           TestExampleTestSuite/TestCreateUser
                Messages:       Failed to create a user%!(EXTRA *errors.errorString=User creation is disabled)
    --- FAIL: TestExampleTestSuite/TestCreateWorkspace (0.00s)
        e2e_test.go:35: 
                Error Trace:    e2e_test.go:35
                Error:          Received unexpected error:
                                No context set, have you authenticated to a cluster?
                Test:           TestExampleTestSuite/TestCreateWorkspace
    --- FAIL: TestExampleTestSuite/TestWorkspaceListCommand (0.00s)
        e2e_test.go:40: 
                Error Trace:    e2e_test.go:40
                Error:          Received unexpected error:
                                No context set, have you authenticated to a cluster?
                Test:           TestExampleTestSuite/TestWorkspaceListCommand
        e2e_test.go:41: 
                Error Trace:    e2e_test.go:41
                Error:          "Error: No context set, have you authenticated to a cluster?
                                " does not contain "test-workspace"
                Test:           TestExampleTestSuite/TestWorkspaceListCommand
                Messages:       Failed to run workspace list%!(EXTRA *errors.errorString=No context set, have you authenticated to a cluster?)
FAIL
FAIL    github.com/astronomer/astro-cli/cmd     0.021s
?       github.com/astronomer/astro-cli/config  [no test files]
ok      github.com/astronomer/astro-cli/deployment      0.036s
ok      github.com/astronomer/astro-cli/docker  0.134s
?       github.com/astronomer/astro-cli/gendocs [no test files]
ok      github.com/astronomer/astro-cli/houston 0.004s
?       github.com/astronomer/astro-cli/logs    [no test files]
?       github.com/astronomer/astro-cli/messages        [no test files]
?       github.com/astronomer/astro-cli/pkg/fileutil    [no test files]
?       github.com/astronomer/astro-cli/pkg/git [no test files]
?       github.com/astronomer/astro-cli/pkg/github      [no test files]
?       github.com/astronomer/astro-cli/pkg/httputil    [no test files]
?       github.com/astronomer/astro-cli/pkg/input       [no test files]
ok      github.com/astronomer/astro-cli/pkg/printutil   0.032s [no tests to run]
?       github.com/astronomer/astro-cli/pkg/testing     [no test files]
ok      github.com/astronomer/astro-cli/serviceaccount  0.012s [no tests to run]
?       github.com/astronomer/astro-cli/settings        [no test files]
?       github.com/astronomer/astro-cli/user    [no test files]
?       github.com/astronomer/astro-cli/version [no test files]
?       github.com/astronomer/astro-cli/workspace       [no test files]
FAIL
make: *** [Makefile:23: integration_test] Error 1
```

What's the recommended way to get these to run? How will they run in drone?

schnie:
@andriisoldatenko can we do the same thing we're doing in orbit?
https://github.com/astronomer/orbit-ui/blob/master/.drone.yml#L23-L106

andriisoldatenko:
@schnie @ashb Please take another look!

.drone.yml
```diff
@@ -9,13 +9,79 @@ pipeline:
...
+      ORBIT__PORT: "8080"
+      DATABASE__CONNECTION: "postgres://postgres:postgres@postgres:5432/postgres"
+      PUBLIC_SIGNUPS: "true"
+      EMAIL_CONFIRMATION: "false"
+      EMAIL__ENABLED: "false"
+      JWT__PASSPHRASE: "secretpassword"
+      DEPLOYMENTS__DATABASE__CONNECTION: "postgres://postgres:postgres@postgres:5432/postgres"
+      DEPLOYMENTS__DATABASE__ENABLED: "true"
+      DEPLOYMENTS__LOG_HELM_VALUES: "true"
+      HELM__RELEASE_VERSION: "0.7.5"
```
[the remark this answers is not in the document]
andriisoldatenko: it's copy/paste from houston drone.yml config. @schnie @samblackk can i remove it?

.drone.yml
```diff
@@ -9,13 +9,79 @@ pipeline:
...
+    when:
+      event: push
+
   # Test the code
   test:
-    image: asoldatenko/dnd-golang
+    image: asoldatenko/dnd-golang:19.03
+    secrets: [ codecov_token ]
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
```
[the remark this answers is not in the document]
andriisoldatenko: yes we need docker cli available, because we have a test where we execute docker command. https://github.com/astronomer/astro-cli/blob/master/docker/docker_test.go#L6-L10

cmd/e2e_test.go
```diff
@@ -0,0 +1,71 @@
...
+	fmt.Println(string(config))
+	output := new(bytes.Buffer)
+	err := auth.Login(suite.TestDomain, true, suite.TestEmail, suite.TestPassword, suite.Client, output)
+	assert.NoError(suite.T(), err)
+	_, out, err := executeCommandC(suite.Client, "workspace", "create", suite.TestWorkspace)
+	assert.NoError(suite.T(), err)
+	expectedOut := "Successfully created workspace"
+	assert.Contains(suite.T(), out, expectedOut)
+}
+
```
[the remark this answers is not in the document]
andriisoldatenko: let's add more tests in new tickets, i would like to keep PR as smaller as we can

.drone.yml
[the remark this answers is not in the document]
sjmiller609: The use case for dind is to run docker within docker. It's not intended for running sibiling-style docker containers where a container mounts the socket from the host. You can run docker CLI just fine inside a normal container if docker is installed within and the docker socket is mounted. For example https://github.com/astronomer/ap-build/blob/91dee715aac7239fa1fdcfa55c4838fbac089694/Dockerfile#L25

.drone.yml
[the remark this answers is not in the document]
schnie: Not sure if it can be removed, but it should only matter for stamping out a new Airflow deployment. This determines the version its going to use. If we are not actually `helm install`ing, the it probably doesn't matter.

cmd/e2e_test.go
```diff
@@ -0,0 +1,71 @@
...
+//func (suite *IntegrationTestSuite) TestWorkspaceListCommand() {
+//	output, err := executeCommand("workspace", "list")
+//	assert.NoError(suite.T(), err)
+//	assert.Contains(suite.T(), output, "test-workspace", "Failed to run workspace list", err)
+//}
+
+// In order for 'go test' to run this suite, we need to create
+// a normal test function and pass our suite to suite.Run
```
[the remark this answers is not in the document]
andriisoldatenko: yes the goal is to run all tests using `go test`

.drone.yml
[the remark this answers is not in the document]
andriisoldatenko: good catch, 
if it's okey i'll change it in separate PR, because we added it not in this PR.

cmd/e2e_test.go
```diff
@@ -0,0 +1,71 @@
...
+	assert.NoError(suite.T(), err)
+	expectedOut := "Successfully created workspace"
+	assert.Contains(suite.T(), out, expectedOut)
+}
+
+//func (suite *IntegrationTestSuite) TestWorkspaceListCommand() {
+//	output, err := executeCommand("workspace", "list")
+//	assert.NoError(suite.T(), err)
+//	assert.Contains(suite.T(), output, "test-workspace", "Failed to run workspace list", err)
+//}
```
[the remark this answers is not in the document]
andriisoldatenko: removed.

auth/auth_test.go
```diff
@@ -0,0 +1 @@
+package auth
```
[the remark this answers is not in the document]
andriisoldatenko: removed

serviceaccount/serviceaccount_test.go
```diff
@@ -0,0 +1,34 @@
...
+//			StatusCode: 200,
+//			Body:       ioutil.NopCloser(bytes.NewBufferString(okResponse)),
+//			Header:     make(http.Header),
+//		}
+//	})
+//	api := houston.NewHoustonClient(client)
+//	out := new(bytes.Buffer)
+//	err = Get("entity", "id", api, out)
+//	assert.Equal(t, expectedOut, out, err)
+//}
```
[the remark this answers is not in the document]
andriisoldatenko: deleted

pkg/printutil/printutil_test.go
```diff
@@ -0,0 +1,3 @@
+package printutil
+
+
```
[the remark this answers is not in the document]
andriisoldatenko: deleted

deployment/deployment_test.go
```diff
@@ -0,0 +1,39 @@
...
+//	client := testUtils.NewTestClient(func(req *http.Request) *http.Response {
+//		return &http.Response{
+//			StatusCode: 200,
+//			Body:       ioutil.NopCloser(bytes.NewBufferString(okResponse)),
+//			Header:     make(http.Header),
+//		}
+//	})
+//	api := houston.NewHoustonClient(client)
+//	err = List("ck05r3bor07h40d02y2hw4n4v", true, api, output)
+//	assert.Equal(t, expectedOut, output.String(), err)
```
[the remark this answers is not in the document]
andriisoldatenko: removed.

cmd/e2e_test.go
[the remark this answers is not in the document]
andriisoldatenko: https://github.com/astronomer/issues/issues/462

andriisoldatenko:
@schnie need your final approve

cmd/e2e_test.go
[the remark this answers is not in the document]
schnie: @andriisoldatenko is there a way to separate unit tests from integration tests using the `go test` command?

In houston we can pass a regex that will only run certain suites of tests (files). Something like:

`jest unit` vs `jest int`, which would run `test.unit.js` or `test.int.js` respectively. How could we do something like that here?

andriisoldatenko:
@ashb @schnie we can't control exit code cause it's function and we can compare return string.
I can compare ONLY part of the string...

cmd/auth_test.go
```diff
@@ -0,0 +1,33 @@
...
+Available Commands:
+  login       Login to Astronomer services
+  logout      Logout of Astronomer services
+
+Flags:
+  -h, --help   help for auth
+
+Use "astro auth [command] --help" for more information about a command.
+`
+	_, output, err := executeCommandC(nil, "auth")
```
andriisoldatenko: @ashb @schnie 
For instance i can check only:
`assert.NoError(err)`, and skip comparing with full output of command or check something like:
`assert.NoError(err)`
`assert.Contains(output, "astro auth")` if somebody forgot to register command.

ashb: > if somebody forgot to register command
That feels to me like it is something that is more easily caught at the "unit" test level/by inspecting the created command objects, or just by calling the sub commands we are interested in.

I can't think of many circumstances where a test like this one as it is currently would catch a bug, but we would have to "keep it in sync" with help text and sub commands.

I think `assert.NoError(err)` would be good enough for cases like this.

(What is the `_` return value we are ignoring btw? Isn't this also what `executeCommand()` func does?)

andriisoldatenko: >(What is the _ return value we are ignoring btw? Isn't this also what executeCommand() func does?)
correct, i'll fix asap.

andriisoldatenko: fixed in 0c08289

cluster/cluster_test.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func TestClusterDoesNotExists(t *testing.T) {
+	// fake filesystem
+	fs := afero.NewMemMapFs()
+	// we have only localhost domain in test config
+	configYaml := testUtils.NewTestConfig()
+	err := afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
+	config.InitConfig(fs)
+	assert.Equal(t, Exists("localhost123"), false, err)
```
ashb: I'm confused by two things here:

1. where the `Exists()` fn is coming from - i don't see any wildcard imports or this defined anywhere?
2. what this is testing (my testify is rusty): Is that that localhost123 appears in the error? What's the `false` parameter for?

pkg/testing/testing.go
```diff
@@ -0,0 +1,53 @@
...
+contexts:
+  HOUSTON_HOST:
+    domain: HOUSTON_HOST
+    token: token
+    last_used_workspace: ck05r3bor07h40d02y2hw4n4v
+    workspace: ck05r3bor07h40d02y2hw4n4v
+`)
+	config := strings.ReplaceAll(string(configRaw), "HOUSTON_HOST", houstonHost)
+	return []byte(config)
+}
```
schnie: @andriisoldatenko need a newline here

cmd/e2e_test.go
```diff
@@ -0,0 +1,63 @@
...
+	suite.TestDomain = testUtils.GetEnv("HOUSTON_HOST", "localhost")
+	suite.TestEmail = fmt.Sprintf("test%d@astronomer.io", rand.Intn(100))
+	suite.TestPassword = "pass"
+	suite.TestWorkspace = "test-workspace"
+	user.Create(suite.TestEmail, suite.TestPassword, suite.Client, new(bytes.Buffer))
+}
+
+func (suite *IntegrationTestSuite) Test2CreateWorkspace() {
+	config, _ := afero.ReadFile(suite.Fs, config.HomeConfigFile)
+	fmt.Println(string(config))
```
ashb: These two lines can go I think? Left over debug?

cmd/user_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/astronomer/astro-cli/config"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUserRootCommand(t *testing.T) {
+	fs := afero.NewMemMapFs()
+	configYaml := testUtil.NewTestConfig()
+	err := afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
+	config.InitConfig(fs)
```
ashb: These 4 lines appear in all of our tests.

Do you think it's worth collecting these in to a single fn in testUtil, say:

```go
func InitConfigForTests() {
	fs := afero.NewMemMapFs()
	configYaml := NewTestConfig()
	err := afero.WriteFile(fs, config.HomeConfigFile, []byte(configYaml), 0777)
	config.InitConfig(fs)
	if err != nil {
		panic(err); // Shouldn't happen
	}
}
```

README.md
```diff
@@ -176,6 +177,15 @@ local:
   orbit: http://localhost:5000
 ```
 
+### Run tests
+
+To run unit-tests you can run:
+
+```bash
+make test
```
ashb: Missing closing fence here.

cluster/cluster_test.go
andriisoldatenko: 1) in go you can use functions from same package in different files:
`cluster/cluster.go:57` https://github.com/astronomer/astro-cli/blob/66e4019ec93551b612c06acfb9682c5bf80ceed7/cluster/cluster.go#L57
2) what about?
```
	// Check that we don't have localhost123 in test config from testUtils.NewTestConfig()
	assert.Equal(t, Exists("localhost123"), false, err)
```

pkg/testing/testing.go
andriisoldatenko: fixed.

cmd/e2e_test.go
andriisoldatenko: done in de413ed.

README.md
andriisoldatenko: fixed.

cmd/e2e_test.go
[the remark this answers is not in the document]
andriisoldatenko: it's tricky: i tried build tags, doesn't work properly, another idea just use go -run RegExp, where IntegrationTests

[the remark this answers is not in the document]
schnie: If that works, should we make two separate `make` commands to wrap the two commands?

schnie:
I think I'm good :+1: Looks like we created a conflict though.

andriisoldatenko:
@ashb @schnie resolved conflicts, squashed commits.

PR #305, opened 2019-10-28
Start maintaining CHANGELOG

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/355


schnie:
Nice, @andriisoldatenko can you copy the same format from astronomer/astronomer just for consistency?

schnie:
@andriisoldatenko looks good, should we just remove the older releases and start now with just the v0.10.3 section?

andriisoldatenko:
@schnie PTAL

PR #306, opened 2019-11-18
remove tabs

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/572


airflow/include/exampleplugin.go
```diff
@@ -35,7 +35,7 @@ class AstroLinksPlugin(AirflowPlugin):
     macros = []
     admin_views = []
     menu_links = [airflow_plugins_ml, astro_docs_ml, astro_guides_ml]
-		appbuilder_views = []
+        appbuilder_views = []
```
ashb:
```suggestion
    appbuilder_views = []
```

otherwise python will complain about indentation.

andriisoldatenko: done.

schnie:
Is this fixed now?

PR #307, opened 2019-11-19
add airflow upgradedb

andriisoldatenko:


schnie:
@andriisoldatenko nice fix. Does this work in practice consistently? I'm wondering what would happen if the webserver boots up first and there is no schema.

Should we just add it on both webserver / scheduler to be safe? It'd be closer to the previous behavior I guess.

andriisoldatenko:
@schnie webserver depends on `scheduler.`
```
  webserver:
    image: {{ .AirflowImage }}
    command: >
      bash -c "airflow create_user -r Admin -u admin -e admin@example.com -f admin -l user -p admin && airflow webserver"
    restart: unless-stopped
    networks:
      - airflow
    user: {{ .AirflowUser }}
    labels:
      io.astronomer.docker: "true"
      io.astronomer.docker.cli: "true"
      io.astronomer.docker.component: "airflow-webserver"
    depends_on:
      - scheduler
      - postgres
```

andriisoldatenko:
or do you mean this:
>depends_on does not wait for db and redis to be “ready” before starting web - only until they have been started. If you need to wait for a service to be ready, see Controlling startup order for more on this problem and strategies for solving it.

schnie:
Oh that might be good then!

PR #312, opened 2019-12-05
Fix readme to match cloud 2 standards

petedejoy:
This PR edits the language in our ReadMe to reflect how CLI versioning should work with new Cloud.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Git Commit: f5cdab8f832da3c6184a7ac167b491e3bac3c022
 #### Troubleshooting
 
 1. Make sure you go through instruction to install Docker on windows properly https://docs.docker.com/docker-for-windows/install/
-2. Make sure you enabled Hyper-V, it's requre to Docker and Linux Containers, also plesae review this document
+2. Make sure you enabled Hyper-V, it's required for Docker and Linux Containers, also please review this document
 https://docs.docker.com/docker-for-windows/troubleshoot/
 
 ## Getting Started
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,31 +31,30 @@ Use "astro [command] --help" for more information about a command.
...
+To install a previous version of Astronomer, you can add the tag at the end of your `curl` command via the following syntax:
+
 ```
 curl -sSL https://install.astronomer.io | sudo bash -s -- [TAGNAME]
 ```
 
 ie:
 ```
-curl -sSL https://install.astronomer.io | sudo bash -s -- v0.3.1
+curl -sSL https://install.astronomer.io | sudo bash -s -- v0.7.5
```
andriisoldatenko: what do you think about 0.10.3?

schnie: Actually lets just drop the version here so it installs the latest stable by default. And we should change the url to be the new prod url for this.

petedejoy: So this is a section specifically for "Previous Versions". It comes after the "Latest Version" section which installs the latest stable by default.

What's the new prod URL?

petedejoy:
Can we get this merged @andriisoldatenko @schnie? Know you're juggling the release right now but pretty confident it's an improvement on what we're showing currently.

README.md
schnie: That's still it. There's always one launched with the platform that installs the platform version, but this on is separate and hosted on its own. It'll always do the latest published.

PR #314, opened 2019-12-11
Implement warning messages during airflow deploy

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/625
b) Warn users on deploy if they are pushing latest or latest-onbuild, that they should replace it with the latest supported tag from Houston. - "You are pushing an image using the latest tag. This is not recommended. Please base your image off <defaultTagFromHouston>"

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -596,19 +596,25 @@ func Deploy(path, name, wsId string, prompt bool) error {
 		return errors.Wrapf(err, "failed to parse dockerfile: %s", filepath.Join(path, "Dockerfile"))
 	}
 
-	tag := docker.GetTagFromParsedFile(cmds)
+	image, tag := docker.GetImageTagFromParsedFile(cmds)
+	if image != messages.VALID_DOCKERFILE_BASE_IMAGE {
+		fmt.Printf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE)
+	}
 
 	// Get valid image tags for platform using Deployment Info request
 	diReq := houston.Request{
 		Query: houston.DeploymentInfoRequest,
 	}
 
-	deResp, err := diReq.Do()
+	diResp, err := diReq.Do()
 	if err != nil {
 		return err
 	}
 
-	deployments := deResp.Data.DeploymentConfig.AirflowImages
+	if !diResp.Data.DeploymentConfig.IsValidTag(tag) {
+		validTags := strings.Join(diResp.Data.DeploymentConfig.GetValidTags(), ",")
+		fmt.Printf("WARNING!!! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s\n", tag, validTags)
+	}
 
 	err = imageBuild(path, deployImage)
 	if err != nil {
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -598,7 +598,7 @@ func Deploy(path, name, wsId string, prompt bool) error {
 
 	image, tag := docker.GetImageTagFromParsedFile(cmds)
 	if config.CFG.ShowWarnings.GetBool() && image != messages.VALID_DOCKERFILE_BASE_IMAGE {
-		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
+		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
 		if !i {
 			fmt.Println("Cancelling deploy...")
 			os.Exit(1)
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -313,7 +313,7 @@ func Logs(airflowHome string, webserver, scheduler, follow bool) error {
 	s := make([]string, 0)
 
 	// Get project name from config
-	projectName := config.CFG.ProjectName.GetString()
+	projectName, err := projectNameUnique()
 
 	// Create libcompose project
 	project, err := createProject(projectName, airflowHome, "")
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -598,7 +598,7 @@ func Deploy(path, name, wsId string, prompt bool) error {
 
 	image, tag := docker.GetImageTagFromParsedFile(cmds)
 	if config.CFG.ShowWarnings.GetBool() && image != messages.VALID_DOCKERFILE_BASE_IMAGE {
-		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
+		i, _ := input.InputConfirm(fmt.Sprintf("WARNING! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
 		if !i {
 			fmt.Println("Cancelling deploy...")
 			os.Exit(1)
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -613,7 +613,11 @@ func Deploy(path, name, wsId string, prompt bool) error {
 
 	if !diResp.Data.DeploymentConfig.IsValidTag(tag) {
 		validTags := strings.Join(diResp.Data.DeploymentConfig.GetValidTags(), ",")
-		fmt.Printf("WARNING!!! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s\n", tag, validTags)
+		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s. Are you sure you want to continue?\n", tag, validTags))
+		if !i {
+			fmt.Println("Cancelling deploy...")
+			os.Exit(1)
+		}
 	}
 
 	err = imageBuild(path, deployImage)
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -8,6 +8,7 @@ import (
 	"html/template"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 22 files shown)


airflow/docker.go
```diff
@@ -589,6 +590,32 @@ func Deploy(path, name, wsId string, prompt bool) error {
...
+	}
+
+	diResp, err := diReq.Do()
+	if err != nil {
+		return err
+	}
+
+	if !diResp.Data.DeploymentConfig.IsValidTag(tag) {
+		validTags := strings.Join(diResp.Data.DeploymentConfig.GetValidTags(), ",")
+		fmt.Printf("WARNING!!! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s\n", tag, validTags)
```
ashb: If it's not recommended we should ask for confirmation?

airflow/docker.go
```diff
@@ -589,6 +590,32 @@ func Deploy(path, name, wsId string, prompt bool) error {
...
 
+	// parse dockerfile
+	cmds, err := docker.ParseFile(filepath.Join(path, "Dockerfile"))
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse dockerfile: %s", filepath.Join(path, "Dockerfile"))
+	}
+
+	image, tag := docker.GetImageTagFromParsedFile(cmds)
+	if image != messages.VALID_DOCKERFILE_BASE_IMAGE {
+		fmt.Printf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE)
```
ashb: This is a little bit heavy handed of a message. I can imagine some of our users would create a new base image for thier company that have libraries/utils pre-installed in to the image, so we need a way of silencing this warning -- probably by something in the .astro project file?

docker/parse.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, IOError{err.Error()}
+	}
+	defer file.Close()
+
+	return ParseReader(file)
+}
+
+// Parse tag from parsed dockerfile:
+// e.g. FROM ubuntu:xenial returns "ubuntu", "xenial"
+func GetImageTagFromParsedFile(cmds []Command) (baseImage, tag string) {
```
ashb: So Dockerfiles can have multiple "targets" in an image -- how does this cope with such a file? For example check out 

https://github.com/astronomer/ap-airflow/blob/7ffb41911ae106b9e110b8bb0e276a089ac1162b/1.10.5/buster/Dockerfile#L18
https://github.com/astronomer/ap-airflow/blob/7ffb41911ae106b9e110b8bb0e276a089ac1162b/1.10.5/buster/Dockerfile#L90
https://github.com/astronomer/ap-airflow/blob/7ffb41911ae106b9e110b8bb0e276a089ac1162b/1.10.5/buster/Dockerfile#L134

airflow/docker.go
andriisoldatenko: @schnie what do you think?

docker/parse.go
andriisoldatenko: @ashb i don't think we are going to parse something complicated, the idea is to check and warn that you are using custom dockerfile and we are not guarantee if it works.

FYI we can parse more complicated Dockerfiles:
```
(dlv) p cmds
[]github.com/astronomer/astro-cli/docker.Command len: 2, cap: 2, [
        {
                Cmd: "from",
                SubCmd: "",
                Json: false,
                Original: "FROM astronomerinc/ap-airflow:latest-onbuild as airflow-apt-deps",
                StartLine: 1,
                EndLine: 1,
                Flags: []string len: 0, cap: 0, [],
                Value: []string len: 3, cap: 4, [
                        "astronomerinc/ap-airflow:latest-onbuild",
                        "as",
                        "airflow-apt-deps",
                ],},
        {
                Cmd: "from",
                SubCmd: "",
                Json: false,
                Original: "FROM second:latest as main",
                StartLine: 2,
                EndLine: 2,
                Flags: []string len: 0, cap: 0, [],
                Value: []string len: 3, cap: 4, [
                        "second:latest",
                        "as",
                        "main",
                ],},
]
(dlv)
```

airflow/docker.go
andriisoldatenko: may be rephrase message to:
`You are using custom image name... on your own risk? `

schnie: Hmm, it may come down to how we plan to deal with the airflow version on the server side. If a user wants to silence this warning, they'd also want to turn off any (future) guardrails that we have in place in Houston. 

Was chatting with @kaxil about this earlier today some. Feels like we're on the completely "loose" end of the spectrum now, but we could build up some guardrails that key on our image labels, or we can go all out and completely manage the image ourselves and give the users less control, but that scares me a little bit.

I like the guardrails approach we've discussed in the past, but just trying to connect the dots on how this would tie in.

airflow/docker.go
schnie: Yea, seems like they should have to manually agree to move forward.

airflow/docker.go
andriisoldatenko: My 5 cents:
Right now we have "Certified" airflow image and it's good to WARN for user if it uses something different on its own risk. 

To skip warning user can change his custom image and reuse authorized `-onbuild` one or propose pull request to add new image to `ap-airflow` repo.

airflow/docker.go
andriisoldatenko:
```
➜  t_123 astro-cli deploy sidereal-vacuum-8961
Deploying: sidereal-vacuum-8961
sidereal-vacuum-8961/airflow
Building image...
WARNING!!! You are using invalid image name 'ubuntu' in your Dockerfile, please replace to astronomerinc/ap-airflow
WARNING!!! You are going to push an image using the 'latest' tag. This is not recommended. Please base your image off to one of valid image tags: 1.10.5-alpine3.10-onbuild. Are you sure you want to continue?
 (y/n) n
Cancelling deploy...
```
and
```
➜  t_123 astro-cli deploy sidereal-vacuum-8961
Deploying: sidereal-vacuum-8961
sidereal-vacuum-8961/airflow
Building image...
WARNING!!! You are using invalid image name 'ubuntu' in your Dockerfile, please replace to astronomerinc/ap-airflow
WARNING!!! You are going to push an image using the 'latest' tag. This is not recommended. Please base your image off to one of valid image tags: 1.10.5-alpine3.10-onbuild. Are you sure you want to continue?
 (y/n) y
Sending build context to Docker daemon  10.75kB
Step 1/1 : FROM ubuntu:latest
 ---> 7698f282e524
Successfully built 7698f282e524
Successfully tagged sidereal-vacuum-8961/airflow:latest
Pushing image to Astronomer registry
The push refers to repository [registry.localhost/sidereal-vacuum-8961/airflow]
```
ok?

docker/parse.go
ashb: To handle this "properly" we'd have to follow the chain "backwards" from the last `FROM` in the file.

I suspect this is so rarely used, especially in one of our projects that we can not worry about it for now.

airflow/docker.go
ashb: I'm thinking of this sort of use case:

```
FROM astronomerinc/ap-airflow:1.10.5-1

RUN pip install a-big-slow-module
```

That is then `docker build -t my-co/ap-airflow-base`, and then the docker file for the astro project looks like

```
FROM my-co/ap-airflow-base
```

I think that should be supported, and _if configured_ not print any warnings. A pull request to `ap-airflow` isn't appropriate for this sort of "baseimage" as it's specific to one company, not all of our users.

andriisoldatenko: @ashb 
yeah, to cover this case i need to docker pull/read image meta info, which is overkill. Probably if we can introduce astro flag/ or config it would be ok for now.

ashb: Yeah, I agree it would be overkill. I think just having this warning disablely by a setting in the .astro/config.yml file would be the way to go.

(Also we don't have a confirm step here, just on the tag one? Was that on purpose?)

andriisoldatenko: @ash something like it?
```
cat ~/.astro/config.yaml
...
show_warnings: true
```
?

schnie: Yep, adding a config to suppress the warning to the project makes sense to me.

andriisoldatenko: done.

andriisoldatenko:
@ashb @schnie PTAL

airflow/docker.go
```diff
@@ -589,6 +590,40 @@ func Deploy(path, name, wsId string, prompt bool) error {
...
 
+	// parse dockerfile
+	cmds, err := docker.ParseFile(filepath.Join(path, "Dockerfile"))
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse dockerfile: %s", filepath.Join(path, "Dockerfile"))
+	}
+
+	image, tag := docker.GetImageTagFromParsedFile(cmds)
+	if config.CFG.ShowWarnings.GetBool() && image != messages.VALID_DOCKERFILE_BASE_IMAGE {
+		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
```
schnie:
```suggestion
		i, _ := input.InputConfirm(fmt.Sprintf("WARNING! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n", image, messages.VALID_DOCKERFILE_BASE_IMAGE))
```

airflow/docker.go
```diff
@@ -589,6 +590,40 @@ func Deploy(path, name, wsId string, prompt bool) error {
...
+	}
+
+	diResp, err := diReq.Do()
+	if err != nil {
+		return err
+	}
+
+	if config.CFG.ShowWarnings.GetBool() && !diResp.Data.DeploymentConfig.IsValidTag(tag) {
+		validTags := strings.Join(diResp.Data.DeploymentConfig.GetValidTags(), ",")
+		i, _ := input.InputConfirm(fmt.Sprintf("WARNING!!! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s. Are you sure you want to continue?\n", tag, validTags))
```
schnie:
```suggestion
		i, _ := input.InputConfirm(fmt.Sprintf("WARNING! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s. Are you sure you want to continue?\n", tag, validTags))
```

messages/messages.go
```diff
@@ -67,5 +67,8 @@ var (
 
 	SETTINGS_PATH = "Error looking for settings.yaml"
 
-	NA = "N/A"
+	NA                          = "N/A"
+	VALID_DOCKERFILE_BASE_IMAGE = "astronomerinc/ap-airflow"
+	WARNING_INVALID_IMAGE_NAME  = "WARNING! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n"
+	WARNING_INVALID_IMAGE_TAG   = "WARNING! You are going to push an image using the '%s' tag. This is not recommended. Please base your image off to one of valid image tags: %s. Are you sure you want to continue?\n"
```
schnie:
```suggestion
	WARNING_INVALID_IMAGE_TAG   = "WARNING! You are going to push an image using the '%s' tag. This is not recommended.\nPlease base your image off to one of valid image tags: %s. Are you sure you want to continue?\n"
```

messages/messages.go
```diff
@@ -67,5 +67,8 @@ var (
 
 	SETTINGS_PATH = "Error looking for settings.yaml"
 
-	NA = "N/A"
+	NA                          = "N/A"
+	VALID_DOCKERFILE_BASE_IMAGE = "astronomerinc/ap-airflow"
+	WARNING_INVALID_IMAGE_NAME  = "WARNING! You are using invalid image name '%s' in your Dockerfile, please replace to %s. Are you sure you want to continue?\n"
+	WARNING_INVALID_IMAGE_TAG   = "WARNING! You are going to push an image using the '%s' tag. This is not recommended.\nPlease base your image off to one of valid image tags: %s. Are you sure you want to continue?\n"
```
schnie:
```suggestion
	WARNING_INVALID_IMAGE_TAG   = "WARNING! You are about to push an image using the '%s' tag. This is not recommended.\nPlease base your image off to one of valid image tags: %s. Are you sure you want to continue?\n"
```

PR #315, opened 2019-12-12
update example dag file

andrewhharmon:
I've written an entirely new example dag to use when astro init is run. This one uses BashOperator, PythonOperator, DummyOperator, looping to create tasks, dag context manager and some other little hints for new users


andriisoldatenko:
practically LGTM
image

PR #316, opened 2020-01-14
Add automated github release

sjmiller609:
If you merge to a release-X.Y branch, and click the manual approval button in Circle CI workflow, then this will find the most recent github release on this repository with major, minor = X, Y and publish a new release with the patch number one greater. If it does not find X, Y (i.e. new minor version), then it will use patch number 0. The release will automatically generate release notes based on the commits since the most recent tag. The automation will not overwrite existing releases.

this requires the environment variable GITHUB_TOKEN


.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
...
+        type: string
+        default:  "$NEXT_TAG"
+      ghr_version:
+        type: string
+        default:  "0.13.0"
+    steps:
+      - run:
+          name: Create a release on GitHub
+          command: |
+            set -xe
```
andriisoldatenko: same here.

.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
...
+            WORK_DIR=$(pwd)
+            cd /tmp
+            wget https://github.com/tcnksm/ghr/releases/download/v<< parameters.ghr_version >>/ghr_v<< parameters.ghr_version >>_linux_amd64.tar.gz
+            tar -xvf ./ghr_v<< parameters.ghr_version >>_linux_amd64.tar.gz
+            mkdir -p /tmp/bin
+            mv ghr_v<< parameters.ghr_version >>_linux_amd64/ghr /tmp/bin/
+            export PATH=/tmp/bin:$PATH
+            cd $WORK_DIR
+
+            # Generate release notes in Markdown format
```
andriisoldatenko: all this code done by goreleaser isn't ?

andriisoldatenko:
@sjmiller609 FYI we are using my fork of goreleaser https://github.com/andriisoldatenko/goreleaser. We had few issues in upstream, but did not have a chance to merge it to official repo.

ashb:
We don't currently run tests in circleci for this repo do we? I was thinking it would make sense to only allow releases that pass tests.

.circleci/config.yml
```diff
@@ -0,0 +1,72 @@
...
+              only: /v[0-9]+(\.[0-9]+)*(-.*)*/
+executors:
+  docker-executor:
+    environment:
+      GIT_ORG: astronomer
+      DOCKER_REPO: astronomerinc
+    docker:
+      - image: circleci/python:3
+commands:
+  commit-next-tag:
```
ashb: I don't see this used anywhere?

ashb: I'm blind. :D

sjmiller609:
I'm testing this on my fork... I'll update here with some screen shots when it's working.

@andriisoldatenko I am considering to not use our fork of goreleaser because we don't need to care about tags in the form of "X.Y.Z-blahblah" any more - we will only be publishing new patch version instead (because platform patch version not synced to CLI patch version)

sjmiller609:
> We don't currently run tests in circleci for this repo do we? I was thinking it would make sense to only allow releases that pass tests.

Hey @ashb , I totally agree. For the sake of incremental change, we were thinking to migrate tests / all non-circle CI over to Circle after the build and release process is working (and we make sure we like the whole process across all of the repos in the organization). That will look like setting up the build and release for all components (which is complete after this PR), then the helm chart (todo), then coming back and migrating everything off of any CI to Circle. In the interim, tests passing is soft-enforced by not merging PRs until all CI jobs are passing. This is the same as it is on all other components at the moment (for example orbit-ui's build is separate from the tests, and the tests are in drone still).

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+            pip install --user PyGithub packaging
+            NEXT_TAG=$(python .circleci/next_version.py)
+            # Make this environment variable available to following steps
+            echo "export NEXT_TAG=v${NEXT_TAG}" >> $BASH_ENV
+      - run:
+          name: Tag the Git commit
+          command: |
+            set -xe
+            git remote rm origin
+            git remote add origin https://sjmiller609:${GITHUB_TOKEN}@github.com/${GIT_ORG}/${CIRCLE_PROJECT_REPONAME}.git
```
sjmiller609: we need to ensure that we use my GITHUB_TOKEN for this to work. The correct token ends with "0cde"

sjmiller609:
It's tested working on my fork. Here are some screen shots:

The first workflow waits for manual approval, then it pushes a new tag. The tag is chosen by parsing the major / minor version from the branch name, then choosing the patch version one greater than the most recent github release with the matching major / minor.
image
excerpt from tagging logic:
```
We are on a release branch: release-1.0, detected major.minor version 1.0
We will find the most recent patch version of 1.0 and return it incremented by one
The most recent tag matching this release is v1.0.0
Parsed base version as 1.0.0
Calculated new version as 1.0.1
```

The second workflow is triggered by the push of the tag (from the previous workflow) and makes use of the goreleaser tool:
image

Here are the releases published on my fork: https://github.com/sjmiller609/astro-cli/releases

sjmiller609:
@andriisoldatenko re-requesting review

andriisoldatenko:
@ashb can you please take a look?

.circleci/config.yml
schnie: @sjmiller609 what should we do about hardcoding your name here?

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+          type: approval
+          filters:
+            branches:
+              only: '/release-.*/'
+      - new-tag:
+          requires:
+            - approve-release
+          filters:
+            branches:
+              only: '/release-.*/'
```
ashb: Are we not running any tests on PRs/master branch etc?

schnie: Tests are still in drone I believe. We'll be migrating all the testing phases over next. Initial step was just to move the release process into Circle, which mostly deals with the `release-x` branches.

andriisoldatenko:
@ashb @schnie @sjmiller609 
how close we are to merge this? how can i release new version of astro-cli with fixes for 0.12.x?

sjmiller609:
I'm not aware of any action required by me here.

sjmiller609:
> @ashb @schnie @sjmiller609
> how close we are to merge this? how can i release new version of astro-cli with fixes for 0.12.x?

iirc there is a reason why that is no longer applicable.

schnie:
We just haven't released the platform yet, so I think we should hold off until we're ready there. Can we release an alpha?

PR #325, opened 2020-04-09
Add missing check on error return from docker log in step

ashb:


ashb:
This error doesn't seem to appear though. But with this change the docker step fails with:

> Failed to authenticate to the registry, this can occur when registry is offline. Until authenticated you will not be able to push new images to your Airflow clusters

PR #330, opened 2020-05-08
allow annotations to be applied to service account

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/851


andriisoldatenko:
@ashb  @schnie PTAL

PR #332, opened 2020-05-12
Minimal change for docker auth to work

verdverm:
This is the minimal changes I needed to get my local setup working


vishwas-astro:
@verdverm this may no longer be needed in light of @andriisoldatenko's correct fix for it in a different PR that has gone into 0.13.1 version of the CLI

PR #339, opened 2020-06-09
Enforce Platform<->CLI 1:1 version mapping and display CLI upgrade/downgrade notice.

adam2k:
This is a new feature to give CLI feedback when the platform version differs from the CLI version.

Resolves https://github.com/astronomer/issues/issues/923

Local testing:

To test this PR locally, first open up your Makefile and change:

`LDFLAGS_VERSION=-X github.com/astronomer/astro-cli/version.CurrVersion=${VERSION}`

To something like: 

`LDFLAGS_VERSION=-X github.com/astronomer/astro-cli/version.CurrVersion=0.12.0`

So that after you build the cli can pick up this mock version.  While you're testing you can alter that number to get different results from the cli.

Run `make build` for your local astro-cli

Start houston-api and astro-ui so that you can grab an auth token.  Using your dev astro-cli ` astro auth login localhost` and use the astro-ui token.

After you are authenticated you can run any of the astro commands to receive output like:

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version
Astro Server Version: 0.15.0
Astro CLI Version: 0.15.0
Git Commit: 55f4fb5d25dfa5c1f9bf1800129869cd8eb769e1
```

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version
Astro Server Version: 0.15.5
Astro CLI Version: 0.15.8
Git Commit: 55f4fb5d25dfa5c1f9bf1800129869cd8eb769e1
```

If the versions match or if your CLI is a patch version ahead of the server CLI there isn't any feedback.  Here are the mismatch cases that give feedback:

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version
Error: There is an update for Astro CLI. You're using version 0.15.0, but 1.1.0 is the server version.
Please upgrade to the matching version before continuing. See https://www.astronomer.io/docs/cli-quickstart for more information.
To skip this check use the --skip-version-check flag.

Usage:
  astro version [flags]

Flags:
  -h, --help   help for version

Global Flags:
      --skip-version-check   skip version compatibility check
```

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version --skip-version-check
Astro Server Version: 1.1.0
Astro CLI Version: 0.14.0
Git Commit: 7683d7c9e6436f90aaa70a845a092e5638b762e0
```

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version
A new minor version of Astro CLI is available. Your version is 0.14.0 and 0.15.0 is the latest.
See https://www.astronomer.io/docs/cli-quickstart for more information.
Astro Server Version: 0.15.0
Astro CLI Version: 0.14.0
Git Commit: 7683d7c9e6436f90aaa70a845a092e5638b762e0
```

```
🚀  astro-cli git:(feature/add-cli-version-feedback) ✗ astroDev version
Your Astro CLI Version (0.16.0) is ahead of the server version (0.15.5).  Consider downgrading your Astro CLI to match.
Astro Server Version: 0.15.5
Astro CLI Version: 0.16.0
Git Commit: 55f4fb5d25dfa5c1f9bf1800129869cd8eb769e1
```


cmd/prerun.go
```diff
@@ -0,0 +1,94 @@
...
+
+// CheckDeploymentVersion for validation of the CLI vs Deployment Version
+func CheckDeploymentVersion(client *houston.Client, cmd *cobra.Command, out io.Writer) {
+	appCfg := deployment.GetAppConfig()
+	cv := version.CurrVersion
+
+	// Skip check if AppConfig is nil or is cv is empty
+	if appCfg != nil && cv != "" {
+		dv := appCfg.Version
+		validateVersions(cv, dv, out)
```
andriisoldatenko, line 25:
```suggestion
		validateVersions(version.CurrVersion, appCfg.Version, out)
```
no need this redundant lines, you are just waste of memory here by copying values.

adam2k:
> @adam2k could you give an example (here, in the PR) of the output please?

@ashb I added some instructions of how to test and what to expect in the PR description.  Let me know if you want me to run you through any of those steps and if you have more questions.  We are not blocking anything right now and only providing feedback.

schnie:
@adam2k with this update there are no hard blocks in place right? Looks like only warnings, which I think makes sense since we wouldn't want to break a customer CI/CD pipelines if we didn't actually need to.

Should we also show users the command they can run to upgrade?

adam2k:
> @adam2k with this update there are no hard blocks in place right? Looks like only warnings, which I think makes sense since we wouldn't want to break a customer CI/CD pipelines if we didn't actually need to.
> 
> Should we also show users the command they can run to upgrade?

@schnie I made a small update that will error when the major version of the cli is behind.  I can revert that change, but I wanted to get feedback from others to find out what's best.

I included a link to the documentation in the message `See https://www.astronomer.io/docs/cli-quickstart for more information.`.  

Do you think we should be more explicit in the instructions and output something like 'Please run brew install astronomer/tap/astro@0.XX.X OR curl -sSL https://install.astronomer.io | sudo bash -s -- v0.13.1 to upgrade`?  There are probably additional steps that we can take to figure out if they have brew installed or not and give them instructions for their particular set up, but that might be out of scope for this issue right now.

version/validate_compatibility.go
```diff
@@ -0,0 +1,99 @@
...
+	// Skip check if AppConfig is nil or is cv is empty
+	if serverCfg != nil && cliVer != "" {
+		return compareVersions(serverCfg.Version, cliVer, out)
+	}
+	return nil
+}
+
+func compareVersions(serverVer string, cliVer string, out io.Writer) error {
+	if isBehindMajor(serverVer, cliVer) {
+		return errors.Errorf(messages.ERROR_NEW_MAJOR_VERSION, cliVer, serverVer)
```
andriisoldatenko, line 31: @adam2k  pay attention here, i prefer to not crash just notify user.
@schnie @ashb thoughts?

adam2k: @andriisoldatenko thanks!  Yeah, I wanted to get feedback to see what everybody thinks is the best approach.  I can revert this change if most people don't like the error here.

andriisoldatenko:
**we have diff options:**
- add option `--skip-version-check` to ignore errors just print and user will accept risk
- configure option to `config.yaml`
- just always print
- always fail with error

version/validate_compatibility.go
sjmiller609: The UX for an error now is that there will be a graphQL error if old CLI that is incompatible with server or vice versa. This usually is resolved by a support ticket. Perhaps we replace graphQL errors with an upgrade error to be more informative.

In terms of enforce by default, my main concern is CI / automation users who are using an old minor version CLI that does still work with a new version. I prefer enforce by default when we have a compatibility matrix, and it's enforced with no --skip option for situations we know will break. This way we avoid breaking users where it happens to work between minor versions and we always stop behavior we know will break.

adam2k:
> **we have diff options:**
> 
>     * add option `--skip-version-check` to ignore errors just print and user will accept risk
> 
>     * configure option to `config.yaml`
> 
>     * just always print
> 
>     * always fail with error

@andriisoldatenko I've implemented the `--skip-version-check`.  We'll only be block on the true "major" version now instead of the "major" or "minor" version.  That means you'll only see the CLI blocking error once we get to 1.0.0 and beyond.  Everything else will be a warning.

vishwas-astro:
@adam2k @andriisoldatenko 

How about we go ahead, going forward, with our plan for generating a "x.y.z" release of the CLI for every "x.y.z" release of the platform.

If the version of CLI is less than the version of the platform, we strongly urge the users to upgrade to that version for everything to work seamlessly but it is only a warning and users continue at their own risk

If it is a major version difference, then make the warning an error and say that a major revision difference between platform and CLI is unacceptable.

Start imposing this logic only from 0.16 onwards, because we have not been generating CLI versions consistently for every platform release over the past few releases.

You can make all of this the default implementation and no need to do skip version check, I feel.

Does that work @schnie @paolaperaza @ashb ??

adam2k:
@vishwas-astro the way that this is implemented now we're only blocking for the true major version i.e. platform is on `1.0.0` and the cli is somewhere on `0.XX.X`.  All other cases (minor and patch mismatches) will return warnings about upgrading.  Do you think that is still too heavy handed?  We won't really be enforcing this until we get up to a `1.0.0` release.

The more I think about the 1:1 releases it seems like we could follow how `kubernetes` handles releases.  We probably need a bot to help automate some of this.

See: https://github.com/kubernetes/kubectl/releases
And: https://github.com/kubernetes/kubernetes/releases

andriisoldatenko:
@schnie PTAL

schnie:
Yea, a 1:1 relationship between the platform and the CLI gets annoying to manage, unless we automated it. A patch release on either the platform or the CLI will not typically require a change in both places, which makes it tedious.

Say we release 0.16 of the platform and CLI, then we discover an issue in the CLI and want to release v0.16.1. It seems weird to also release a v0.16.1 of the platform for no reason. The same goes for the other way around.

This is why I was suggesting a sync at every <major>.<minor> release, then let them evolve on patches independently. It gets a little confusing I guess, but feels cleaner than the previous scenario. Any thoughts?

adam2k:
@schnie in the short term I think there are two options:

1. We deal with the annoyance of manually managing the 1:1 relationship and then start work on automating that with a bot like k8s.
2. In this PR, we don't warn for different patch versions and we add that later once we have the 1:1 relationship automated (we'd probably end up revisiting this conversation again too).

In the long term it could simplify things going with option no. 1 because we won't have multiple change logs and there won't be the immediate situation of the platform updating patch versions independently of the CLI.  I think it would be more confusing in a situation where we have a `0.15.6` for the platform that is unrelated to a `0.15.6` version of the CLI.

The downside is that we need to have reminders to manually release CLI versions that match the platform.

schnie:
What about the situation where we have a CLI fix that has no changes on the platform?

vishwas-astro:
I am cool with what @schnie is suggesting. 

It probably has the following implications on the implementation:
1. If there is a major version mismatch between CLI and Platform, throw a blocking error on the CLI
2. If there is a minor version mismatch, throw a warning on the CLI and let users continue at their own risk. (This will require us to have a CLI release for every minor platform release)
3. If there is only a patch version mismatch, do nothing (Could that lead to a problem where we patched a CLI on a minor release that is necessary for it to function against that minor release of the platform?)

schnie:
I think that's the best route _for now_ @vishwas-astro. Purely so we can iterate on either the platform or the CLI via patch releases.

adam2k:
@vishwas-astro @schnie I can quickly make the updates to ignore patch for now.  We can always revisit and open up a new PR later to make more changes.

PR #340, opened 2020-06-18
Wip fix/update version tests

adam2k:


version/version.go
```diff
@@ -51,14 +50,14 @@ func CheckForUpdate(out io.Writer) error {
 	}
 
 	// fetch latest cli version
-	latestTagResp, err := api.RepoLatestRequest("astronomer", "astro-cli")
+	latestTagResp, err := client.RepoLatestRequest("astronomer", "astro-cli")
```
adam2k, line 53: I need to figure out how to mock `RepoLatestRequest` and `RepoTagRequest` because the responses should both be different to effectively test all of the cases.

adam2k: Currently, all of the client responses are the same.

PR #342, opened 2020-06-19
Increase code coverage to 50%

andriisoldatenko:
Increase code coverage up to 50% in total.

fix astronomer/issues#1329


airflow/airflow_test.go
```diff
@@ -0,0 +1,92 @@
...
+	assert.NoError(t, err)
+
+	exist, err := fileutil.Exists(filepath.Join(tmpDir, "requirements.txt"))
+
+	assert.NoError(t, err)
+	assert.True(t, exist)
+
+}
+
+
```
adam2k, line 65: This is a nit pick, but when I linted `gofmt` removed this line.

adam2k:
I didn't notice this was [WIP].  I'll wait to approve until you have all the changes in.

andriisoldatenko:
@schnie @adam2k PTAL

andriisoldatenko:
>Maybe we should consider moving those really long okResponse strings into separate test files similarly to test tables to make for shorter testing files?
in general it's a good idea but in practise we loose readability especially when we need to demonstrate the repose from houston. E.g. two similar tests have 1 line diff in okResponse from houston.

Makefile
```diff
@@ -29,7 +29,7 @@ cover:
 
 format:
 	@echo "--> Running go fmt"
-	@go fmt $(GOFILES)
+	@go fmt ./...
```
adam2k, line 32: 🚀👍

adam2k:
> > Maybe we should consider moving those really long okResponse strings into separate test files similarly to test tables to make for shorter testing files?
> 
> in general it's a good idea but in practise we loose readability especially when we need to demonstrate the repose from houston. E.g. two similar tests have 1 line diff in okResponse from houston.

This is OK for now.  We can come up with something to improve this later.

andriisoldatenko:
@schnie can i merge this?

PR #343, opened 2020-06-19
Use preferred, more inclusive language in code naming and error messaging

ryanahamilton:
Tests appear to pass, but I don't have enough experience with this application to be confident that I'm not overlooking any unintended consequences?


ryanahamilton:
Closing—didn't realize I was editing all vendor files!

PR #353, opened 2020-08-07
Remove Houston version check on upgrade

adam2k:
## Description

`astro upgrade` should not require Houston context.

## 🎟 Issue(s)

Resolves astronomer/issues#1606

## 🧪 Functional Testing

1. You can test locally by build a local version that is the same as the latest:

Without authentication to a server:
```
$ ./astro upgrade
Astro CLI Version: v0.17.0 (2020.07.24)
Astro CLI Latest: v0.17.0  (2020.07.24)
Astro Server Version: Please authenticate to a cluster to see server version
You are running the latest version.
```

With server authentication:
```
./astro upgrade
Astro CLI Version: v0.17.0 (2020.07.24)
Astro CLI Latest: v0.17.0  (2020.07.24)
Astro Server Version: 0.18.0
You are running the latest version.
```

2. You can test locally by building a local version that using an outdated `CurrVersion`:

Without server authentication:
```
$ rm -rf ~/.astro
$ go build -o astro -ldflags "-X github.com/astronomer/astro-cli/version.CurrVersion=0.15.0  -X github.com/astronomer/astro-cli/version.CurrCommit=5ae31076d9f2dfc874c533e6c729e1fa7681b7f4" main.go
$./astro upgrade
Astro CLI Version: v0.15.0 (2020.06.01)
Astro CLI Latest: v0.17.0  (2020.07.24)
Astro Server Version: Please authenticate to a cluster to see server version
A newer version of the Astronomer CLI is available.
To upgrade to latest, run:
	$ curl -sL https://install.astronomer.io | sudo bash
OR for homebrew users:
	$ brew install astronomer/tap/astro
```

With server authentication:
```
$ ./astro upgrade
Astro CLI Version: v0.15.0 (2020.06.01)
Astro CLI Latest: v0.17.0  (2020.07.24)
Astro Server Version: 0.18.0
A newer version of the Astronomer CLI is available.
To upgrade to latest, run:
	$ curl -sL https://install.astronomer.io | sudo bash
OR for homebrew users:
	$ brew install astronomer/tap/astro
```

## 📋 Checklist

- [x] Tested against [Houston-API](https://github.com/astronomer/houston-api/) 
- [ ] Updated any related [documentation](https://github.com/astronomer/astro-docs/)

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -2,15 +2,13 @@ package version
 
 import (
 	"fmt"
-
 	"io"
-
-	"github.com/pkg/errors"
 
 	"github.com/astronomer/astro-cli/deployment"
 	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/messages"
 	"github.com/astronomer/astro-cli/pkg/github"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -30,21 +28,7 @@ func PrintVersion(client *houston.Client, out io.Writer) error {
 	fmt.Fprintf(out, messages.CLI_CURR_VERSION+", ", version)
 	fmt.Fprintf(out, messages.CLI_CURR_COMMIT+"\n", gitCommit)
 
-	PrintServerVersion(client, out)
-
-	return nil
-}
-
-// PrintServerVersion outputs current server version
-func PrintServerVersion(client *houston.Client, out io.Writer) error {
-	appCfg, err := deployment.AppConfig(client)
-	if err != nil {
-		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", "Please authenticate to a cluster to see server version")
-	}
-
-	if appCfg != nil {
-		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", appCfg.Version)
-	}
+	printServerVersion(client, out)
 
 	return nil
 }
@@ -82,7 +66,7 @@ func CheckForUpdate(client *houston.Client, ghClient *github.Client, out io.Writ
 	fmt.Fprintf(out, messages.CLI_CURR_VERSION_DATE+"\n", currentTag, currentPub)
 	fmt.Fprintf(out, messages.CLI_LATEST_VERSION_DATE+"\n", latestTag, latestPub)
 
-	PrintServerVersion(client, out)
+	printServerVersion(client, out)
 
 	if latestTag > currentTag {
 		fmt.Fprintln(out, messages.CLI_UPGRADE_PROMPT)
@@ -99,4 +83,18 @@ func isValidVersion(version string) bool {
 		return false
 	}
 	return true
+}
+
+// printServerVersion outputs current server version
+func printServerVersion(client *houston.Client, out io.Writer) error {
+	appCfg, err := deployment.AppConfig(client)
+	if err != nil {
+		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", "Please authenticate to a cluster to see server version")
+	}
+
+	if appCfg != nil {
+		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", appCfg.Version)
+	}
+
+	return nil
 }
--- a/version/version.go
+++ b/version/version.go
@@ -30,6 +30,13 @@ func PrintVersion(client *houston.Client, out io.Writer) error {
 	fmt.Fprintf(out, messages.CLI_CURR_VERSION+", ", version)
 	fmt.Fprintf(out, messages.CLI_CURR_COMMIT+"\n", gitCommit)
 
+	PrintServerVersion(client, out)
+
+	return nil
+}
+
+// PrintServerVersion outputs current server version
+func PrintServerVersion(client *houston.Client, out io.Writer) error {
 	appCfg, err := deployment.AppConfig(client)
 	if err != nil {
 		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", "Please authenticate to a cluster to see server version")
@@ -43,7 +50,7 @@ func PrintVersion(client *houston.Client, out io.Writer) error {
 }
 
 // CheckForUpdate checks current version against latest on github
-func CheckForUpdate(client *github.Client, out io.Writer) error {
+func CheckForUpdate(client *houston.Client, ghClient *github.Client, out io.Writer) error {
 	version := CurrVersion
 
 	if !isValidVersion(version) {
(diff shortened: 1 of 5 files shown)


adam2k:
@paolaperaza can you give feedback on the upgrade message changes here?

andriisoldatenko:
>Coverage   43.40%   51.89%   +8.48%     
looks strange to me.

adam2k:
> > Coverage   43.40%   51.89%   +8.48%
> 
> looks strange to me.

I think it's a codecov error.

Screen Shot 2020-08-07 at 4 53 45 PM

paolaperaza:
@adam2k Thanks for doing this! A few general comments:

- UX looks good for users not authenticated to Astronomer
- Should we list `Server version` in the way `astro version` also does? Seems like that'd be more consistent, enforce the 1:1 platform <> CLI minor version compatibility for users to see/be reminded of.
- I wonder if we can push people to authenticate if they run this command but aren't... i.e.I'm thinking about a user that forgets to login, run astro uprade, updates to latest, then it turns out they're on an earlier platform version and get an error message about compatibility and have to downgrade. Including `Server Version` and a reminder to authenticate if that's not available might be enough?

In terms of copy, could simplify message to:

`To upgrade to latest, run:...` and then keep `OR for homebrew users`... What do you think?

adam2k:
> @adam2k Thanks for doing this! A few general comments:
> 
>     * UX looks good for users not authenticated to Astronomer
> 
>     * Should we list `Server version` in the way `astro version` also does? Seems like that'd be more consistent, enforce the 1:1 platform <> CLI minor version compatibility for users to see/be reminded of.
> 
>     * I wonder if we can push people to authenticate if they run this command but aren't... i.e.I'm thinking about a user that forgets to login, run astro uprade, updates to latest, then it turns out they're on an earlier platform version and get an error message about compatibility and have to downgrade. Including `Server Version` and a reminder to authenticate if that's not available might be enough?
> 
> 
> In terms of copy, could simplify message to:
> 
> `To upgrade to latest, run:...` and then keep `OR for homebrew users`... What do you think?

Thanks @paolaperaza!  Yes, we can add in the server version here too.  Do you think if we show the server version OR give the warning that they're not authenticated with a server, that would mitigate the last issue that you're describing enough?  We might need to think about that experience some more and create a follow up issue because it might increase the scope of this issue a lot.

In the short term, do you think having some warning message like `Upgrading beyond your server version may cause issues` would be enough?

adam2k:
@paolaperaza so this issue doesn't get lost.  We discussed making some of the additional changes outside of the message change inside of this issue: https://github.com/astronomer/issues/issues/1636

paolaperaza:
@adam2k Cool, commit looks good. Yea, I think if we list `Server Version` and whatever message is already there as-is that tells users to authenticate if they haven't already is enough to cover that problem/use case - definitely at least for now.

We can continue convo on the rest in astronomer/issues#1636!

version/version.go
```diff
@@ -2,15 +2,13 @@ package version
 
 import (
 	"fmt"
-
 	"io"
 
-	"github.com/pkg/errors"
```
andriisoldatenko, line 8: pls undo this change, bad formating:
must be 
STD lib
3rd party
custom pkgs

version/version.go
```diff
@@ -30,20 +28,13 @@ func PrintVersion(client *houston.Client, out io.Writer) error {
...
-		fmt.Fprintf(out, messages.HOUSTON_CURRENT_VERSION+"\n", appCfg.Version)
-	}
+	printServerVersion(client, out)
 
 	return nil
 }
 
 // CheckForUpdate checks current version against latest on github
-func CheckForUpdate(client *github.Client, out io.Writer) error {
+func CheckForUpdate(client *houston.Client, ghClient *github.Client, out io.Writer) error {
```
andriisoldatenko, line 37: JFYI for go code styling absolutely normal to rename `ghClient ` to `ghc`.

version/version.go
```diff
@@ -53,14 +44,14 @@ func CheckForUpdate(client *github.Client, out io.Writer) error {
 	}
 
 	// fetch latest cli version
-	latestTagResp, err := client.RepoLatestRequest("astronomer", "astro-cli")
+	latestTagResp, err := ghClient.RepoLatestRequest("astronomer", "astro-cli")
```
andriisoldatenko, line 47:
```suggestion
	latestTagResp, err := ghc.RepoLatestRequest("astronomer", "astro-cli")
```

version/version.go
```diff
@@ -53,14 +44,14 @@ func CheckForUpdate(client *github.Client, out io.Writer) error {
...
 	if err != nil {
 		fmt.Fprintln(out, err)
 		latestTagResp.TagName = messages.NA
 	}
 
 	// fetch meta data around current cli version
-	currentTagResp, err := client.RepoTagRequest("astronomer", "astro-cli", string("v")+version)
+	currentTagResp, err := ghClient.RepoTagRequest("astronomer", "astro-cli", string("v")+version)
```
andriisoldatenko, line 54:
```suggestion
	currentTagResp, err := ghc.RepoTagRequest("astronomer", "astro-cli", string("v")+version)
```

version/version.go
```diff
@@ -91,3 +84,17 @@ func isValidVersion(version string) bool {
 	}
 	return true
 }
+
+// printServerVersion outputs current server version
+func printServerVersion(client *houston.Client, out io.Writer) error {
```
andriisoldatenko, line 89: why do we need to copy this?
DRY? for https://github.com/astronomer/astro-cli/blob/e7a3f35b98584e3685e9a8f9764a23f4b5c17c71/version/version.go#L33-L40

adam2k: Yeah, I was trying to keep it DRY.  This method is now being invoked twice.  Once on line `33` and once on `69`.

version/version.go
adam2k, line 37:
```Suggestion
func CheckForUpdate(client *houston.Client, ghc *github.Client, out io.Writer) error {
```

version/version.go
adam2k: Oh, weird.  My editor auto organized those.  It looks like it's alphabetical from github.  I'll try to revert.

adam2k: I see what happened.  I removed the empty space between `"github.com/pkg/errors" ` and the astro imports and it grouped them altogether.

adam2k:
@andriisoldatenko do these updates look good to you now?

adam2k:
I validated this on dev.  The versions look off because it's expecting 0.19.0 to be the latest tag in GH (right now it's still 0.18.0).  As soon as we take 0.19.0 out of prerelease this will all look correct.

```
$ astro upgrade
A new minor version of Astro CLI is available. Your version is 0.18.0 and 0.19.0 is the latest.
See https://www.astronomer.io/docs/cli-quickstart for more information.
Astro CLI Version: v0.18.0 (2020.08.07)
Astro CLI Latest: v0.18.0  (2020.08.07)
You are running the latest version.
```

PR #354, opened 2020-08-07
Add pull request template

adam2k:
Add pull request template


adam2k:
@andriisoldatenko it's doing the same thing with the tests here too.

PR #355, opened 2020-08-18
Validate email

adam2k:
## Description

Client side test for email address validation.  Another PR for Houston API could also validate the email address using similar logic.

## 🎟 Issue(s)

Resolves astronomer/issues#1630

## 🧪 Functional Testing

Creates a user: `./astro workspace user add steve@apple.com`
Should fail: `./astro workspace user add steve@test.com`
Fails: `./astro workspace user add steve@notAnAddress`
Fails: `./astro workspace user add notAnEmail`

## 📸 Screenshots

N/A

## 📋 Checklist

- [ ] Updated any related [documentation](https://github.com/astronomer/astro-docs/)

### DIFF
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -1,0 +1,15 @@
+package user
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsValidEmail(t *testing.T) {
+	email := "steve@apple.com"
+	actual := IsValidEmail(email)
+	expected := true
+
+	assert.Equal(t, actual, expected)
+}
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -1,15 +1,50 @@
 package user
 
 import (
+	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestIsValidEmail(t *testing.T) {
+	// Test a valid email
 	email := "steve@apple.com"
 	actual := IsValidEmail(email)
 	expected := true
+
+	assert.Equal(t, actual, expected)
+
+	// Test a invalid email that is too long
+	const chars = "abcdefghijklmnopqrstuvwxyz0123456789@."
+	sb := make([]byte, 256)
+	for i := range sb {
+		sb[i] = chars[rand.Intn(len(chars))]
+	}
+	email = string(sb)
+	actual = IsValidEmail(email)
+	expected = false
+
+	assert.Equal(t, actual, expected)
+
+	// Test a invalid email that is too short
+	email = "abc"
+	actual = IsValidEmail(email)
+	expected = false
+
+	assert.Equal(t, actual, expected)
+
+	// Test a invalid email
+	email = "this@isnotanemail"
+	actual = IsValidEmail(email)
+	expected = false
+
+	assert.Equal(t, actual, expected)
+
+	// Test real address without MX records
+	email = "testing@test.com"
+	actual = IsValidEmail(email)
+	expected = false
 
 	assert.Equal(t, actual, expected)
 }
--- a/workspace/user.go
+++ b/workspace/user.go
@@ -4,15 +4,22 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/pkg/errors"
+
 	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/pkg/printutil"
+	"github.com/astronomer/astro-cli/user"
 )
 
 // Add a user to a workspace with specified role
-func Add(workspaceId, email, role string, client *houston.Client, out io.Writer) error {
+func Add(workspaceID string, email, role string, client *houston.Client, out io.Writer) error {
+	if !user.IsValidEmail(email) {
+		return errors.New(email + " is an invalid email address")
+	}
+
 	req := houston.Request{
 		Query:     houston.WorkspaceUserAddRequest,
-		Variables: map[string]interface{}{"workspaceId": workspaceId, "email": email, "role": role},
+		Variables: map[string]interface{}{"workspaceId": workspaceID, "email": email, "role": role},
 	}
 
 	r, err := req.DoWithClient(client)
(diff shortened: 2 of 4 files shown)


andriisoldatenko:
@adam2k wdyt if we add validations to houston-api so we can have similar rules in astro-cli/astro-ui?
cc @schnie

adam2k:
> @adam2k wdyt if we add validations to houston-api so we can have similar rules in astro-cli/astro-ui?
> cc @schnie

Yeah, I'm planning on doing another PR like this for Houston API too for the API only users.  I think it's good practice to check client side as well.

workspace/user.go
```diff
@@ -4,15 +4,22 @@ import (
...
-func Add(workspaceId, email, role string, client *houston.Client, out io.Writer) error {
+func Add(workspaceID string, email, role string, client *houston.Client, out io.Writer) error {
+	if !user.IsValidEmail(email) {
+		return errors.New(email + " is an invalid email address")
+	}
+
 	req := houston.Request{
 		Query:     houston.WorkspaceUserAddRequest,
-		Variables: map[string]interface{}{"workspaceId": workspaceId, "email": email, "role": role},
+		Variables: map[string]interface{}{"workspaceId": workspaceID, "email": email, "role": role},
```
andriisoldatenko, line 22: @adam2k can you pls explain this change to me? it's go style to use `workspaceId`.

workspace/user_test.go
```diff
@@ -24,14 +26,14 @@ func TestAdd(t *testing.T) {
 	api := houston.NewHoustonClient(client)
 	id := "ck1qg6whg001r08691y117hub"
 	role := "test-role"
-	email := "andrii@test.com"
+	email := "andrii@apple.com"
```
andriisoldatenko, line 29: @adam2k i'm not sure that we need this change.

workspace/user.go
adam2k: Are we using go lint?  The linter is suggesting to make this change.

Screen Shot 2020-08-19 at 10 31 38 AM

andriisoldatenko:
> > @adam2k wdyt if we add validations to houston-api so we can have similar rules in astro-cli/astro-ui?
> > cc @schnie
> 
> Yeah, I'm planning on doing another PR like this for Houston API too for the API only users. I think it's good practice to check client side as well.
problem that all clients should be in sync, so do we have such validations in astrio-ui or GraphQL server? So in this case we must duplicate this RegExp in 3 places. I don't think it's a good idea.

workspace/user_test.go
adam2k: That test fails unless I change that because @test.com doesn't have any MX records.

workspace/user.go
andriisoldatenko: IMO it's out of scope of this ticket. If you prefer to add `golint` it OK with me, but it should be seperate PR/issue also we need to update circleci pipeline to run it on each commit.

adam2k:
> > > @adam2k wdyt if we add validations to houston-api so we can have similar rules in astro-cli/astro-ui?
> > > cc @schnie
> > 
> > 
> > Yeah, I'm planning on doing another PR like this for Houston API too for the API only users. I think it's good practice to check client side as well.
> 
> problem that all clients should be in sync, so do we have such validations in astrio-ui or GraphQL server? So in this case we must duplicate this RegExp in 3 places. I don't think it's a good idea.

We have this client side check in Astro UI already.  That's why this issue is only reported for Astro CLI.  I don't think it's too hard to keep the regex the same in all three places.  This isn't something that will change frequently after it's implemented.

workspace/user.go
andriisoldatenko: so add to Makefile and to https://github.com/astronomer/astro-cli/blob/90c23704094c4bc047523c3db13a8f2868063565/.circleci/config.yml#L91 and to Dockerfile/ or dev deps.

andriisoldatenko:
> We have this client side check in Astro UI already. That's why this issue is only reported for Astro CLI. I don't think it's too hard to keep the regex the same in all three places. This isn't something that will change frequently after it's implemented.

what is the source of RegExp? how can we validate it?

andriisoldatenko:
>Should fail: ./astro workspace user add steve@test.com
-1 on this test case.
what about EE clients? i'm not sure it's a good idea to validate domain itself.

cc @paolaperaza

workspace/user.go
adam2k: oh okay, I thought this is what we were using.  I can revert and ignore in my editor.  I'll make that a separate issue.

andriisoldatenko, line 22: also this change breaks graphql query please revert it back:
https://github.com/astronomer/astro-cli/blob/99d2ef08ecc3c512129fc073d92ac233f69edc6d/houston/queries.go#L352

workspace/user.go
```diff
@@ -35,10 +42,10 @@ func Add(workspaceId, email, role string, client *houston.Client, out io.Writer)
 }
 
 // Remove a user from a workspace
-func Remove(workspaceId, userId string, client *houston.Client, out io.Writer) error {
+func Remove(workspaceID, userID string, client *houston.Client, out io.Writer) error {
 	req := houston.Request{
 		Query:     houston.WorkspaceUserRemoveRequest,
-		Variables: map[string]interface{}{"workspaceId": workspaceId, "userId": userId},
+		Variables: map[string]interface{}{"workspaceId": workspaceID, "userId": userID},
```
andriisoldatenko, line 48: same as above. https://github.com/astronomer/astro-cli/blob/99d2ef08ecc3c512129fc073d92ac233f69edc6d/houston/queries.go#L352

workspace/user_test.go
andriisoldatenko: and it's the idea of test. i disagree that we want to test MX record.
cc @sjmiller609

user/user_test.go
```diff
@@ -0,0 +1,50 @@
+package user
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsValidEmail(t *testing.T) {
```
andriisoldatenko, line 10: 1 test 1 assert or at least 1 expected results. 
you can refactor using table driven tests if you like or split into more tests.
https://github.com/golang/go/wiki/TableDrivenTests

andriisoldatenko (changes requested):

adam2k:
No longer required.  We are doing the validation in Houston API for now.

PR #363, opened 2020-10-13
Make Example Plugin 2.0 compatible

kaxil:
I have tested this on Staging both with 1.10.7 and with 2.0.0

closes https://github.com/astronomer/issues/issues/1922

### DIFF
--- a/airflow/include/exampleplugin.go
+++ b/airflow/include/exampleplugin.go
@@ -5,42 +5,65 @@ import "strings"
 // ExamplePlugin created with astro airflow init
 var ExamplePlugin = strings.TrimSpace(`
 from airflow.plugins_manager import AirflowPlugin
-from flask_admin.base import MenuLink
+from airflow.version import version
 
 """
 Look for the Astronomer tab in the UI.
 """
-airflow_plugins_ml = MenuLink(
-    category='Astronomer',
-    name='Airflow-Plugins',
-    url='https://github.com/airflow-plugins/')
+airflow_plugins_ml = {
+    "name": "Airflow-Plugins",
+    "category": "Astronomer",
+    "category_icon": "fa-rocket",
+    "href": "https://github.com/airflow-plugins/"
+}
 
-astro_docs_ml = MenuLink(
-    category='Astronomer',
-    name='Astronomer Docs',
-    url='https://www.astronomer.io/docs/')
+astro_docs_ml = {
+    "name": "Astronomer Docs",
+    "category": "Astronomer",
+    "category_icon": "fa-rocket",
+    "href": "https://www.astronomer.io/docs/"
+}
 
-astro_guides_ml = MenuLink(
-    category='Astronomer',
-    name='Airflow Guides',
-    url='https://www.astronomer.io/guides/')
+astro_guides_ml = {
+    "name": "Airflow Guide",
+    "category": "Astronomer",
+    "category_icon": "fa-rocket",
+    "href": "https://www.astronomer.io/guides/"
+}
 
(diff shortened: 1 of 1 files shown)


airflow/include/exampleplugin.go
```diff
@@ -5,42 +5,65 @@ import "strings"
 // ExamplePlugin created with astro airflow init
 var ExamplePlugin = strings.TrimSpace(`
 from airflow.plugins_manager import AirflowPlugin
-from flask_admin.base import MenuLink
+from airflow.version import version
```
andriisoldatenko, line 8: hm... if user still using airflow prior 2.0 and new astro cli?
may be
```suggestion
try:
    from flask_admin.base import MenuLink
except ImportError:
    from airflow.version import version
```
cc @ashb

andriisoldatenko (approved):

airflow/include/exampleplugin.go
```diff
@@ -5,42 +5,65 @@ import "strings"
...
+    "name": "Airflow Guide",
+    "category": "Astronomer",
+    "category_icon": "fa-rocket",
+    "href": "https://www.astronomer.io/guides/"
+}
+
+_appbuilder_menu_items = [airflow_plugins_ml, astro_docs_ml, astro_guides_ml]
+
+# Airflow >= 2.0
+if version.startswith("2"):
```
ashb, line 37: I generally prefer duck-typing to explicit version checks. I wonder if there is something we could do here...

airflow/include/exampleplugin.go
```diff
@@ -5,42 +5,65 @@ import "strings"
...
+    class AstroLinksPlugin(AirflowPlugin):
+        name = 'astronomer_menu_links'
+        operators = []
+        flask_blueprints = []
+        hooks = []
+        executors = []
+        macros = []
+        admin_views = []
+        appbuilder_views = []
+        appbuilder_menu_items = [airflow_plugins_ml, astro_docs_ml, astro_guides_ml]
```
ashb, line 47:
```suggestion
        appbuilder_menu_items = _appbuilder_menu_items
```
I think?

airflow/include/exampleplugin.go
kaxil: Unfortunately, airflow still consumes menu_link and does nothing with it, I will raise a PR to remove that but until than the only other duck-typing that is possible is to try importing `flask-admin`:

```
try:
     from flask_admin.base import MenuLink
     # Airflow <2
except ImportError:
     # Airflow >= 2
```

but that doesn't feel ideal

ashb: Is there a problem setting `menu_link` if we also set `appbuilder_menu_items`?

kaxil: `menu_link` is passed instance of `MenuLink` from `flask_admin.base` which does not exist for 2.0

kaxil: But to answer your question, there is no problem in setting `menu_link` if we also set `appbuilder_menu_items`?

ashb: TBH we don't even need to support non-RBAC - we've used RBAC only for a long time, and this is nothing to do with Airflow 2 or not, right?

So we can just only set the FAB-related fields, right?

kaxil: Updated

ashb (approved):

andriisoldatenko:
@kaxil i can merge if code practically works in airflow 1.x and 2.x alpha.

kaxil:
> @kaxil i can merge if code practically works in airflow 1.x and 2.x alpha.

yes it will

PR #364, opened 2020-10-13
Fix webserver command in docker-compose

kaxil:
When I built the CLI locally and tested it out, it failed with ` && airflow webserver` command not found error. 

After applying the patch in the PR it worked fine

### DIFF
--- a/airflow/docker_test.go
+++ b/airflow/docker_test.go
@@ -112,7 +112,7 @@ services:
   webserver:
     image: test-project-name/airflow:latest
     command: >
-      bash -c 'airflow create_user "$$@" && airflow webserver || airflow users create "$$@" && airflow webserver' -- -r Admin -u admin -e admin@example.com -f admin -l user -p admin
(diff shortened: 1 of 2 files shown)


airflow/docker_test.go
```diff
@@ -112,7 +112,7 @@ services:
   webserver:
     image: test-project-name/airflow:latest
     command: >
-      bash -c 'airflow create_user "$$@" || airflow users create "$$@"' -- -r Admin -u admin -e admin@example.com -f admin -l user -p admin && airflow webserver
+      bash -c 'airflow create_user "$$@" && airflow webserver || airflow users create "$$@" && airflow webserver' -- -r Admin -u admin -e admin@example.com -f admin -l user -p admin
```
ashb, line 115:
```suggestion
      bash -c '{ airflow create_user "$$@" || airflow users create "$$@"; } && airflow webserver' -- -r Admin -u admin -e admin@example.com -f admin -l user -p admin
```

I think will work.

kaxil: that worked, thanks

ashb (approved):

PR #380, opened 2020-11-05
Add pre commit

danielhoherd:
## Description

Add pre-commit to this project. I chose hooks that I thought would be relevant, but if we want to lighten it up or change things out that's fine. This is just a starting point, and people who are use this repo regularly are invited to deny or change any parts of what I've included.


adam2k (approved):
Thanks!  This LGTM 👍.  

Do you think we should fork your settings repo into the `Astronomer` org to use in the `.pre-commit-config.yml`?

danielhoherd:
> Thanks! This LGTM 👍.
> 
> Do you think we should fork your settings repo into the `Astronomer` org to use in the `.pre-commit-config.yml`?

I had definitely thought about that. I'm all for it. Let me do that real quick.

PR #381, opened 2020-11-10
Run sync_perm / sync-perm command in Webserver

kaxil:
After https://github.com/astronomer/ap-airflow/pull/146 and since we need to support CLI commands for <2.0 and >=2.0 the entrypoint does not sync-perm (https://github.com/astronomer/ap-airflow/pull/149) since the command run does not match to the condition. Hence lets run this command separately

### DIFF
--- a/airflow/docker_test.go
+++ b/airflow/docker_test.go
@@ -112,7 +112,7 @@ services:
   webserver:
     image: test-project-name/airflow:latest
     command: >
-      bash -c '{ airflow create_user "$$@" || airflow users create "$$@"; } && airflow webserver' -- -r Admin -u admin -e admin@example.com -f admin -l user -p admin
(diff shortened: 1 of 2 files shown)


ashb (approved):

andriisoldatenko:
@kaxil @ashb can you demonstrate practical tests in atleast 2 versions...before merging this.

kaxil:
Sure,

Use the following Dockerfile and run `astro dev start` with 0.21.0:

```dockerfile
FROM quay.io/astronomer/ap-airflow-dev:2.0.0-buster-onbuild-21752
```

and try to sign in in your Webserver, it will fail with Too Many Redirects error.

Build the cli that includes commits in this PR, try the above steps again and it will work, I have tested it locally.

PR #382, opened 2020-11-11
Set AIRFLOW__WEBSERVER__UPDATE_FAB_PERMS=True

schnie:


kaxil (approved):

kaxil (commented):
We don't strictly need this as https://github.com/astronomer/astro-cli/pull/381 solves it already

kaxil (changes requested):
We don't strictly need this as #381 solves it already

schnie:
Oh great.

PR #391, opened 2020-11-28
Improve command help text

ryw:
Will continue to improve this text, but wanted to at least normalize capitalization and consolidate verbiage. We should review with @astrorfox to develop an official glossary of platform terms.

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -43,9 +43,9 @@ func newAirflowRootCmd(client *houston.Client, out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:        "airflow",
 		Aliases:    []string{"a"},
-		Short:      "Manage airflow projects",
-		Long:       "Airflow projects are a single top-level directory which represents a single production Airflow deployment",
-		Deprecated: "could please use new command instead `astro dev [subcommands] [flags]`",
+		Short:      "Manage Airflow projects",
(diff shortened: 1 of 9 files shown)


schnie (approved):
Nice! Agree on defining a glossary of terms to help define the product ASAP.

astrorfox:
For the glossary, are you referring to consistent definitions around clusters, deployments, etc?

schnie:
Exactly. I don't think we've quite nailed down everything, and the things we do have nailed down now, _could_ change in AEO/AstroHub world, if we wanted to.

Once we have the glossary defined, we can spread it around so we all speak the same language.

Another thing is also to simply define the relationship between a few things. For instance _right now_ there is an implicit relationship between an astro "project" and astro "deployments". That relationship is one to many. One project (DAG source code and scaffolding) can be pushed to many deployments in the remote "[kubernetes] cluster".

I don't know if we've ever documented this structure officially. Some of those relationships may change, and we may want to revisit the words we use here. @astrorfox interested to hear if you have any thoughts. Would be nice to have some of the core glossary nailed down as we start building the data model.

ryw:
Yes a lot of this nomenclature is still informal - let's formalize in new product.

astrorfox:
@jwitz is going to start a glossary in our style guide. TBD on whether we'll be able to harvest cleanly proper nouns from our current docs to prime it, but otherwise we'll crowdsource.

PR #395, opened 2020-12-07
Add airflow 2.0 min requirements

adam2k:
## Description

Make sure that an Airflow deployment is using version `1.10.14` before allowing the upgrade to `2.0.x`

## 🎟 Issue(s)

Resolves astronomer/issues#2289

## 🧪 Functional Testing

It's possible to test this change locally.

1. Make build with a valid version number in the Makefile
2. Start up Houston API and Astro UI
3. Create a new deployment with a version prior to `1.10.14`.  (I was testing with `1.10.12`) 
4. Make sure you have Airflow 2.0 added into your database in the `AirflowRelease` table.  If not, manually add it (you can copy and existing deployment and change the version number for testing purposes)
5. Attempt to upgrade the airflow deployment to 2.0 (see the screen shots below).
6.  Make sure you see an error message.
7. After getting the error message, update your deployment Airflow version to `1.10.14` and attempt to upgrade again.  You should now see the success message.

## 📸 Screenshots

Screen Shot 2020-12-07 at 12 16 01 PM
Screen Shot 2020-12-07 at 12 16 23 PM

## 📋 Checklist

### DIFF
--- a/deployment/deployment.go
+++ b/deployment/deployment.go
@@ -1,6 +1,7 @@
 package deployment
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"sort"
@@ -11,6 +12,7 @@ import (
 	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/pkg/input"
 	"github.com/astronomer/astro-cli/pkg/printutil"
+	"github.com/astronomer/astro-cli/settings"
 	"github.com/fatih/camelcase"
 )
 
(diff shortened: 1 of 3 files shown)


settings/settings.go
```diff
@@ -32,7 +32,7 @@ var (
 	settings Config
 
 	// Version 2.0.0
-	newAirflowVersion uint64 = 2
+	NewAirflowVersion uint64 = 2
```
andriisoldatenko, line 35: New - is slightly confuse me, cause it's de facto constructor of objects in golang.
can we rename to something more readable rather than just `NewAirflowVersion`?
`AirflowVersionSecond`?

deployment/deployment.go
```diff
@@ -331,3 +339,36 @@ func getDeployment(deploymentId string, client *houston.Client) (*houston.Deploy
...
+	upgradeVersion := strconv.FormatUint(settings.NewAirflowVersion, 10)
+	minRequiredVersion := "1.10.14"
+	airflowUpgradeVersion, err := semver.NewVersion(upgradeVersion)
+	if err != nil {
+		return err
+	}
+
+	desiredVersion, err := semver.NewVersion(desiredAirflowVersion)
+	if err != nil {
+		return err
```
adam2k, line 353: @andriisoldatenko I can add more tests for these cases if you think we need them.  I can put it invalid version numbers.

settings/settings.go
adam2k: Yeah, I didn't know what we should call it because it's a special case.  `AirflowVersionSecond` or `AirflowVersionTwo`?

deployment/deployment.go
andriisoldatenko: it's okey.

settings/settings.go
andriisoldatenko: yep but i. prefer to name explain value, since it's airflow 2.0.0 specific.
`AirflowVersionTwo` +1

adam2k: Updated to `AirflowVersionTwo`

andriisoldatenko (approved):

PR #401, opened 2021-03-03
Replaces the default astronomer image with 2.0.0-buster-onbuild

cmarteepants:
The image with the tag "latest_onbuild" references image b17129847bec, which is an alpine image for Airflow 1.10.7. 

This image is broken. An error is received when the user attempts to rebuild (aka deploy their dags to Astronomer Platform). There is an issue to implement a proper fix (https://github.com/astronomer/issues/issues/2275), but until that can be done, this PR changes the default to a 2.0 on-build debian image. This ensures that users who are using Astro cli for the first time to create a new deployment project will start with a non-broken image, and hopefully have a smoother experience.

Resolves astronomer/issues#2660

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
         environment:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
-      - image: quay.io/astronomer/ap-houston-api:latest
+      - image: quay.io/astronomer/ap-houston-api:0.23.15
         name: houston
         entrypoint: "yarn start"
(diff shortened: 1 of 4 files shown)


andriisoldatenko (approved):

danielhoherd (approved):

adam2k (approved):
Thanks @cmarteepants! 🥳

vishwas-astro:
@adam2k could you pre-release an 0.23.4 with this and we can have someone in @ernest-kr's do a quick test against it?

PR #418, opened 2021-06-30
Add an insecure TLS option

andriisoldatenko:
resolves https://github.com/astronomer/issues/issues/2976


pkg/github/github.go
```diff
@@ -12,10 +12,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-var (
-	httpclient = httputil.NewHTTPClient()
```
andriisoldatenko, line 16: unused code.

bote795:
can a test be written for this?

andriisoldatenko:
> can a test be written for this?

i'm not sure, i'll try.

andriisoldatenko:
@bote795 i think it's not possible to write test, since  it's just transport option. Please explain which specific test you want?

bote795:
> @bote795 i think it's not possible to write test, since it's just transport option. Please explain which specific test you want?

Wasn't sure how to either, just wanted to make sure

bote795 (approved):
lgtm

PR #452, opened 2021-11-09
fix: cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.ReadPassword

andriisoldatenko:
fix https://app.circleci.com/pipelines/github/astronomer/astro-cli/798/workflows/06126152-9324-40cb-9174-f4a84bc78e32/jobs/1162

### DIFF
--- a/pkg/input/input.go
+++ b/pkg/input/input.go
@@ -23,7 +23,7 @@ func Text(promptText string) string {
 // Password requests a users passord, does not print out what they entered, and returns it
 func Password(promptText string) (string, error) {
 	fmt.Print(promptText)
-	bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
+	bytePassword, err := terminal.ReadPassword(int(syscall.Stdin)) // nolint: unconvert
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


pkg/input/input.go
```diff
@@ -23,7 +23,7 @@ func Text(promptText string) string {
 // Password requests a users passord, does not print out what they entered, and returns it
 func Password(promptText string) (string, error) {
 	fmt.Print(promptText)
-	bytePassword, err := terminal.ReadPassword(syscall.Stdin)
+	bytePassword, err := terminal.ReadPassword(int(syscall.Stdin)) // nolint: unconvert
```
bote795, line 26: Do we need to add a test

andriisoldatenko: i just reverted what was before. Linter suggested to remove int(), but it breaks:
```
 astro-cli git:(main) CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build
# github.com/astronomer/astro-cli/pkg/input
pkg/input/input.go:26:44: cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.ReadPassword
```


```
 make lint
golangci/golangci-lint info checking GitHub for tag 'v1.37.1'
golangci/golangci-lint info found version: 1.37.1 for v1.37.1/darwin/amd64
golangci/golangci-lint info installed /Users/andrii/workspace/src/github.com/astronomer/astro-cli/testbin/golangci-lint
pkg/input/input.go:26:48: unnecessary conversion (unconvert)
        bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
                                                      ^
make: *** [lint] Error 1
```

bote795 (approved):

PR #461, opened 2021-12-13
added changes required to support Podman on RHEL

neel-astro:
## Description

Changes:
- disabled labels from volume mounting by adding `type: spc_t` for RHEL, as adding `:z` or `:Z` options to mountPath is not supported in non-SELinux based OS (Ubuntu, macOS)
- set built image format to `Dockerv2ImageManifest` which wasn't the default value at least in RHEL and is needed as per Houston logic in `astro deploy` logic
- had to move from `kube --down` to `pod rm` in order to kill local airflow dev setup, as kube down options is only present in 3.4.x and latest Podman version in RHEL is still 3.3.x
- updated podman mocks to include `pod rm` function

## 🎟 Issue(s)

Related astronomer/issues#3425

## 🧪 Functional Testing

Have tested these changes in RHEL with Podman versions 3.1.x, 3.2.x, 3.3.x, and also in Ubuntu & MacOS with additional Podman version 3.4.x.

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/include/podconfigyml.go
+++ b/airflow/include/podconfigyml.go
@@ -75,6 +75,8 @@ spec:
       readOnlyRootFilesystem: false
       runAsGroup: 50000
       runAsUser: 50000
+      seLinuxOptions:
+        type: spc_t
     volumeMounts:
     - mountPath: /usr/local/airflow/dags
       name: airflow-dags-dir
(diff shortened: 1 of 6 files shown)


andriisoldatenko (approved):

andriisoldatenko:
@neel-astro how can i practically test it? do i need to run VM with redhat or i can run it in docker?

neel-astro:
> how can i practically test it? do i need to run VM with redhat or i can run it in docker?

@andriisoldatenko theoretically it can be tried in a docker container: https://www.redhat.com/sysadmin/podman-inside-container, but would recommend to run in an VM with redhat.

bote795 (approved):

PR #481, opened 2022-01-11
added changes to allow darwin/arm64 images

neel-astro:
## Description
Changes:
- Regenerated godownloader.sh with latest version of godownloader
- Added darwin/arm64 in allowed platforms only when release version is greater than or equal to 0.27.2 

Result on Mac M1:
Screenshot 2022-01-12 at 1 22 31 PM

Result on Ubuntu:
Screenshot 2022-01-12 at 1 26 34 PM

TODO:
- [x] verify the changes on other OS
- [x] might want to fix the log line in case to better reflect the case when version is found (Added meaningful response for cases such as trying to install CLI version 0.27.1 on darwin/arm64 platform will return `platform darwin/arm64 is not supported for 0.27.1`)

## 🎟 Issue(s)

Related astronomer/issues#3323

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


bote795:
did u add code to script or was this just pulled from latest sorry a bit confused

neel-astro:
@bote795 this is the part of code which I had to add manually: https://github.com/astronomer/astro-cli/pull/481/files#diff-89a2b81a9fb596df6e472e79564bbc8eae0c5f41ee84aa03bf4a616610d043b0R71-R78

rest of the changes are because of using latest version of [godownloader](https://github.com/goreleaser/godownloader) to regenerate this bash file from goreleaser.yml

bote795:
> @bote795 this is the part of code which I had to add manually: https://github.com/astronomer/astro-cli/pull/481/files#diff-89a2b81a9fb596df6e472e79564bbc8eae0c5f41ee84aa03bf4a616610d043b0R71-R78
> 
> rest of the changes are because of using latest version of [godownloader](https://github.com/goreleaser/godownloader) to regenerate this bash file from goreleaser.yml

is there anything we can do so that we can give a warning or something when we try to regenerate it?

neel-astro:
> is there anything we can do so that we can give a warning or something when we try to regenerate it?

Sorry, I didn't exactly understood what you meant over here regarding giving warning

bote795:
> > is there anything we can do so that we can give a warning or something when we try to regenerate it?
> 
> Sorry, I didn't exactly understood what you meant over here regarding giving warning

like when someone tries to update this file again since its generated it will get replaced right, what can we do to inform whoever tries to do that that we added that block of code

neel-astro:
good point @bote795
I can't think of anything other than maybe comments around the code added manually, so that it can be caught either by the developer or reviewer (also, this manual code might not be needed in far future, once the user base using CLI <= 0.27.1 diminishes).
Another thing to note, the godownloader library itself is deprecated, so ideally there is no need to regenerate this file in future, unless migrating away from the library itself.

bote795 (approved):
lgtm

andriisoldatenko (approved):

PR #484, opened 2022-01-21
replace pkg/errors with stdlib errors

neel-astro:
## Description
Changes:
- Moved from pkg/errors to stdlib errors package as both are functionally almost the same with later being the only package which would receive new functional updates in future. Changes for this migration were:
  - Replace Wrap method with fmt.Errorf with %w formatting verb
  - As we moved away from pkg/errors, golinter was able to pick up more of err113 linting issues for moving the wrapped static errors from code logic to package level vars

## 🎟 Issue(s)

Related astronomer/issues#3965

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -12,7 +12,6 @@ import (
 	"github.com/astronomer/astro-cli/pkg/fileutil"
 
 	semver "github.com/Masterminds/semver/v3"
-	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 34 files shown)


bote795:
do they both return a string? sorry not super familiar with the libs just want to make sure we are not losing any behavior

neel-astro:
Both of them will return an interface of type `error` which is simply having one method `Error()` which would return the string of that error.

The only major functional difference between the two libs are that pkg/errors support error stack trace which isn't present in stdlib errors, but that's something which we anyways don't use it in CLI.

Found a decent blog post illustrating that: https://blog.dharnitski.com/2019/09/09/go-errors-are-not-pkg-errors/

bote795 (approved):

andriisoldatenko (approved):
well done!

PR #486, opened 2022-01-25
added health check in webserver docker container

neel-astro:
## Description
Changes:
- Added healthcheck on webserver container in docker setup, and waiting for webserver to come up before applying settings.yaml file, which use to fail before as postgres DB at times wasn't up when connections/pools/variables were being created from settings.yaml file.

## 🎟 Issue(s)

Related astronomer/issues#3370
Related astronomer/issues#2997

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


andriisoldatenko (approved):

airflow/docker.go
```diff
@@ -316,6 +323,26 @@ func checkServiceState(serviceState, expectedState string) bool {
 	return scrubbedState == expectedState
 }
 
+func (d *DockerCompose) webserverHealthCheck() error {
+	err := d.composeService.Events(context.Background(), d.projectName, api.EventsOptions{
+		Services: []string{config.CFG.WebserverContainerName.GetString()}, Consumer: func(event api.Event) error {
+			if event.Status == "health_status: healthy" {
```
bote795, line 329: can this status be a const string? followed by the exec_die as well?

neel-astro: make sense, will move it to a package level constant

airflow/docker.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var errComposeProjectRunning = errors.New("project is up and running")
```
AmFlint, line 30: Maybe it would be best to put all the error message in a `airflow/errors.go` file.

So it the package grows in the futur, we know where to look for error messages

airflow/docker_test.go
```diff
@@ -290,6 +296,25 @@ func TestDockerGetContainerIDFailure(t *testing.T) {
 	assert.Contains(t, id, "")
 }
 
+func TestDockerWebserverHealthCheck(t *testing.T) {
+	composeMock, docker, _ := getComposeMocks()
+	mockCall := composeMock.On("Events", mock.Anything, mock.Anything, mock.Anything)
```
AmFlint, line 301: Can't we use a more specific mock method than "Mock.Anything" ?

This is not a big deal, but we prefer to make mock assertions as specific as possible to avoid side effects in the tests.

airflow/docker.go
```diff
@@ -316,6 +323,26 @@ func checkServiceState(serviceState, expectedState string) bool {
...
 }
 
+func (d *DockerCompose) webserverHealthCheck() error {
+	err := d.composeService.Events(context.Background(), d.projectName, api.EventsOptions{
+		Services: []string{config.CFG.WebserverContainerName.GetString()}, Consumer: func(event api.Event) error {
+			if event.Status == "health_status: healthy" {
+				fmt.Println("\nProject is running! All components are now available.")
+				// have to return an error to break from the event listener loop
+				return errComposeProjectRunning
+			} else if event.Status == "exec_die" {
```
AmFlint, line 333: The `exec_die` event status is only returned when Airflow components are not up yet ?

If it's also returned during unknown errors, how do you know you should stop the health check and return an error from this function ?

airflow/docker_test.go
neel-astro: Given that we had to override the function logic to unit test our code and also since we were passing a function as an argument to the method, made it difficult to be more specific than `mock.Anything`. Hence I had skipped it for all the arguments initially, as of now have added specific arguments except the last one which takes in a function as an argument.

airflow/docker.go
neel-astro: moved to a separate file

airflow/docker.go
neel-astro: `exec_die` event corresponds to ending of any `exec` call, which in this case will be for each health check call invoked by docker (heath check is defined here: https://github.com/astronomer/astro-cli/pull/486/files#diff-3025f0d00360d8168d677ee4b64f6f24b11069b778029c2127a80b50a01d62b3R143-R148).

This function is only involved to check the result of the health check done internally by Docker. States which are usually reported by Docker would be starting, healthy or unhealthy. Docker on its own, doesn't take any action even in case container is unhealthy, it will simply report it and we in this function are just consuming that information to figure out when the container is healthy.

Just added a counter and a break condition to break out of that loop as a fail safe condition.

airflow/docker.go
neel-astro: moved to a package level constant

AmFlint (approved):

bote795 (approved):
🚢

PR #487, opened 2022-01-25
astro-cli should prompt the user for namespace name input when pre-deployment validation is configured

ajayy004:
## Description

> based on pre-deployment validation config, user should be prompted for k8 namespace name to be enter instead of select from available k8 namespaces

## 🎟 Issue(s)

Related astronomer/issues#4132

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/deployment/deployment.go
+++ b/deployment/deployment.go
@@ -33,6 +33,7 @@ var (
 	ErrKubernetesNamespaceNotAvailable = errors.New("no kubernetes namespaces are available")
 	ErrNumberOutOfRange                = errors.New("number is out of available range")
 	ErrMajorAirflowVersionUpgrade      = fmt.Errorf("Airflow 2.0 has breaking changes. To upgrade to Airflow 2.0, upgrade to %s first and make sure your DAGs and configs are 2.0 compatible", minAirflowVersion) //nolint:golint,stylecheck
+	ErrKubernetesNamespaceNotSpecified = errors.New("no kubernetes namespaces name specified")
 )
 
 type ErrParsingInt struct {
@@ -133,6 +134,14 @@ func CheckPreCreateNamespaceDeployment(client *houston.Client) bool {
 	return appConfig.Flags.ManualNamespaceNames
 }
 
+func CheckNamespaceFreeformEntryDeployment(client *houston.Client) bool {
+	appConfig, err := AppConfig(client)
+	if err != nil {
+		return false
+	}
+	return appConfig.Flags.NamespaceFreeformEntry
+}
+
 func CheckTriggererEnabled(client *houston.Client) bool {
 	logrus.Debug("Checking for triggerer flag")
 	appConfig, err := AppConfig(client)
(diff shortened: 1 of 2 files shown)


deployment/deployment.go
```diff
@@ -280,6 +297,15 @@ func getDeploymentSelectionNamespaces(client *houston.Client, out io.Writer) (st
 	return names[i-1].Name, nil
 }
 
+func getDeploymentNamespaceName() (string, error) {
+	namespaceName := input.Text("\nKubernetes Namespace Name: ")
+	noSpaceString := strings.ReplaceAll(namespaceName, " ", "")
+	if noSpaceString == "" {
```
ajayy004, line 303: Does it make sense to do empty string validation?

deployment/deployment.go
```diff
@@ -133,6 +134,14 @@ func CheckPreCreateNamespaceDeployment(client *houston.Client) bool {
 	return appConfig.Flags.ManualNamespaceNames
 }
 
+func CheckNamespaceFreeformEntryDeployment(client *houston.Client) bool {
```
neel-astro, line 137: Nit: `CheckNamespaceFreeformEntryDeployment` => `checkNamespaceFreeformEntryDeployment`

If the scope of this function is just limited to this package, then we can keep this function private (any field, method or function starting with uppercase letter will be a public function, and if it is starting with lowercase letter then it will be a private function).

deployment/deployment.go
neel-astro: It does make sense to me to validate the input, but it would also be decent idea to strip unwanted spaces from the input text entered by the user, i.e.
```
namespaceName = strings.ReplaceAll(namespace, " ", "")
if namespace == "" {
    return "", ErrKubernetesNamespaceNotSpecified
}
return namespaceName, nil
```

deployment/deployment.go
```diff
@@ -146,14 +155,22 @@ func CheckTriggererEnabled(client *houston.Client) bool {
 func Create(label, ws, releaseName, cloudRole, executor, airflowVersion, dagDeploymentType, nfsLocation, gitRepoURL, gitRevision, gitBranchName, gitDAGDir, sshKey, knownHosts string, gitSyncInterval, triggererReplicas int, client *houston.Client, out io.Writer) error {
 	vars := map[string]interface{}{"label": label, "workspaceId": ws, "executor": executor, "cloudRole": cloudRole}
 
-	if CheckPreCreateNamespaceDeployment(client) {
+	if CheckPreCreateNamespaceDeployment(client) && !CheckNamespaceFreeformEntryDeployment(client) {
```
neel-astro, line 158: nit: we can get rid of individual function call to get a field value from appConfig, but rather have a single call to Houston to get all the appConfig and then directly picking the individual field throughout the `Create` method. This will save multiple API calls to Houston to fetch the same appConfig data.

```
appConfig, err := AppConfig(client)
if err != nil {
                 fmt.Println("Error fetching feature flags from Houston: %s", err.Error())
}
if appConfig.Flags.ManualNamespaceNames && !appConfig.Flags.NamespaceFreeformEntry {
		namespace, err := getDeploymentSelectionNamespaces(client, out)
		if err != nil {
			return err
		}
		vars["namespace"] = namespace
}

if appConfig.Flags.ManualNamespaceNames && appConfig.Flags.NamespaceFreeformEntry {
		namespace, err := getDeploymentNamespaceName()
		if err != nil {
			return err
		}
		vars["namespace"] = namespace
}
if releaseName != "" && appConfig.ManualReleaseNames {
		vars["releaseName"] = releaseName
}
.......
if appConfig.Flags.TriggererEnabled {
		vars["triggererReplicas"] = triggererReplicas
}
```

neel-astro (approved):
Great work 👍 , have added few minor suggestions

deployment/deployment.go
ajayy004: Thanks, I think it will make more sense to make it private since I don't see a use case for now where we can call this from outside.

deployment/deployment.go
ajayy004: Yes I think it does make sense to do.

ajayy004:
@neel-astro Will it be fine If I open another PR with the suggested changes?

neel-astro:
> @neel-astro Will it be fine If I open another PR with the suggested changes?

yeah sure 👍, you can reference [this issue](https://github.com/astronomer/issues/issues/4147) if you are consolidating all Houston API calls in `Create` method.

neel-astro (approved):

andriisoldatenko (approved):

PR #489, opened 2022-01-28
added command to fetch triggerer logs

neel-astro:
## Description
Changes:
- Added command `astro deployment logs triggerer` to fetch Airflow triggerer logs, command will only be visible if the feature flag is set.
- Consolidate `webserverRemoteLogs`, `schedulerRemoteLogs` & `workerRemoteLogs` into a single `fetchRemoteLogs` method.

## 🎟 Issue(s)

Related astronomer/issues#3807

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/deployment.go
+++ b/cmd/deployment.go
@@ -112,7 +112,7 @@ func newDeploymentRootCmd(client *houston.Client, out io.Writer) *cobra.Command 
 		newDeploymentListCmd(client, out),
 		newDeploymentUpdateCmd(client, out),
 		newDeploymentDeleteCmd(client, out),
-		newLogsCmd(),
+		newLogsCmd(client),
 		newDeploymentSaRootCmd(client, out),
 		newDeploymentUserRootCmd(client, out),
 		newDeploymentAirflowRootCmd(client, out),
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -1,11 +1,21 @@
 package cmd
 
 import (
+	"fmt"
 	"time"
 
+	"github.com/astronomer/astro-cli/deployment"
+	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/logs"
 
 	"github.com/spf13/cobra"
+)
+
+const (
+	logWebserver = "webserver"
+	logScheduler = "scheduler"
+	logWorker    = "worker"
+	logTriggerer = "triggerer"
 )
 
 var (
@@ -27,7 +37,7 @@ var (
 `
 )
 
-func newLogsCmd() *cobra.Command {
+func newLogsCmd(client *houston.Client) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "logs",
 		Aliases: []string{"log", "l"},
@@ -40,10 +50,18 @@ func newLogsCmd() *cobra.Command {
 		newSchedulerLogsCmd(),
 		newWorkersLogsCmd(),
 	)
+
+	appConfig, err := deployment.AppConfig(client)
+	if err != nil {
+		fmt.Println("Error checking feature flag", err)
+	} else if appConfig.Flags.TriggererEnabled {
+		cmd.AddCommand(newTriggererLogsCmd())
+	}
+
 	return cmd
 }
 
-func newLogsDeprecatedCmd() *cobra.Command {
+func newLogsDeprecatedCmd(client *houston.Client) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:        "logs",
 		Aliases:    []string{"log", "l"},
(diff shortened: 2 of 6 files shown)


cmd/logs.go
```diff
@@ -71,7 +90,9 @@ func newWebserverLogsCmd() *cobra.Command {
 astro deployment logs webserver YOU_DEPLOYMENT_ID -s string-to-find
 `,
 		Args: cobra.ExactArgs(1),
-		RunE: webserverRemoteLogs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return fetchRemoteLogs("webserver", args)
```
AmFlint, line 94: I think we should use constants for Airflow components to fetch the logs for like `const LOG_WEBSERVER = "webserver"`

cmd/logs.go
```diff
@@ -91,7 +112,9 @@ func newSchedulerLogsCmd() *cobra.Command {
 astro deployment logs scheduler YOU_DEPLOYMENT_ID -s string-to-find
 `,
 		Args: cobra.ExactArgs(1),
-		RunE: schedulerRemoteLogs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return fetchRemoteLogs("scheduler", args)
```
AmFlint, line 116: See my previous comment, we should use constants for `scheduler`

cmd/logs.go
```diff
@@ -110,7 +134,9 @@ func newWorkersLogsCmd() *cobra.Command {
 astro deployment logs workers YOU_DEPLOYMENT_ID -s string-to-find
 `,
 		Args: cobra.ExactArgs(1),
-		RunE: workersRemoteLogs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return fetchRemoteLogs("worker", args)
```
AmFlint, line 138: Same as previous comments, use constant for `worker`

cmd/logs_test.go
```diff
@@ -1,16 +1,33 @@
...
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestDeploymentLogsRootCommand(t *testing.T) {
 	testUtil.InitTestConfig()
-	output, err := executeCommand("deployment", "logs")
+	okResponse := `{
+		"data": {
+		  "appConfig": {"triggererEnabled": true, "featureFlags": { "triggererEnabled": true}}
```
AmFlint, line 19: Maybe you could also add a test with the triggererEnabled = false to test that the command is skipped if the feature is disabled

cmd/deployment.go
```diff
@@ -112,7 +112,7 @@ func newDeploymentRootCmd(client *houston.Client, out io.Writer) *cobra.Command
 		newDeploymentListCmd(client, out),
 		newDeploymentUpdateCmd(client, out),
 		newDeploymentDeleteCmd(client, out),
-		newLogsCmd(),
+		newLogsCmd(client),
```
AmFlint, line 115: This is more of a general comment, but why don't we use a singleton variable inside the `houston` package to set a "global" HTTP Client `client`, instead of having to pass it everywhere in the project as a method parameter ?

Like:
```go
package "houston"

var Client *http.Client

// and initialize this variable either by setting it directly in the cmd when the program is initialized ?
```

I was also thinking about a different way to do it, but once again this is a more general discussion, outside the scope of the PR:
Right now, we're using the `houston` package as a low level HTTP client for the Houston API, each cmd has to re-create its own houston.Request object etc....

Why don't we make it a higher level abstraction over the Houston API ? It could expose methods like `FetchLogs(component string) (*returnType[], error)` or `CreateUser(email, password string) (*UserResponse, error)`, and all of the underlying implementations (like the HTTP Client and each cmd having to format the Request itself) could be done inside the `houston` package.

I've described this comment in the following notion document: https://www.notion.so/astronomerio/Review-CLI-27067e25a2314686b42d37e0bd9bf73e

What you do think @andriisoldatenko ?

AmFlint: Some good things about this approach:
- each component is de-coupled, the CLI needs something from the Houston API, it uses the houston package, gives it few parameters, and receive the response object without any knowledge about the underlying implementation or HTTP CLient manipulations.
- We can easily create a `type interface` for the houston package, and mock it for unit tests, instead of directly testing the http Responses in these CMD tests.
- We can write unit tests for the houston package independently and test the http behaviours.

cmd/logs.go
neel-astro: moved to a constant

cmd/logs.go
neel-astro: moved to a constant

cmd/logs.go
neel-astro: moved to a constant

cmd/logs_test.go
neel-astro: test case added

cmd/deployment.go
neel-astro: as discussed this change could be a part of separate PR

bote795 (approved):
lgtm

AmFlint (approved):

AmFlint:
Can we merge this PR please?

PR #495, opened 2022-02-02
Use system environment variables when building docker-compose

tkmcc:
## Description

Since version v0.27.2, host system environment variables are no longer able to be used in `docker-compose.override.yml` files for e.g. local development, this change passes along a pointer to the `compose-go` to use the system environment variables.

We have a local `docker-compose.override.yml` that tries to define volume which references a local path that includes an environment variable, like:

```
version: "2"
services:
  scheduler:
    volumes:
      - $HOME/myfile:/home/astro/myfile
 ```

As our team has gradually updated our `astro-cli` version, we noticed when running `astro dev` commands that we would get failures starting up because of a missing expected file and warnings that say `WARN[0000] The "HOME" variable is not set. Defaulting to a blank string.`

I believe the issue was introduced in the migration from `libcompose`, which handles parsing environment variables for you by default if you pass `Environment = nil`, and `compose-go`, which does not. Unfortunately I do not know go, but I managed to put this change together and it works locally. Please let me know if you have any pointers on how to add tests or make any further changes :) Thanks!

## 🎟 Issue(s)

Can't create issues on the repo externally.

## 🧪 Functional Testing

Create `docker-compose.override.yml` similar to above and try running `astro dev start` or `astro dev stop`

## 📸 Screenshots

image

## 📋 Checklist

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -15,6 +15,7 @@ import (
 	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/messages"
 
+	composeInterp "github.com/compose-spec/compose-go/interpolation"
 	"github.com/compose-spec/compose-go/loader"
 	composeTypes "github.com/compose-spec/compose-go/types"
 	"github.com/docker/cli/cli/config/configfile"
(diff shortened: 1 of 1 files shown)


neel-astro (approved):
Thanks for the contribution, lgtm 👍

PR #496, opened 2022-02-03
add afero.fs as an argument to fileutil.Exists

neel-astro:
## Description
Changes:
- Added `afero.FS`interface as an argument to `fileutil.Exists` method, to cover cases where file is created in-memory in some of the test cases and have to be checked in the in-memory fs created.

## 🎟 Issue(s)

Related astronomer/issues#4187

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -36,7 +36,7 @@ func initDirs(root string, dirs []string) error {
 		fullpath := filepath.Join(root, dir)
 
 		// Move on if already exists
-		_, err := fileutil.Exists(fullpath)
+		_, err := fileutil.Exists(fullpath, nil)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


bote795 (approved):

AmFlint:
I have a question regarding testing: Why don't we use the same FS as in production usage ?

Create a directory on the developer's FS etc...

Here we're running tests in a different environment (in-memory FS) than production and we might encounter production issues that have not been detected in tests because of this difference.

neel-astro:
I was there when these tests were written, but if I have to defend it then I would say that by using memory as FS, it easier to make sure each tests are independent when compared to create an actual file for a test in developer's FS, and also it makes writing a test that much simpler. Given that the only usecase of FS in our test is to store and retrieve the file content, the chances of it having a difference in production env compared to test case env are fairly minimal.

PR #502, opened 2022-02-10
Fix duplicated houston calls, initialization debug messages, get workspace query, skip version check for completion cmd

AmFlint:
## Description

In this PR, I'm fixing multiple issues:
- Initialization logs creating logs which might not be displayed properly since the `--verbosity` flag is not yet interpreted by the command (and set these logs to the `debug` level).
- Fix a houston call using the `workspaces` (plulary) GraphQL Query instead of the `workspace(workspaceUuid: ID!)` GraphQL query for the `GetWorkspace` method.
- Refactor the logic regarding `houston.GetAppConfig` since it's called many times in the project. For example, when CLI is initialized, we need to get the app config in order to set up additional flags based on the feature flags enabled in Houston. Here, I added a `caching` mechanism to make sure that we only make the GraphQL call once.
- Update/Add tests according to these changes
- Ignore root command's PersistentPreRunE inside `completion` sub-commands: allows users to generate shell completion without being authenticated to a cluster.

## 🎟 Issue(s)

- resolves https://github.com/astronomer/issues/issues/4147
- resolves https://github.com/astronomer/issues/issues/4220
- resolves https://github.com/astronomer/issues/issues/4241

## 📋 Checklist


cmd/deployment.go
```diff
@@ -134,9 +136,9 @@ func newDeploymentCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	var nfsMountDAGDeploymentEnabled, triggererEnabled, gitSyncDAGDeploymentEnabled bool
-	appConfig, err := houstonClient.GetAppConfig()
+	appConfig, err := deployment.GetAppConfig(houstonClient)
 	if err != nil {
-		fmt.Println("Error checking feature flag", err)
+		initDebugLogs = append(initDebugLogs, fmt.Sprintf("Error checking feature flag: %s", err.Error()))
```
neel-astro, line 141: Just realised, storing logs in memory can lead to other folks misusing it, and also increasing memory footprint of the CLI, without us having direct control over it, as some of these are errors are being thrown from external packages.

I think its not a bad idea to simply convert this to `logrus.Debugln` like we have done it for other similar error in this PR. These error checks where simply added as linter was recommending for it, and wasn't a request from customer end, and in case these logs are needed then one would be still be able to access it, by setting the `verbosity` as `debug` in the config.

neel-astro (changes requested):

cmd/deployment.go
AmFlint: The problem is, as discussed on Slack, this line of code here is executed before the `setupLogs` statement in PersistentPrerunE calls.

- CLI Initializes
- Get App Config
- Check Feature flags
- Add flags on the CLI if feature flag enabled in App Config
- Execute root command's PersistentPreRunE
- Execute the command

If there is an error in the `Get App Config`, the `logrus.Debugln` will not be printed when user specifies `--verbosity=debug` since root's PersistentPreRunE (and SetUpLogs) has not been executed yet. This is why we need to "buffer" debug logs produced before CLI initialization.

In order to fix the "memory footprint", I can just delete the slice from memory after the printDebugLogs occured.

deployment/config.go
```diff
@@ -0,0 +1,19 @@
+package deployment
+
+import "github.com/astronomer/astro-cli/houston"
+
+var (
+	appConfig    *houston.AppConfig
+	appConfigErr error
+)
+
+var GetAppConfig = func(client houston.ClientInterface) (*houston.AppConfig, error) {
```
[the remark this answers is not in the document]
AmFlint: I was thinking about doing something like that but I'm not sure the "Storage/caching" mechanism should be a part of this.

What if we want to reuse this houston.GetAppConfig later on to force a new call to get the app config ? 

Since this is not the case yet, I don't want to over-engineer the whole project, so I'll just change the code as you suggested. If in the futur we need to do a new call to Houston API without cache, we'll just add a bool param to the function.

cmd/deployment.go
neel-astro: cool 👍, also it would be great if we can add a comment around the `initDebugLogs` var (https://github.com/astronomer/astro-cli/pull/502/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR22), so that in future folks don't end up misinterpreting the usecase.

PR #503, opened 2022-02-10
Fix deploy link to point to airflow UI instead of astro UI

neel-astro:
## Description
Changes:
- Fixed the UI link in `astro deploy` output to point to airflow UI instead of astro UI

## 🎟 Issue(s)

Related astronomer/issues#3826

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -20,8 +20,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const airflowURLType = "airflow"
-
 var (
 	errNoWorkspaceID             = errors.New("no workspace id provided")
 	errNoDomainSet               = errors.New("no domain set, re-authenticate")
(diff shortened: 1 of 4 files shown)


AmFlint:
Maybe this URL should be provided by Houston.

And we could request it by sending a GraphQL Request to houston to get the deployment and its URLs (leading to the airflow deployment).

With the current solution, if we change something in the back-end, or if we introduce some configuration variables to allow customers to choose their URLs for airflow deployments, the CLI will not work anymore.

AmFlint:
Good!

We have to be careful when we merge, I updated the get deployment query in one of my PR I'll have to update it to include the URL as well

cmd/deploy.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const airflowURLType = "airflow"
```
AmFlint, line 23: I feel like this constant should be part of the `houston` package since it strictly relates to houston deployment URLs.

AmFlint (changes requested):
LGTM except for the airflowURL variable, which should be inside the houston package

cmd/deploy.go
neel-astro: moved to houston package

AmFlint (approved):
LGTM

PR #507, opened 2022-02-15
Fix command parsing logic

neel-astro:
## Description
Changes:
- Fixed command parsing logic in Podman to handle cases when there is space inside an escaped strings
- Matched the logic of webserver health check between docker & podman (`/health` is more reliable compared to `airflow db check`)

## 🎟 Issue(s)

Related astronomer/issues#4236

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


AmFlint:
Can we add some tests?

From what I can see it seems hard to add some test on the current code, maybe you could create a function for the command parsing and write test for it independently?

airflow/podman.go
```diff
@@ -175,12 +178,14 @@ func (p *Podman) Stop() error {
...
 func (p *Podman) ExecCommand(containerID, command string) string {
-	command = strings.TrimLeft(command, " ")
-	command = strings.TrimRight(command, " ")
 	execConfig := new(handlers.ExecCreateConfig)
 	execConfig.AttachStdout = true
 	execConfig.AttachStderr = true
-	execConfig.Cmd = strings.Split(command, " ")
+	cmd, err := parseCommand(command)
+	if err != nil {
+		return err.Error()
```
AmFlint, line 186: Why do we return err.Error() everywhere in this method ?

I mean, if docker daemon is not available or podman not up, we should return an error if this function isn't able to execute a command inside a container, shouldn't we ?

neel-astro: I was thinking about doing that 🤔, but I didn't do that just to avoid changing method signatures with re-convergence in mind, but I will add it now, given that we have already updated a few other method signatures.

AmFlint:
You need to rebase on main and recreate the ContainerHandler mock file, as I've just merged my PR which also changed the method signature on this same interface.

neel-astro:
Since our PRs are updating different methods in the mock file, hence git would be able to merge them without any conflicts.

settings/settings.go
```diff
@@ -204,15 +191,51 @@ func AddPools(containerEngine airflow.ContainerHandler, id string, airflowVersio
...
+					fmt.Printf("Added Pool: %s\n", pool.PoolName)
+				}
 			} else {
 				fmt.Printf("Skipping %s: Pool Slot must be set.\n", pool.PoolName)
 			}
 		}
 	}
 }
 
+func prepareAddConnCmd(baseAddCmd, connIDArg, connTypeArg, connURIArg, connExtraArg, connHostArg, connLoginArg, connPasswordArg, connSchemaArg, connPortArg string, conn *Connection) string {
```
AmFlint, line 207: I feel like this logic could be reworked, we have many different string variables being initialized and passed around.

I believe we can affort to wait for the CLI Convergence to fix/clean the code in the repository as we go through the migration.

AmFlint (approved):
LGTM

airflow/podman.go
```diff
@@ -404,8 +409,8 @@ func (p *Podman) webserverHealthCheck() {
 		if err != nil {
 			goto sleep
 		}
-		resp = p.ExecCommand(containerID, "airflow db check")
-		if strings.Contains(resp, "Connection successful.") {
+		resp, _ = p.ExecCommand(containerID, "curl --fail http://127.0.0.1:8080/health")
```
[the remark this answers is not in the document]
andriisoldatenko: let's extract to constant variable.

andriisoldatenko:
@neel-astro  can you please add more tests? or somehow to increase test coverage?

settings/settings.go
andriisoldatenko: +1 to rework this part.

neel-astro:
@andriisoldatenko somehow codecov numbers don't make sense, adding legit test cases seems to be bringing down the coverage change percentage 😅 . Also, the coverage change percentage seems to be negative even relative change is a positive number. Seems to be happening because the base commit used for comparison by the codecov isn't accurate when doing it for the PR, but is accurate when manually comparing branches and shows an increase in coverage.

Codecov report for PR: https://app.codecov.io/gh/astronomer/astro-cli/compare/507/diff
Codecov report when comparing branch manually: https://app.codecov.io/gh/astronomer/astro-cli/compare/main...fix%2Fpodman-airflow-settings/diff

andriisoldatenko (approved):

PR #508, opened 2022-02-16
Fix: Print friendly message when server and CLI version mismatch (GraphQL Schema errors)

AmFlint:
## Description

When the CLI and server version mismatch, users encounter "hard-to-understand" errors like the following:
```bash
astro user create
Error checking feature flag API error (400): {"errors":[{"message":"Cannot query field \"triggererEnabled\" on type \"AppConfig\". Did you mean \"prometheusEnabled\" or \"elasticSearchEnabled\"?","locations":[{"line":12,"column":4}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}},{"message":"Cannot query field \"featureFlags\" on type \"AppConfig\".","locations":[{"line":13,"column":4}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}}]}
```

As the above error states, the CLI requests certain fields (`triggererEnabled` in this example) which are not supported in the server's GraphQL schema. But this error is not friendly and the user can hardly understand how to fix it.

In this PR, I added error management to return a friendly message to the user, and also log the base error to the debug level so that the user can still access the error trace with the `--verbosity=debug` flag.

Now the error looks like this:
```bash
./astro user create
Error:
  Some fields requested by the CLI are not available in the server schema, it seems there is a version mismatch between your CLI and the server.
  In order to fix this issue, you might want to use the same version for both the server and the CLI.
  You can use the command "astro version" to check your CLI and server versions.
  You can use the flag --verbosity=debug to get the full error trace.
```

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/3964

## 🧪 Functional Testing

- To test this fix, you'll need to use an old version of Houston, in my case, I used release-0.25 locally and set up my astro CLI to use localhost.
- Then, make sure you start the authentication process:
```bash
astro auth login localhost
# in my case, i had just initialized houston locally, so I had to create a user, but to do this with the CLI, I had to start the auth process
```
- Then, run whatever command you'd like, in my case, I needed to create the first user in fresh houston:
```bash
astro user create
Error checking feature flag Some fields requested by the CLI are not available in the server schema.
Error checking feature flag Some fields requested by the CLI are not available in the server schema.
Error checking feature flag Some fields requested by the CLI are not available in the server schema.
Error checking feature flag Some fields requested by the CLI are not available in the server schema.
Error:
  Some fields requested by the CLI are not available in the server schema, it seems there is a version mismatch between your CLI and the server.
  In order to fix this issue, you might want to use the same version for both the server and the CLI.
  You can use the command "astro version" to check your CLI and server versions.
  You can use the flag --verbosity=debug to get the full error trace.

Usage:
  astro user create [flags]

Aliases:
  create, cr

Flags:
  -e, --email string      Supply user email at runtime
  -h, --help              help for create
  -p, --password string   Supply user password at runtime

Global Flags:
      --skip-version-check   skip version compatibility check
      --verbosity string     Log level (debug, info, warn, error, fatal, panic (default "warning")
```

**Warning**: The first 4 `Error checking for feature flags` have been moved to "debug level" in this PR: https://github.com/astronomer/astro-cli/pull/502

## 📋 Checklist

### DIFF
--- a/houston/errors.go
+++ b/houston/errors.go
@@ -1,6 +1,27 @@
 package houston
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
+
+type ErrFieldsNotAvailable struct {
+	BaseError error
+}
+
+func (e ErrFieldsNotAvailable) Error() string {
+	return "Some fields requested by the CLI are not available in the server schema."
+}
+
+func handleAPIErr(err error) error {
+	if strings.Contains(err.Error(), "Cannot query field") {
+		return ErrFieldsNotAvailable{
+			BaseError: err,
+		}
+	}
+
+	return err
+}
 
 type ErrWorkspaceNotFound struct {
 	workspaceID string
--- a/version/errors.go
+++ b/version/errors.go
@@ -5,7 +5,7 @@ type ErrVersionMismatch struct{}
 func (e ErrVersionMismatch) Error() string {
 	return `
   Some fields requested by the CLI are not available in the server schema, it seems there is a version mismatch between your CLI and the server.
-  In order to fix this issue, you might want to use the same version for both the server and the CLI.
+  In order to fix this issue, you might want to use the same minor version for both the server and the CLI.
   You can use the command "astro version" to check your CLI and server versions.
   You can use the flag --verbosity=debug to get the full error trace.
 	`
--- a/version/errors.go
+++ b/version/errors.go
@@ -1,0 +1,12 @@
+package version
+
+type ErrVersionMismatch struct{}
+
+func (e ErrVersionMismatch) Error() string {
+	return `
+  Some fields requested by the CLI are not available in the server schema, it seems there is a version mismatch between your CLI and the server.
+  In order to fix this issue, you might want to use the same version for both the server and the CLI.
+  You can use the command "astro version" to check your CLI and server versions.
+  You can use the flag --verbosity=debug to get the full error trace.
+	`
+}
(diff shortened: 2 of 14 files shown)


version/errors.go
```diff
@@ -0,0 +1,12 @@
+package version
+
+type ErrVersionMismatch struct{}
+
+func (e ErrVersionMismatch) Error() string {
+	return `
+  Some fields requested by the CLI are not available in the server schema, it seems there is a version mismatch between your CLI and the server.
+  In order to fix this issue, you might want to use the same version for both the server and the CLI.
```
vishwas-astro, line 8: same minor version (example 0.27.3 of CLI with 0.27.2 of platform) and not the exact same full SemVer.

houston/errors.go
```diff
@@ -1,6 +1,27 @@
...
+	"fmt"
+	"strings"
+)
+
+type ErrFieldsNotAvailable struct {
+	BaseError error
+}
+
+func (e ErrFieldsNotAvailable) Error() string {
+	return "Some fields requested by the CLI are not available in the server schema."
```
neel-astro, line 13: nitpicking: should we log the actual error received from Houston as well, to be able to root cause the issue, since this log will only be present in case the verbosity is set to debug.

neel-astro (approved):
looks fine to me 👍, have added a minor comment do have a look at it.

houston/errors.go
AmFlint: Just above, the `ErrFieldsNotAvailable` error struct has a property `BaseError`.

The houston package should not be responsibel of logging errors, this is the responsibility of the package calling houston methods.

So in this case, I added the `BaseError`, so that in the version package I can do an error casting and logrus.Debug the base error so that the user is still able to read the houston error.

PR #510, opened 2022-02-17
Fix: Set Celery as default executor for astro deployment create cmd

AmFlint:
## Description

- Set `celery` as default value for `executor` type on the `astro deployment create` command. Previously, if user did not provide the `--executor` flag, the command threw an error. However, the doc for this commands states that the default value for --executor is celery, and this is the behaviour in the UI.
- Enforce use of constants for houston-related values (Deployment Types, RBAC roles, Executor types)...

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/3711

## 🧪 Functional Testing

- try and create a new deployment without specifying the --executor flag:
```bash
astro deployment create my-deployment
```
This should create the deployment successfully, with the `Celery` executor type.

## 📋 Checklist

### DIFF
--- a/cmd/deployment.go
+++ b/cmd/deployment.go
@@ -17,16 +17,25 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	celeryExecutorArg     = "celery"
+	localExecutorArg      = "local"
+	kubernetesExecutorArg = "kubernetes"
+	k8sExecutorArg        = "k8s"
+)
+
 var (
 	errUpdateDeploymentInvalidArgs = errors.New("must specify a deployment ID and at least one attribute to update")
 	errServiceAccountNotPresent    = errors.New("must provide a service-account label with the --label (-l) flag")
 )
 
 var (
-	allDeployments        bool
-	cancel                bool
-	hardDelete            bool
-	executor              string
+	allDeployments bool
+	cancel         bool
+	hardDelete     bool
+	executor       string
+	// have to use two different executor flags for create and update commands otherwhise both commands override this value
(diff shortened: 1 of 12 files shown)


cmd/deployment.go
```diff
@@ -15,16 +15,25 @@ import (
...
-	allDeployments        bool
-	cancel                bool
-	hardDelete            bool
-	executor              string
+	allDeployments bool
+	cancel         bool
+	hardDelete     bool
+	executor       string
+	// have to use two different executor flags for create and update commands otherwhise both commands override this value
+	executorUpdate        string
```
neel-astro, line 36: from this PR changes, I didn't get why we need two separate executor flags for create & update deployment because in any given execution only one of the two commands would be executed 🤔 , can you explain further.

AmFlint: I tested this change locally and I got into an error where `executor` was `""` (empty string).

That's because in both `update` and `create` we write the line:
```go
cmd.Flags().StringVarP(&executor, "executor", "e", "", "Add executor parameter: local, celery, or kubernetes")
```
This flag --executor is handled 2 times, and the declaration in the "update" method overrides the one in create deployment, which means the --executor flag will ALWAYS default to empty string.

If I change it to default to "celery" in both places, the update method will always have the `executor` variable populated with the default value "celery", even when the user does not want to update his deployment's executor configuration. This would lead to unwanted side-effects, so the best solution to avoid bugs is to decouple the executor variable for both commands.

neel-astro (approved):
lgtm 👍

PR #514, opened 2022-02-23
Move to multi auth config model for Podman `astro deploy`

neel-astro:
## Description
Changes:
- Moved to multi auth config model instead of passing our own token, where underlying Podman library in case no token is passed will encode all the content of all auth file present in the local dev's machine and pass it to libpod API.

## 🎟 Issue(s)

Related astronomer/issues#4282

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


AmFlint:
Can you provide more context on why we need to make these changes in the podman adapter ?

neel-astro:
@AmFlint 
> Can you provide more context on why we need to make these changes in the podman adapter ?

Adding the overall workflow:
When a user login via `astro auth login` internally we would also login to cluster registry, via Podman or Docker based on config. When Podman/Docker logins into the registry it will store a token in their own auth config file, which is usually at `~/.config/containers/auth.json` for Podman & `~/.docker/config.json` for Docker.
Then when we try and push an image to the registry, we would either need to pass a token provided from the registry for authentication, or else in the case of Podman, if we don't send any token, then Podman library internally reads all the auth config files (both Podman's & Docker's) and pick up the auth based on registry domain.
Till now we used to send the token in our request, but the internal libpod API (Podman internally calls libpod API to orchestrate and change) wasn't able to validate our token on Mac OS with just Podman installed, most likely because Podman runs inside a VM on Mac. Hence moved to the other model of authentication, where Podman will simply read all the auth configs and figure out the token on its own.

andriisoldatenko (approved):
code LGTM

PR #515, opened 2022-02-24
Fix dev logs logic for custom container name

neel-astro:
## Description
Changes:
- Fix `astro dev logs` returning no logs in case of using custom container names for airflow components when using Docker container engine.

## 🎟 Issue(s)

Related astronomer/issues#4284

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/container.go
+++ b/airflow/container.go
@@ -205,3 +205,39 @@ func getFmtEnvFile(envFile string, containerEngine Container) (string, error) {
 		return "", nil
 	}
 }
+
+func GetWebserverServiceName() string {
+	containerEngine := config.CFG.ContainerEngine.GetString()
+	switch containerEngine {
+	case string(DockerEngine):
+		return webserverServiceName
+	case string(PodmanEngine):
+		return config.CFG.WebserverContainerName.GetString()
+	default:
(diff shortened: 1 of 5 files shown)


airflow/container.go
```diff
@@ -205,3 +205,39 @@ func getFmtEnvFile(envFile string, containerEngine Container) (string, error) {
...
+	}
+}
+
+func GetTriggererServiceName() string {
+	containerEngine := config.CFG.ContainerEngine.GetString()
+	switch containerEngine {
+	case string(DockerEngine):
+		return triggererServiceName
+	case string(PodmanEngine):
+		return config.CFG.TriggererContainerName.GetString()
```
AmFlint, line 239: Why is it different in Podman ?

If we run multiple projects locally with Podman, we do not encounter the same "container name already exists" ?

neel-astro: With Podman the entire logic is based on the container names, which could be configured by the user, whereas for Docker the logic is based on the service name, which is a static value in the docker-compose file.

AmFlint (approved):

andriisoldatenko (approved):

PR #516, opened 2022-02-24
Initialize projects using Astro runtime images

AmFlint:
## Description

With the recent additions of Astro runtimes, we want to allow users to initialize new airflow projects using the Astro runtime as a base image instead of the Astronomer Certified Airflow images.

By default, we want to initialize a project with Astro runtime images, but still allow users to use Astronomer Certified images by providing a flag like:
```bash
# this will initialize a project using Astronomer Certified Airflow images, which is the behaviour in current releases 0.28.* and below.
astro dev init --use-astronomer-certified
```

**This PR should not be merged before Astro Runtime is supported on all components of the platform (Houston etc...)**.

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/4253

## 🧪 Functional Testing

To test this feature:
- create a new directory 
- Initalize a new project with default latest astronomer runtime tag/version:
```bash
# this should initialize a project using the latest "stable" Astro Runtime tag
astro dev init

# you can check it by looking at the content of the generated Dockerfile
cat Dockerfile
# image should be astro-runtime:whichever version is the latest stable one
```

You can also try to generate a project using a specific airflow version:
```bash
astro dev init -v 2.2.0
```

Also, try and generate a project using Astronomer Certified images:
```bash
astro dev init --use-astronomer-certified

# with given airflow version
astro dev init --use-astronomer-certified -v 2.0.0
```

## 📋 Checklist

### DIFF
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -76,14 +76,14 @@ func initFiles(root string, files map[string]string) error {
 }
 
 // Init will scaffold out a new airflow project
-func Init(path, airflowImageTag string) error {
+func Init(path, airflowImageName, airflowImageTag string) error {
 	// List of directories to create
 	dirs := []string{"dags", "plugins", "include"}
 
 	// Map of files to create
 	files := map[string]string{
 		".dockerignore":         include.Dockerignore,
-		"Dockerfile":            fmt.Sprintf(include.Dockerfile, airflowImageTag),
+		"Dockerfile":            fmt.Sprintf(include.Dockerfile, airflowImageName, airflowImageTag),
 		".gitignore":            include.Gitignore,
 		"packages.txt":          "",
 		"requirements.txt":      "",
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -21,13 +21,16 @@ const (
 	defaultAirflowVersion = uint64(0x1) //nolint:gomnd
 	componentName         = "airflow"
 
-	airflowVersionLabelName        = "io.astronomer.docker.airflow.version"
 	triggererAllowedAirflowVersion = "2.2.0"
+	triggererAllowedRuntimeVersion = "4.0.0"
 
 	webserverHealthCheckInterval = 10 * time.Second
 )
 
-var repoNameSanitizeRegexp = regexp.MustCompile(`^[^a-z0-9]*`) // must not start with anything except lowercase letter or number
+var (
+	repoNameSanitizeRegexp = regexp.MustCompile(`^[^a-z0-9]*`) // must not start with anything except lowercase letter or number
+	runtimeVersionCheck    = fmt.Sprintf(">= %s", triggererAllowedRuntimeVersion)
+)
 
 func initDirs(root string, dirs []string) error {
 	// Create the dirs
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -27,10 +27,7 @@ const (
 	webserverHealthCheckInterval = 10 * time.Second
 )
 
-var (
-	repoNameSanitizeRegexp = regexp.MustCompile(`^[^a-z0-9]*`) // must not start with anything except lowercase letter or number
-	runtimeVersionCheck    = fmt.Sprintf(">= %s", triggererAllowedRuntimeVersion)
-)
+var repoNameSanitizeRegexp = regexp.MustCompile(`^[^a-z0-9]*`) // must not start with anything except lowercase letter or number
 
 func initDirs(root string, dirs []string) error {
 	// Create the dirs
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -22,7 +22,7 @@ const (
 	componentName         = "airflow"
 
 	triggererAllowedAirflowVersion = "2.2.0"
-	triggererAllowedRuntimeVersion = "4.0.0"
+	triggererAllowedRuntimeVersion = "4.2.5"
 
 	webserverHealthCheckInterval = 10 * time.Second
 )
--- a/airflow/container.go
+++ b/airflow/container.go
@@ -130,7 +130,6 @@ func generateConfig(projectName, airflowHome, envFile string, imageLabels map[st
 		return "", err
 	}
 
-	// check if Airflow triggerer is enabled, on AC images, look if version is > 2.1.0, for runtime if version >= 4.1.0
 	triggererEnabled, err := CheckTriggererEnabled(imageLabels)
 	if err != nil {
 		return "", err
@@ -273,7 +272,7 @@ var CheckTriggererEnabled = func(imageLabels map[string]string) (bool, error) {
 	runtimeVersion, ok := imageLabels[runtimeVersionLabelName]
 	if !ok {
 		// image doesn't have either runtime version or airflow version
-		// we don't want to block the user's experience in case this happens, so we disabled triggerer and warn error
+		// we don't want to block the user's experience in case this happens, so we disable triggerer and warn error
 		fmt.Println(messages.WarningTriggererDisabledNoVersionDetected)
 
 		return false, nil
--- a/airflow/container.go
+++ b/airflow/container.go
@@ -274,6 +274,9 @@ var CheckTriggererEnabled = func(imageLabels map[string]string) (bool, error) {
 	runtimeVersion, ok := imageLabels[runtimeVersionLabelName]
 	if !ok {
 		// image doesn't have either runtime version or airflow version
+		// we don't want to block the user's experience in case this happens, so we disabled triggerer and warn error
+		fmt.Println(messages.WarningTriggererDisabledNoVersionDetected)
+
 		return false, nil
 	}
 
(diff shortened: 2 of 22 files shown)


neel-astro (commented):
overall LGTM 👍 , have added few minor comments

airflow_versions/airflow_versions.go
```diff
@@ -17,9 +27,60 @@ func GetDefaultImageTag(httpClient *Client, airflowVersion string) (string, erro
...
+		return "", ErrNoTagAvailable{airflowVersion: airflowVersion}
+	}
+
+	// get latest runtime version
+	latestVersion, _ := NewAirflowVersion(tagsToUse[0], []string{tagsToUse[0]})
+	for i := 1; i < len(tagsToUse); i++ {
+		tag := tagsToUse[i]
+		nextVersion, err := NewAirflowVersion(tag, []string{tag})
+		if err == nil {
+			if nextVersion.GreaterThan(latestVersion) {
```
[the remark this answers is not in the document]
AmFlint: I went to read the cloud-cli code for this same feature, they look at the runtime version (5.0.0 which is in an `alpha` channel for now and can't be used for production purposes).

Just like I do, I just added another check for the runtime version's release channel to only allow users to use stable releases.

Anyway, we probably won't merge this PR into this repository, since we'll do the convergence soon, and I'll probably just submit a PR to allow AC images and default to AC images if context is software (CLI logged into a gen1 cluster).

But I guess that most of this PR will be dropped since cloud-cli already developped this feature.

airflow/container.go
```diff
@@ -131,10 +134,10 @@ func generateConfig(projectName, airflowHome, envFile string, imageLabels map[st
 		return "", err
 	}
 
-	var triggererEnabled bool
-	airflowVersion := imageLabels[airflowVersionLabelName]
-	if versions.GreaterThanOrEqualTo(airflowVersion, triggererAllowedAirflowVersion) {
-		triggererEnabled = true
+	// check if Airflow triggerer is enabled, on AC images, look if version is > 2.1.0, for runtime if version >= 4.1.0
+	triggererEnabled, err := CheckTriggererEnabled(airflowHome, dockerfile, runtimeVersionCheck)
```
neel-astro, line 138: any reason to move away from image labels to parsing to the dockerfile? I feel picking version from image label is more reliable compared to parsing dockerfile

airflow_versions/airflow_versions.go
```diff
@@ -17,9 +27,60 @@ func GetDefaultImageTag(httpClient *Client, airflowVersion string) (string, erro
...
+				latestVersion = nextVersion
+			}
+		}
```
neel-astro, line 74:
```suggestion
		if err == nil && nextVersion.GreaterThan(latestVersion) {
			latestVersion = nextVersion
		}
```

neel-astro (commented):
overall looks fine, have added few minor comments

airflow/container.go
AmFlint: Why is it more reliable ? The Airflow version is added to 2 places:
- On the actual tag of the image, which can be found by simply reading from a file, which is always present inside an astronomer/airflow project (Dockerfile)
- The label on the image, which needs to be first pulled, then docker inspect.

Also, I made the decision of moving to reading Dockerfile since runtime code in Cloud-cli already uses this approach, and I wanted to use a consistent approach for both AC and Runtimes.

neel-astro: > Why is it more reliable? 

Because users can use custom images built on top of AC or Runtime image, in which case it is not necessary for users will carry forward the same version tag, but those images will inherit label from the base image.

> The label on the image, which needs to be first pulled, then docker inspect

We don't need to separately pull the image in this case, since we are already building the image from Dockerfile beforehand.

AmFlint: OK I see, in that case I'll use labels since customer can use a custom image.

airflow/airflow.go
```diff
@@ -21,13 +21,16 @@ const (
 	defaultAirflowVersion = uint64(0x1) //nolint:gomnd
 	componentName         = "airflow"
 
-	airflowVersionLabelName        = "io.astronomer.docker.airflow.version"
 	triggererAllowedAirflowVersion = "2.2.0"
+	triggererAllowedRuntimeVersion = "4.0.0"
```
andriisoldatenko, line 25:
```suggestion
	triggererAllowedRuntimeVersion = "4.2.5"
```

airflow/container.go
andriisoldatenko: @AmFlint  / @neel-astro may be it can help https://github.com/astronomer/astro-runtime/blob/main/image/Dockerfile#L31

AmFlint: I'll take a look at this to see if it's reliable for runtimes too

airflow/airflow.go
AmFlint: I think the triggerer resource was introduced in runtime v 4.0.0

andriisoldatenko: it's less important, Gen1 works only starting 4.2.5.

andriisoldatenko: https://github.com/astronomer/houston-api/blob/main/src/lib/constants/index.js#L209 i still need to update it here after i merge and release https://github.com/astronomer/astro-runtime/pull/235

airflow/container.go
AmFlint: Added for both AC and Runtimes

neel-astro (changes requested):
looks fine apart from one minor comment, also some tests are failing which you might want to have a look at

airflow/podman.go
```diff
@@ -345,15 +340,16 @@ func (p *Podman) startPod(projectName, projectDir, envFile string, imageLabels m
...
+func generatePodState(projectName, projectDir, envFile string) error {
 	// Create the podman config yaml if not exists
 	if _, err := os.Stat(podConfigFile); errors.Is(err, os.ErrNotExist) {
 		err = initFiles("", map[string]string{podConfigFile: include.PodmanConfigYml})
 		if err != nil {
 			return err
 		}
 	}
-	configYAML, err := generateConfig(projectName, projectDir, envFile, imageLabels, PodmanEngine)
+	labels := map[string]string{airflowVersionLabelName: triggererAllowedAirflowVersion}
```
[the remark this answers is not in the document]
AmFlint: Local testing, I fixed it sorry about that

neel-astro (approved):
lgtm

neel-astro (approved):

andriisoldatenko:
> By default, we want to initialize a project with Astro runtime images

Yes

Regarding:
>By default, we want to initialize a project with Astro runtime images
I think you need to check feature flag ` "astroRuntimeEnabled": true` otherwise default to AC 

```
{
  "data": {
    "appConfig": {
      "version": "0.27.1-103-g86f6bfb7-build38298",
      "baseDomain": "dev.astrodev.io",
      "smtpConfigured": true,
      "configureDagDeployment": true,
      "nfsMountDagDeployment": true,
      "manualReleaseNames": false,
      "prometheusEnabled": true,
      "elasticSearchEnabled": true,
      "hardDeleteDeployment": true,
      "manualNamespaceNames": false,
      "appSubName": "Astronomer Software",
      "featureFlags": {
        "manualReleaseNames": false,
        "configureDagDeployment": true,
        "nfsMountDagDeployment": true,
        "hardDeleteDeployment": true,
        "manualNamespaceNames": false,
        "namespaceFreeFormEntry": false,
        "kibanaUIEnabled": true,
        "grafanaUIEnabled": true,
        "triggererEnabled": false,
        "sysAdminScalabilityImprovementsEnabled": true,
        "allowedSystemLevelDomains": [
          "astronomer.io"
        ],
        "allowedSystemLevelDomainsEnabled": true,
        "gitSyncDagDeployment": true,
        "idpGroupsEnabled": true,
        "astroRuntimeEnabled": true
      },
     ....
        "statusPage": null
      },
      "__typename": "AppConfig"
    }
  }
}
``` 

wdyt @cmarteepants ?

andriisoldatenko:
@AmFlint pls fix conflicts / and add more tests.

AmFlint:
I can add the logic to check for the astroRuntimeEnabled feature flag if you want.

In the converged cli, we removed all logic coupled to Houston in dev commands, since it's used to develop locally, we don't want to tie it to the Astronomer Software.

However, in this CLI, it's only used by Software users, so I think it makes sense to check this feature flag.

andriisoldatenko:
> this

correct, i'm worried if customer don't want to enable runtime and cli will automatically init runtime based image, and than he can deploy it by mistake.

andriisoldatenko (approved):

andriisoldatenko:
@AmFlint  / @neel-astro 

practically LGTM:
```
➜  t_20220413_130921 astro-cli dev init
Initialized empty astronomer project in /Users/andrii/workspace/src/github.com/astronomer/t_20220413_130921
➜  t_20220413_130921 cat Dockerfile
FROM quay.io/astronomer/astro-runtime:4.2.5%
➜  t_20220413_130921
```

Issue #627, opened 2022-06-29
`--image-name` flag for `astro dev start`


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.*

make it so users can run `astro dev start` with a custom image

**🧩 Describe the solution you'd like**

The custom image name is passed to `dev start` with `astro dev start --image-name  <custom image  name>`

**🤔 Describe alternatives you've considered**

**💡 Additional context**

productboard feature: https://astronomer.productboard.com/feature-board/2193567-step-1-feature-backlog/features/13905313/detail

PR #634, opened 2022-07-06
redo image-name feature

sunkickr:
## Description

update --image-name feature to include pytest and parse commands

## 🎟 Issue(s)

-  https://github.com/astronomer/astro-cli/issues/627

## 🧪 Functional Testing

- added tests to increase coverage

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -39,8 +39,17 @@ func DockerImageInit(image string) *DockerImage {
 	return &DockerImage{imageName: image}
 }
 
-func (d *DockerImage) Build(config airflowTypes.ImageBuildConfig) error {
+func (d *DockerImage) Build(localImage string, config airflowTypes.ImageBuildConfig) error {
 	// Change to location of Dockerfile
+	if localImage != "" {
+		// skip build if an local image is passed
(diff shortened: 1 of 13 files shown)


airflow/docker_image.go
```diff
@@ -39,8 +39,17 @@ func DockerImageInit(image string) *DockerImage {
 	return &DockerImage{imageName: image}
 }
 
-func (d *DockerImage) Build(config airflowTypes.ImageBuildConfig) error {
+func (d *DockerImage) Build(localImage string, config airflowTypes.ImageBuildConfig) error {
 	// Change to location of Dockerfile
+	if localImage != "" {
```
neel-astro, line 44: I think this logic should not be part of the Docker image build method, it is a business logic around how the image should be built for `astro dev` command, so it might be more accurate for this to remain in the `airflow/docker.go` itself like before, and thus keeping `ImageHandler` interface agnostic of images.

neel-astro (changes requested):

airflow/docker_image.go
sunkickr: If that's the case I'll probably just scrap this PR

PR #654, opened 2022-07-20
start up docker

sunkickr:
## Description

Automatically start docker if it is not running on mac machines

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/656

## 🧪 Functional Testing

need to write tests

## 📸 Screenshots

Screen Shot 2022-07-22 at 2 59 51 PM

## 📋 Checklist

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -72,9 +72,9 @@ var (
 	inspectContainer = inspect.Inspect
 	initSettings     = settings.ConfigSettings
 
-	openURL = browser.OpenURL
-	timeoutSeconds = time.Duration(60)
-	tickMS = time.Duration(500)
+	openURL    = browser.OpenURL
+	timeoutNum = 60
+	tickNum    = 500
 )
 
 // ComposeConfig is input data to docker compose yaml template
@@ -705,15 +705,15 @@ func startDocker() error {
 }
 
 func waitForDocker(buf io.Writer) error {
-	timeout := time.After(timeoutSeconds * time.Second)
-	ticker := time.Tick(tickMS * time.Millisecond)
+	timeout := time.After(time.Duration(timeoutNum) * time.Second)
+	ticker := time.NewTicker(time.Duration(tickNum) * time.Millisecond)
 	for {
 		select {
 		// Got a timeout! fail with a timeout error
 		case <-timeout:
 			return errors.New("timed out waiting for docker")
 		// Got a tick, we should check on checkSomething()
-		case <-ticker:
+		case <-ticker.C:
 			err := cmdExec(DockerCmd, buf, buf, "ps")
 			if err != nil {
 				continue
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -686,7 +686,7 @@ func startDocker() error {
 	if err != nil {
 		// open docker
 		fmt.Println("\nDocker is not running. Starting up the Docker engine…")
-		err = cmdOpenExec(OpenCmd, buf, os.Stderr, "-a", "docker")
+		err = cmdExec(OpenCmd, buf, os.Stderr, "-a", "docker")
 		if err != nil {
 			return err
 		}
@@ -714,7 +714,7 @@ func waitForDocker() error {
 		// Got a tick, we should check on checkSomething()
 		case <-ticker.C:
 			buf.Reset()
-			err := cmdPsExec(DockerCmd, buf, buf, "ps")
+			err := cmdExec(DockerCmd, buf, buf, "ps")
 			if err != nil {
 				continue
 			} else {
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"runtime"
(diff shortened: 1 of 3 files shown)


kushalmalani (approved):

airflow/docker_image.go
```diff
@@ -26,6 +26,7 @@ import (
 const (
 	DockerCmd          = "docker"
 	EchoCmd            = "echo"
+	OpenCmd            = "open"
```
neel-astro, line 29: nitpick: This should be in `docker.go` where it is being used

airflow/docker.go
```diff
@@ -142,6 +150,14 @@ func DockerComposeInit(airflowHome, envFile, dockerfile, imageName string, isPyT
 
 // Start starts a local airflow development cluster
 func (d *DockerCompose) Start(imageName string, noCache bool) error {
+	// check of docker is up for mac
```
neel-astro, line 153:
```suggestion
	// check if docker is up for macOS
```

airflow/docker.go
```diff
@@ -663,3 +679,46 @@ func checkServiceState(serviceState, expectedState string) bool {
...
+	timeout := time.After(time.Duration(timeoutNum) * time.Second)
+	ticker := time.NewTicker(time.Duration(tickNum) * time.Millisecond)
+	for {
+		select {
+		// Got a timeout! fail with a timeout error
+		case <-timeout:
+			return errors.New("timed out waiting for docker")
+		// Got a tick, we should check on checkSomething()
+		case <-ticker.C:
+			err := cmdPsExec(DockerCmd, buf, buf, "ps")
```
neel-astro, line 716:
```suggestion
                        buf.Reset()
			err := cmdPsExec(DockerCmd, buf, buf, "ps")
```
nitpick: It would be ideal to reset the buffer, since that's stored in memory, and it would keep acquiring memory that it does not require. Also because the `buf` is not used after that.

Note: this would require `buf` to be of type `bytes.Buffer` and not `io.Writer`. Should work fine if this is implemented with the suggestion above this.

airflow/docker.go
```diff
@@ -663,3 +679,46 @@ func checkServiceState(serviceState, expectedState string) bool {
...
+		// poll for docker
+		err = waitForDocker(buf)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func waitForDocker(buf io.Writer) error {
```
neel-astro, line 706:
```suggestion
func waitForDocker() error {
    buf := new(bytes.Buffer)
```
nitpick: `buf` is not a required argument to be passed, since that's not being used after anywhere in the caller method.

airflow/docker.go
```diff
@@ -69,7 +72,12 @@ var (
 	inspectContainer = inspect.Inspect
 	initSettings     = settings.ConfigSettings
 
-	openURL = browser.OpenURL
+	openURL    = browser.OpenURL
+	timeoutNum = 60
+	tickNum    = 500
+
+	cmdPsExec   = cmdExec
+	cmdOpenExec = cmdExec
```
neel-astro, line 80: why do we need this? cmdExec is already a variable in itself, so we can directly update that in our unit test. Not defining this, will reduce one extra hop for users while reading the code.

neel-astro (commented):
looks fine, have added a few minor comments

Haven't really thought through it, but do you think we can do something similar for Linux OS as well?

airflow/docker.go
sunkickr: cmdExec is ran three times during the function startDocker. Each time it needs a different output not sure how to accomplish this within the unit test. All instances seem cmdExec take on the the last definition

sunkickr:
> looks fine, have added a few minor comments
> 
> Haven't really thought through it, but do you think we can do something similar for Linux OS as well?

@neel-astro I think it would be exactly the same for linux just not sure what command one would run to start docker in linux

airflow/docker.go
neel-astro: in that case, the unit test can be modified as:
```Golang
t.Run("start docker success", func(t *testing.T) {
		cmdExec = func(cmd string, stdout, stderr io.Writer, args ...string) error {
                         switch cmd {
                         case "open":
                             return nil
                         case "ps":
                             return nil
                         default:
                             return errExecMock
                         }
		}

		err := startDocker()
		assert.NoError(t, err)
	})

	t.Run("start docker fail", func(t *testing.T) {
		timeoutNum = 5

		cmdExec = func(cmd string, stdout, stderr io.Writer, args ...string) error {
                         switch cmd {
                         case "open":
                             return nil
                         case "ps":
                             return errExecMock
                         default:
                             return errExecMock
                         }
		}
		err := startDocker()
		assert.Contains(t, err.Error(), "timed out waiting for docker")
	})
```

sunkickr: Thanks!

airflow/docker.go
```diff
@@ -663,3 +676,47 @@ func checkServiceState(serviceState, expectedState string) bool {
...
+func waitForDocker() error {
+	buf := new(bytes.Buffer)
+	timeout := time.After(time.Duration(timeoutNum) * time.Second)
+	ticker := time.NewTicker(time.Duration(tickNum) * time.Millisecond)
+	for {
+		select {
+		// Got a timeout! fail with a timeout error
+		case <-timeout:
+			return errors.New("timed out waiting for docker")
+		// Got a tick, we should check on checkSomething()
```
neel-astro, line 711:
```suggestion
		// Got a tick, we should check if docker is up & running
```

neel-astro (approved):

PR #662, opened 2022-07-27
Integrated workspace paginated query

ajayy004:
## Description

> Client is having scale issues with our workspace users query. This PR adds a paginated query.
The current default behavior still uses the older workspace implementation which is without pagination, workspace pagination is only enabled through flag `-p` to enable pagination with default `100` page size and `-s 10` to set the page size

## 🎟 Issue(s)

Related 
- https://github.com/astronomer/issues/issues/4855
- https://github.com/astronomer/issues/issues/4579

## 🧪 Functional Testing

## 📸 Screenshots

#### With pagination and page-size 
Screen Shot 2022-08-09 at 10 41 01 AM

#### With config page size and interactive already set
Screen Shot 2022-08-09 at 10 48 06 AM

## 📋 Checklist


software/workspace/workspace.go
```diff
@@ -102,11 +116,57 @@ func GetCurrentWorkspace() (string, error) {
 	return c.Workspace, nil
 }
 
-func getWorkspaceSelection(client houston.ClientInterface, out io.Writer) (string, error) {
+// WorkspacesPromptPaginatedOption Show pagination option based on page size and total record
+var WorkspacesPromptPaginatedOption = func(pageSize, pageNumber, totalRecord int) PaginationOptions {
```
neel-astro, line 120:
```suggestion
var workspacesPromptPaginatedOption = func(pageSize, pageNumber, totalRecord int) PaginationOptions {
```

function doesn't needs to be public, since it is only referred inside this package

neel-astro (changes requested):
Had a high-level look, and have added a few minor comments

cmd/auth.go
```diff
@@ -35,6 +39,8 @@ func newLoginCommand(astroClient astro.Client, out io.Writer) *cobra.Command {
 	cmd.Flags().BoolVarP(&shouldDisplayLoginLink, "login-link", "l", false, "Get login link to login on a separate device for cloud CLI login")
 	cmd.Flags().BoolVarP(&shouldLoginWithToken, "token-login", "t", false, "Login with a token for browserless cloud CLI login")
 	cmd.Flags().BoolVarP(&oAuth, "oauth", "o", false, "Do not prompt for local auth for software login")
+	cmd.Flags().BoolVarP(&interactive, "interactive", "i", false, "Prompt for interactive cli")
+	cmd.Flags().IntVarP(&pageSize, "page-size", "s", defaultPageSize, "Pagination page size")
```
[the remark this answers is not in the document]
ajayy004: Done

config/context.go
```diff
@@ -44,6 +44,8 @@ type Context struct {
 	Token             string `mapstructure:"token"`
 	RefreshToken      string `mapstructure:"refreshtoken"`
 	UserEmail         string `mapstructure:"user_email"`
+	Interactive       bool   `mapstructure:"interactive"`
+	PageSize          int    `mapstructure:"pageSize"`
```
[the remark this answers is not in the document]
ajayy004: @neel-astro correct if my understanding is not correct, the user will set config using `astro config set <name> <value> -g` cmd.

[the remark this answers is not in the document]
neel-astro: yes or `astro config set <name> <value>`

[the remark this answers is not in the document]
ajayy004: Done

config/context.go
```diff
@@ -124,7 +124,7 @@ func (c *Context) SetContext() error {
 		return err
 	}
 
-	context := map[string]string{
+	context := map[string]interface{}{
```
neel-astro, line 127: is this still relevant?

config/config.go
```diff
@@ -72,6 +72,8 @@ var (
 		HoustonDialTimeout:   newCfg("houston.dial_timeout", "10"),
 		HoustonSkipVerifyTLS: newCfg("houston.skip_verify_tls", "false"),
 		SkipParse:            newCfg("skip_parse", "false"),
+		Interactive:          newCfg("interactive", "false"),
+		PageSize:             newCfg("page_size", "100"),
```
neel-astro, line 76: isn't the default page size too large for any terminal? how about 20?

cmd/software/workspace.go
```diff
@@ -99,6 +104,8 @@ func newWorkspaceSwitchCmd(out io.Writer) *cobra.Command {
 			return workspaceSwitch(cmd, out, args)
 		},
 	}
+	cmd.Flags().BoolVarP(&workspacePaginated, "paginated", "p", false, "Paginated workspace list")
+	cmd.Flags().IntVarP(&workspacePageSize, "pageSize", "s", 0, "Page size")
```
neel-astro, line 108:
```suggestion
	cmd.Flags().IntVarP(&workspacePageSize, "pageSize", "s", 0, "Page size of the workspace list if paginated is set to true")
```

cmd/software/workspace.go
```diff
@@ -177,5 +184,17 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
...
-	return workspace.Switch(id, houstonClient, out)
+	pageSize := config.CFG.PageSize.GetInt()
+
+	if config.CFG.Interactive.GetBool() || workspacePaginated {
+		if workspacePageSize < 0 && pageSize > 0 {
+			workspacePageSize = pageSize
+		}
+
+		if !(workspacePageSize > 0 && workspacePageSize <= defaultPageSize) {
+			workspacePageSize = defaultPageSize
```
neel-astro, line 195: Just slightly confused here, why do we have a `defaultPageSize` in the code and then `pageSize` as a configurable value as well? Shouldn't we fall back to configured pageSize in all cases?

neel-astro (commented):
looks fine to me, have added a few minor comments do check them out

cmd/software/workspace.go
ajayy004: I'm not sure how we can restrict at the config level that the page should not be less than 0 and not greater than 100 since houston level we have 100 as the max limit.

its if fine to let user set any number at the CLI level and if invalid houston will throw the error & will be shown in CLI

config/context.go
ajayy004: good catch

config/config.go
ajayy004: Yes, we can make that 20 default, anyhow user has the ability to set that.

cmd/software/workspace.go
neel-astro: maybe then we can add a warning message if we are using a value not specified by the user, to avoid confusion, wdyt?
Ex: `logrus.Warn("Page size cannot be more than 100, reducing the page size to 100")` or whatever you may think appropriate

neel-astro (approved):
LGTM

vishwas-astro:
@ajayy004 can you have someone other than Neel to review this PR please (possibly someone from your Apps team)?

Let's target merging this in today. Is that ok, @neel-astro ?

rishkarajgi (approved):

rishkarajgi:
@ajayy004 Would be great to have a screenshot/preview of the commands

Issue #659, opened 2022-07-25
New command Worker Queue Create


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

need to create a command that creates worker queue

**🧩 Describe the solution you'd like**

`astro deployment worker queue create`

**🤔 Describe alternatives you've considered**

**💡 Additional context**

see command details here: https://www.notion.so/astronomerio/Worker-Queue-CLI-Commands-ae1238842d4c43f49793c4c1313aae1e

PR #674, opened 2022-08-04
worker queue create

jemishp:
## Description

This PR adds a command to create a new non-default worker queue for a deployment discussed in #659. This command checks if the queue being requested exists and only allows for new worker queues to be created. The command also validates `min-count`, `max-count` and `concurrency` against astrohub api's default config and returns errors if any of the inputs are invalid. It prompts for `deployment` to use; a queue `name` and `worker-type` to use if none were specified.

Only one worker queue at a time can be created with this command.

## 🎟 Issue(s)

Related #659

### create a non default worker-queue
```
$ astro deployment worker-queue create --deployment-name jp-test-queues  --worker-type t3.2xlarge -n test-q
cl7fcde9x90222d3wrsd9l0wp
 NAME               NAMESPACE          WORKSPACE ID                   CLUSTER ID                    DEPLOYMENT ID                 DOCKER IMAGE TAG     RUNTIME VERSION
 jp-test-queues     new-phase-7131     cl0v1p6lc728255byzyfs7lw21     ckt3779dp00000rvx94xucbx7     cl7fcde9x90222d3wrsd9l0wp     5.0.8                5.0.8 (based on Airflow 2.3.4)

 Successfully updated Deployment
worker queue test-q for jp-test-queues in cl0v1p6lc728255byzyfs7lw21 workspace created
```
### prompts for name if one was not specified

```
$ astro deployment worker-queue create --deployment-name jp-test-queues  --worker-type t3.2xlarge
Enter a name for the worker queue
> test-q
...
```
### prompts for worker type if one was not specified

```
$ astro deployment worker-queue create --deployment-name jp-test-queues -n test-q
No worker-type was specified.  Select the worker type to use:
 #     WORKER TYPE      ISDEFAULT     ID
 1     t3.2xlarge       true          cl1y87zxz00046dywhzh50k9a
 2     c6i.12xlarge     false         cl639x2u6114361hx0cqw9oehw
...
```

### prompts for deployment if one was not specified

```
$ astro deployment worker-queue create -n test-q
Select a Deployment
 #     DEPLOYMENT NAME     RELEASE NAME                 DEPLOYMENT ID
 1     neel-dev-test       advanced-precession-1762     cl60i43ye406311hvr3amdaxvy
 2     jp-test-queues      new-phase-7131               cl7fcde9x90222d3wrsd9l0wp
...
```

### errors when 2 default queues are being requested

```
$astro deployment worker-queue create -n default2
Select which Deployment you want to update
 #     DEPLOYMENT NAME       RELEASE NAME                  DEPLOYMENT ID
 1     neel-dev-test         advanced-precession-1762      cl60i43ye406311hvr3amdaxvy
 2     jp-test               devoid-declination-6370       cl6omxtrn1901511jwqqjm15tkg
 3     worker-queue-test     telescopic-intensity-5645     cl6p6330i368002cripuzq1l8u

> 2
cl6omxtrn1901511jwqqjm15tkg
Error: Cannot connect to Astronomer. Try to log in with astro login or check your internet connection and user permissions.

Details: A single default worker queue must be included in the worker queue list (found 2)
```
### errors when existing queue is being modified

```
$ astro deployment worker-queue create --deployment-name jp-test-queues -n test-q --min-count 0 --max-count 30 --concurrency 25 --worker-type t3.2xlarge
Error: worker queue already exists: use worker queue update test-q instead
```

### errors when input values for min, max, concurrency are out of bounds

```
$ astro deployment worker-queue create --deployment-name jp-test-queues -n test-q --min-count 0 --max-count 30 --concurrency 85 --worker-type t3.2xlarge
Error: worker queue option is invalid: worker concurrency must be between 1 and 64
```
### errors when worker type is incorrect

```
$ astro deployment worker-queue create --deployment-name jp-test-queues -n test-q --worker-type t3.xlarge
Error: worker type selection failed: workerType t3.xlarge is not available for this deployment
```

### errors when min-count is > max-count

```
$ astro deployment worker-queue create --deployment-name jp-test-queues  --worker-type t3.2xlarge --min-count 10 --max-count 5
Enter a name for the worker queue
> test-q
cl7fcde9x90222d3wrsd9l0wp
Error: Cannot connect to Astronomer. Try to log in with astro login or check your internet connection and user permissions.

Details: Failed to create workerQueues mutation - Requested minimum worker count 10 cannot be greater than requested maximum worker count 5.
```

### errors when any api call made to get workspace, deployments, get default values, etc. fails

```
$ astro deployment worker-queue create --deployment-name jp-test-queues  --worker-type t3.2xlarge
Error: no context set, have you authenticated to Astro or Astronomer Software? Run astro login and try again
...
```

## 📋 Checklist


cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,58 @@
...
+		MaxWorkerCount:    0,
+		MinWorkerCount:    0,
+		WorkerConcurrency: 0,
+		NodePoolID:        "", // TODO query clusters to get the nodepoolID
+	}
+	// TODO is this the correct way to get worker queues?
+	if len(requestedDeployment.WorkerQueues) > 0 {
+		wQueueListToCreate = append(requestedDeployment.WorkerQueues, workerQueueToCreate) //nolint
+		// TODO if worker-queue exists, update it with requested values
+		// TODO should workerqueue.Create() do this?
```
sunkickr, line 39: This is up to you and how you structure the functions but the create command should not update an existing worker queue only the update the command. Both the create and update commands can use `workerqueue.Create()` and the function could change based on an update bool. That may be more confusing than just having two functions not sure

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	// no worker-queues exist so create a new list of one worker queue
```
sunkickr, line 42: wouldn't you also need a way to add the new worker queue to a list of already existing worker queues?

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/astronomer/astro-cli/astro-client"
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/pkg/errors"
+)
+
+var (
+	ErrWorkerQueueDefaultOptions = errors.New("failed to get worker-queue default options")
+	ErrInvalidWorkerQueueOption  = errors.New("worker-queue option is invalid")
+	ErrCannotUpdateExistingQueue = errors.New("worker-queue exists")
```
sunkickr, line 15: maybe "worker-queue already exists"

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// Get the default nodepool ID to use
+	for _, pool = range requestedDeployment.Cluster.NodePools {
+		if pool.IsDefault {
+			nodePoolIDForWorkerQueue = pool.ID
+		}
+	}
+
+	// TODO user selects nodePoolID for creating a queue on non-default nodepools
```
sunkickr, line 49: Users should be able to use a worker-type flag to select a non-default nodepool

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	return cmd
+}
+
+func newDeploymentWorkerQueueCreateCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "create",
+		Aliases: []string{"cr"},
+		Short:   "Create a Deployment's worker queue",
+		Long:    "Create a worker queue for an Astro Deployment",
+		Example: "", // TODO should we add some examples for default and custom worker-queue commands?
```
sunkickr, line 394: Would a user add a default worker queue. I thought the default worker queue already existed. If they made changes to it wouldn't that be an update?

cmd/cloud/deployment.go
```diff
@@ -51,6 +55,12 @@ var (
 		# Update a deployment variables from a file
 		$ astro deployment variable update --deployment-id <deployment-id> --load --env .env.my-deployment
 		`
+	wQueueConcurrency int
+	wQueueMin         int
```
frank-ye, line 59: We should name these more closely to the variable names in Astrohub, for consistency. At least `wQueueMinWorkerCount` or `wQueueMinWorkers`. Same comment goes for max worker.

cmd/cloud/deployment.go
frank-ye: There can only be one default worker queue, for which we can't rename (but can update other configurations). That would be an update yep

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+		Use:     "create",
+		Aliases: []string{"cr"},
+		Short:   "Create a Deployment's worker queue",
+		Long:    "Create a worker queue for an Astro Deployment",
+		Example: "", // TODO should we add some examples for default and custom worker-queue commands?
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return deploymentWorkerQueueCreate(cmd, args, out)
+		},
+	}
+	cmd.Flags().StringVarP(&deploymentID, "deployment-id", "d", "", "The deployment where the worker queue should be created.")
```
frank-ye, line 399: maybe: The deployment to create the worker queue on

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	cmd.Flags().StringVarP(&wQueueName, "name", "n", "", "The name of the worker queue. If the name contains a space, specify the entire name within quotes \"\" .")
```
frank-ye, line 400: We don't allow queue names to have spaces, they actually have to go through regex validation: 
```
export function validateQueueName(name, isDefault) {
  // Shouldn't be reached due to GraphQL schema constraint
  if (isEmpty(name)) {
    throw new Error(`No name provided for worker queue`);
  }
  const nameConstraint = /[a-z]([-a-z0-9]*[a-z0-9])?/;
  // Check non-default WQ
  if (!name.match(nameConstraint).includes(name) || name.length < 1 || name.length > 63) {
    throw new Error(
      `Non-default worker queue must contain only lowercase alphanumeric characters or '-' and start with an alphabetical character. Queue names must not exceed 63 characters.
      `,
    );
  }

  if (name === 'default' && !isDefault) {
    throw new Error(`Non-default worker queue cannot use reserved name: default.`);
  }
  // reserved name due to constraint defined in the AF operator, the operator expects us to pass down a string 'default'
  if (isDefault && name !== 'default') {
    throw new Error(`Default worker queue needs to be named: default`);
  }
}
```

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	cmd.Flags().BoolVarP(&isDefaultWQueue, "isDefault", "", false, "Make this the default worker queue by setting this to true.")
```
frank-ye, line 401: If we expose this, does this mean under the hood we set the original default queue's `isDefault` field to false? I don't think this flag is intuitive for the user. Additionally, the default queue has naming constraints.

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	cmd.Flags().IntVarP(&wQueueMin, "min-count", "", 0, "The min worker count of the worker queue. Possible values are between 1 and 10.")
```
frank-ye, line 402: The default is set to 0 but the doc string says 1 to 10. Can we set the minimum to 1? The expectation is that if a queue is defined we have at least one worker pod active in the deployment.

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	cmd.Flags().IntVarP(&wQueueMax, "max-count", "", 0, "The max worker count of the worker queue. Possible values are between 11 and 20.")
```
frank-ye, line 403: Same comment here if the max default is 0, no pods will ever spin up. The max default set in the UI / Astrohub is 10 for reference.

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
...
+	cmd.Flags().IntVarP(&wQueueConcurrency, "concurrency", "", 0, "The concurrency(number of slots) of the worker queue. Possible values are between 21 and 30.")
```
frank-ye, line 404: Default here is 16.

cmd/cloud/deployment_test.go
```diff
@@ -274,3 +274,151 @@ func TestDeploymentVariableUpdate(t *testing.T) {
...
+				},
+			},
+			WorkerQueues: []astro.WorkerQueue{},
+		},
+		{
+			ID:             "test-deployment-id-1",
+			Label:          "test-deployment-label-1",
+			RuntimeRelease: astro.RuntimeRelease{Version: "4.2.5"},
+			DeploymentSpec: astro.DeploymentSpec{
+				Workers: astro.Workers{
```
frank-ye, line 325: PS we can get rid of all the worker AU invocations, the field is deprecated

astro-client/astro_test.go
```diff
@@ -896,3 +896,59 @@ func TestGetWorkspace(t *testing.T) {
...
+			}
+		})
+		astroClient := NewAstroClient(client)
+
+		resp, err := astroClient.GetWorkerQueueOptions()
+		assert.NoError(t, err)
+		assert.Equal(t, resp, mockResponse.Data.GetWorkerQueueOptions)
+	})
+
+	t.Run("error path", func(t *testing.T) {
```
frank-ye, line 941: Are there other error paths that can occur when retrieving this info? If so, it might be worth renaming `error path` to something more descriptive.

cloud/deployment/workerqueue/workerqueue.go
frank-ye: nit: we don't use the `-` when referring to worker queue

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	workerQueueToCreate = setWorkerQueueValues(wQueueMin, wQueueMax, wQueueConcurrency, workerQueueToCreate, workerQueueDefaultOptions)
+
+	err = IsWorkerQueueInputValid(workerQueueToCreate, workerQueueDefaultOptions)
+	if err != nil {
+		return err
+	}
+
+	if QueueExists(requestedDeployment.WorkerQueues, workerQueueToCreate) {
+		// create does not allow updating existing queues
+		errHelp = fmt.Sprintf("use worker-queue update %s instead", workerQueueToCreate.Name)
```
frank-ye, line 65: same comment here on the `-`, will omit further comments in the review regarding this but might be worth a sweep changing the notation across the PR!

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+		if pool.IsDefault {
+			nodePoolIDForWorkerQueue = pool.ID
+		}
+	}
+
+	// TODO user selects nodePoolID for creating a queue on non-default nodepools
+
+	workerQueueToCreate = &astro.WorkerQueue{
+		Name:       name,
+		IsDefault:  isDefaultWQueue,
```
frank-ye, line 53: I thought we weren't allowing users to create a default worker queue from this particular command, we could set this to false here.

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+		return errors.Wrap(ErrInvalidWorkerQueueOption, errorMessage)
+	}
+	return nil
+}
+
+// QueueExists takes a []existingQueues and a queueToCreate as arguments
+// It returns true if queueToCreate exists in []existingQueues
+// It returns false if queueToCreate does not exist in []existingQueues
+
+func QueueExists(existingQueues []astro.WorkerQueue, queueToCreate *astro.WorkerQueue) bool {
```
frank-ye, line 153: Vaguely remember discussing handling the ID in the create / update paths, do we expect an ID in the create command? (Due to the  WQ struct being shared)

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+
+// IsWorkerQueueInputValid checks if the requestedWorkerQueue adheres to the floor and ceiling set in the defaultOptions
+// if it adheres to them, it returns nil
+// ErrInvalidWorkerQueueOption is returned if min, max or concurrency are out of range
+
+func IsWorkerQueueInputValid(requestedWorkerQueue *astro.WorkerQueue, defaultOptions astro.WorkerQueueDefaultOptions) error {
+	var errorMessage string
+	if !(requestedWorkerQueue.MinWorkerCount >= defaultOptions.MinWorkerCount.Floor) ||
+		!(requestedWorkerQueue.MinWorkerCount <= defaultOptions.MinWorkerCount.Ceiling) {
+		errorMessage = fmt.Sprintf("min worker count must be between %d and %d", defaultOptions.MinWorkerCount.Floor, defaultOptions.MinWorkerCount.Ceiling)
```
frank-ye, line 133: We should probably prefix these errors with - `Failed to create worker queue` so that users know which part of their input is malformatted / invalid

cmd/cloud/deployment.go
frank-ye: Note: @paolaperaza helped write some docs for the Cloud UI that are exposed to users as tool tips in the Deployment Create page. I think it's worth having uniform messaging between both clients! Also might be a typo, concurrency between `21 and 30` here?

cmd/cloud/deployment.go
frank-ye: Also `11 and 20` - are these placeholders? They don't reflect the numbers we impose in the UI

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
+package workerqueue
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/astronomer/astro-cli/astro-client"
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/pkg/errors"
```
neel-astro, line 9: let's use `errors` over `github.com/pkg/errors`, as `github.com/pkg/errors` is not being actively developed, and most if not all features of that package have been imported into the standard `errors` package over years

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+func Create(ws, deploymentID, name string, isDefaultWQueue bool, wQueueMin, wQueueMax, wQueueConcurrency int, client astro.Client, out io.Writer) error {
+	var (
+		requestedDeployment       astro.Deployment
+		err                       error
+		errHelp                   string
+		workerQueueToCreate       *astro.WorkerQueue
+		wQueueListToCreate        []astro.WorkerQueue
+		workerQueueDefaultOptions astro.WorkerQueueDefaultOptions
+		pool                      astro.NodePool
+		nodePoolIDForWorkerQueue  string
```
neel-astro, line 27:
```suggestion
		nodePoolID  string
```
nitpick: do we have to append WorkerQueue to the variable names, given that package itself is called worker queue, isn't it implicit that Create method in this package will have variables related to the worker queue, same for other variables naming as well

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(out, "worker-queue %s for %s in %s workspace created\n", workerQueueToCreate.Name, requestedDeployment.ID, ws)
+	return nil
+}
+
+// setWorkerQueueValues sets values for MinWorkerCount, MaxWorkerCount and WorkerConcurrency
+// Default values are used if the user did not request any
+
```
neel-astro, line 83:
```suggestion
```

Generally in golang, we don't leave an empty line between method and method's comment

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return astro.WorkerQueueDefaultOptions{}, err
+	}
+	return workerQueueDefaultOptions, nil
+}
+
+// IsWorkerQueueInputValid checks if the requestedWorkerQueue adheres to the floor and ceiling set in the defaultOptions
+// if it adheres to them, it returns nil
+// ErrInvalidWorkerQueueOption is returned if min, max or concurrency are out of range
+
```
neel-astro, line 128:
```suggestion
```

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+		errorMessage = fmt.Sprintf("worker concurrency must be between %d and %d", defaultOptions.WorkerConcurrency.Floor, defaultOptions.WorkerConcurrency.Ceiling)
+		return errors.Wrap(ErrInvalidWorkerQueueOption, errorMessage)
+	}
+	return nil
+}
+
+// QueueExists takes a []existingQueues and a queueToCreate as arguments
+// It returns true if queueToCreate exists in []existingQueues
+// It returns false if queueToCreate does not exist in []existingQueues
+
```
neel-astro, line 152:
```suggestion
```

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	} else {
+		workerQueueToCreate.WorkerConcurrency = workerQueueDefaultOptions.WorkerConcurrency.Default
+	}
+	return workerQueueToCreate
+}
+
+// GetWorkerQueueDefaultOptions calls the workerqueues query
+// It returns WorkerQueueDefaultOptions if the query succeeds
+// An error is returned if it fails
+
```
neel-astro, line 112:
```suggestion
```

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,165 @@
...
+	// get defaults for min-count, max-count and concurrency from API
+	workerQueueDefaultOptions, err = GetWorkerQueueDefaultOptions(client)
+	if err != nil {
+		return err
+	}
+
+	// Get the default nodepool ID to use
+	for _, pool = range requestedDeployment.Cluster.NodePools {
+		if pool.IsDefault {
+			nodePoolIDForWorkerQueue = pool.ID
```
neel-astro, line 45:
```suggestion
			nodePoolIDForWorkerQueue = pool.ID
			break
```
assuming there would be just one default

cmd/cloud/deployment.go
```diff
@@ -3,6 +3,10 @@ package cloud
 import (
 	"io"
 
+	"github.com/astronomer/astro-cli/astro-client"
+
+	"github.com/astronomer/astro-cli/cloud/deployment/workerqueue"
+
```
neel-astro, line 9:
```suggestion
	"github.com/astronomer/astro-cli/astro-client"

	"github.com/astronomer/astro-cli/cloud/deployment/workerqueue"

```

nitpick: for readability we can club imports paths, for example:
```
import (
    <stdlib>

    <third_party_lib>

    <internal_imports>
)
```

```
import (
	"io"

	"github.com/pkg/errors"
	"github.com/spf13/cobra"

	"github.com/astronomer/astro-cli/astro-client"
	"github.com/astronomer/astro-cli/cloud/deployment/workerqueue"
        airflowversions "github.com/astronomer/astro-cli/airflow_versions"
	"github.com/astronomer/astro-cli/cloud/deployment"
	"github.com/astronomer/astro-cli/pkg/httputil"
)
```

cmd/cloud/deployment.go
```diff
@@ -360,3 +371,50 @@ func deploymentVariableUpdate(cmd *cobra.Command, args []string, out io.Writer)
 
 	return deployment.VariableModify(deploymentID, variableKey, variableValue, ws, envFile, variableList, useEnvFile, makeSecret, true, astroClient, out)
 }
+
+func newDeploymentWorkerQueueRootCmd(out io.Writer) *cobra.Command {
```
neel-astro, line 375: worker queue commands can be part of `deployment_workerqueue.go` so that this file doesn't end up being very long once other worker queue CRUD commands are being added

neel-astro (commented):
Have left few minor nitpicks, do let me know once all the TODOs are done, will have another look

astro-client/astro_test.go
kushalmalani: I doubt the comment. Can we describe what is the expected error path here, so its easier to understand what is covered and if some more cases are required

cmd/cloud/deployment.go
jemishp: @frank-ye Will chance it. Thanks!

cmd/cloud/deployment.go
jemishp: @frank-ye does that mean that a user should not be able to create a new queue and make it the "default" queue? In other words we should not have this flag at all? 

@sunkickr thoughts on why we need this flag?

cmd/cloud/deployment.go
jemishp: @frank-ye @neel-astro the 0 here indicates that the user never requested a min value. However the queue that gets created will always gets defaults from astrohub set by [setWorkerQueueValues](https://github.com/astronomer/astro-cli/pull/674/files#diff-2da8cc598e86753626b6841394c1b4a35ea0e83200681100523822b2fc002979R84-R107)

cmd/cloud/deployment.go
jemishp: > Also `11 and 20` - are these placeholders? They don't reflect the numbers we impose in the UI

@frank-ye Yes these are placeholders as I was trying to not hard code astrohub's defaults in CLIs help messages. Thinking of how to do this so we do not have to change the CLI's messages each time astrohub defaults change. How often do we change these defaults in astrohub?

cloud/deployment/workerqueue/workerqueue.go
jemishp: @neel-astro good catch and thanks. Will change to use the correct package.

cloud/deployment/workerqueue/workerqueue.go
jemishp: Agreed. I will clean up the variable names to not have WorkerQueue in them.

cloud/deployment/workerqueue/workerqueue.go
jemishp: @neel-astro that is correct and I will break out of the for loop once I find the default node pool.

cmd/cloud/deployment.go
jemishp: I am surprised

cmd/cloud/deployment.go
jemishp: like this and will refactor workerqueue CRUD commands to their own file.

cloud/deployment/workerqueue/workerqueue.go
jemishp: @frank-ye The error is wrapped as below for any invalid min, max or concurrency option that a user requests:

```
min worker count must be between 1 and 20: worker queue option is invalid
```
The range above (1 - 20) is fake from a test. I can walk you through the real interaction if you'd like to see if we can improve the user facing error message. The reason I stayed away from adding a verb like `create` in the error output is so that we can re-use the same method regardless of a user trying to create or update a worker queue.

cmd/cloud/deployment_test.go
jemishp: @frank-ye we will take care of it shortly in this [issue](https://app.zenhub.com/workspaces/managed-airflow---data-practitioner-6258e7ff010e82001090f49a/issues/astronomer/astro-cli/639)

cmd/cloud/deployment.go
jemishp: @frank-ye I spoke to @sunkickr and understand the use case better so have removed the `--isDefault` flag as users can not create a default worker queue.

cloud/deployment/workerqueue/workerqueue.go
jemishp: @frank-ye yes you are correct;  users can not create a default queue so I have made this `false`. Thanks.

cmd/cloud/deployment.go
jemishp: @frank-ye Thanks for sharing the name validation logic. I have changed the user facing help message.

astro-client/astro_test.go
jemishp: @frank-ye @kushalmalani `GetWorkerQueueOptions()` is what the `astro.Client` calls to get the default worker queue options from astrohub API  via a graphql query.  The error path here is a catch all for any error returned by the public API call below:

```
	wqResp, err := wqReq.DoWithPublicClient(c)
	if err != nil {
		return WorkerQueueDefaultOptions{}, err
	}
``` 
I have made the description: `catch all for any API error` Is that better and more descriptive?

kushalmalani: I see, then i am good either way.

cmd/cloud/deployment.go
frank-ye: These defaults don't change often, there's also a query that the UI invokes to fetch these default values. It could be overkill though.

cloud/deployment/workerqueue/workerqueue.go
frank-ye: No need your justification is very clear!

cmd/cloud/deployment_workerqueue.go
```diff
@@ -0,0 +1,55 @@
...
+		Long:    "Create a worker queue for an Astro Deployment",
+		Example: "", // TODO should we add some examples for default and custom worker-queue commands?
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return deploymentWorkerQueueCreate(cmd, args, out)
+		},
+	}
+	cmd.Flags().StringVarP(&deploymentID, "deployment-id", "d", "", "The deployment where the worker queue should be created.")
+	cmd.Flags().StringVarP(&deploymentName, "deployment-name", "", "", "Name of the deployment where the worker queue should be created.")
+	cmd.Flags().StringVarP(&wQueueName, "name", "n", "", "The name of the worker queue. Queue names must not exceed 63 characters and contain only lowercase alphanumeric characters or '-' and start with an alphabetical character.")
+	cmd.Flags().IntVarP(&wQueueMinWorkerCount, "min-count", "", 0, "The min worker count of the worker queue. Possible values are between 1 and 10.")
```
sunkickr, line 38: On the UI possible values for min and max are between 0 and 30 where min can't be larger than max

cmd/cloud/deployment_workerqueue.go
```diff
@@ -0,0 +1,55 @@
...
+	cmd.Flags().IntVarP(&wQueueMaxWorkerCount, "max-count", "", 0, "The max worker count of the worker queue. Possible values are between 11 and 20.")
+	cmd.Flags().IntVarP(&wQueueConcurrency, "concurrency", "", 0, "The concurrency(number of slots) of the worker queue. Possible values are between 21 and 30.")
```
sunkickr, line 40: In the UI possible values are between 1 and 64

sunkickr (changes requested):
This PR looks good. Everything seems to be working right just requesting a few copy changes

Also not sure if this extra deployment id is coming from code in this PR but I think we should remove it
image

cmd/cloud/deployment.go
jemishp: removed the `--is-default` flag as it was not needed. Users should not be able to create a default worker queue.

cloud/deployment/workerqueue/workerqueue.go
jemishp: added a node pool selection if user does not specify a node pool.

cloud/deployment/workerqueue/workerqueue.go
jemishp: we will handle updates in #670. For now `QueueExists()` is using `ID` and `Name` to see if the queue to create exists. If it does, it fails with an `ErrCannotUpdateExistingQueue`

cmd/cloud/deployment_workerqueue.go
jemishp: @sunkickr I have removed the `"Possible values...` from the min and max help text so we do not have to hard code any values here. The CLI gets these values from astrohub and eforces them. If you are out of the allowed range an error is returned showing you what the allwoed range is. 

Let me know if that works better.

cmd/cloud/deployment_workerqueue.go
jemishp: @sunkickr I have removed the `"Possible values...` from the min and max help text so we do not have to hard code any values here. The CLI gets these values from astrohub and eforces them. If you are out of the allowed range an error is returned showing you what the allwoed range is. 

Same as min and max values above. Let me know if that works better.

jemishp:
> This PR looks good. Everything seems to be working right just requesting a few copy changes
> 
> Also not sure if this extra deployment id is coming from code in this PR but I think we should remove it image
> 
> Why is a table with information about the deployment being printed? If we are going to print a table. it would be more helpful if it was a table containing information about the worker queue just created

@sunkickr the extra deploymentID is being printed from [this line(https://github.com/astronomer/astro-cli/blob/main/cloud/deployment/deployment.go#L641) NOT part of this PR. 

The table print out is coming from `deployment.Update()` [here](https://github.com/astronomer/astro-cli/blob/main/cloud/deployment/deployment.go#L498..L500) also NOT part of this PR. 

Should we create a separate issue/s to change `deployment.Update()` and `deployment.GetUpdate()`?

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,222 @@
...
+	default:
+		// Get the nodePoolID for pool that matches workerType
+		for _, pool = range nodePools {
+			if pool.NodeInstanceType == workerType {
+				nodePoolID = pool.ID
+				return nodePoolID, errToReturn
+			}
+		}
+		// did not find a matching workerType in any node pool
+		errToReturn = fmt.Errorf("%w: workerType %s did not match any node pool", ErrInvalidNodePool, workerType)
```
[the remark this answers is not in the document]
jemishp: @sunkickr Agreed and changed it. Thanks.

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,222 @@
...
+
+			nodePoolMap[strconv.Itoa(index)] = nodePools[i]
+		}
+
+		tab.Print(out)
+		choice := input.Text("\n> ")
+		selectedPool, ok := nodePoolMap[choice]
+		if !ok {
+			// returning empty nodePoolID
+			errToReturn = fmt.Errorf("%w: invalid node pool: %s selected", ErrInvalidNodePool, choice)
```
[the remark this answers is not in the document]
jemishp: Agreed and changed it.

The Node pool selection now prints `Header:         []string{"#", "WORKER TYPE", "ISDEFAULT", "ID"},` as the header row.

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// queueToCreate does not exist so we add it
+	listToCreate = append(requestedDeployment.WorkerQueues, *queueToCreate) //nolint
+
+	// update the deployment with the new list of worker queues
+	err = deployment.Update(requestedDeployment.ID, "", ws, "", "", 0, 0, 0, listToCreate, true, client)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(out, "worker queue %s for %s in %s workspace created\n", queueToCreate.Name, requestedDeployment.ID, ws)
```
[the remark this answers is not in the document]
jemishp: @sunkickr changed it to use `deployment.label` (name).

cmd/cloud/deployment_workerqueue.go
sunkickr: I think that works better thanks

cloud/deployment/deployment.go
```diff
@@ -22,7 +22,7 @@ import (
 
 var (
 	errInvalidDeployment    = errors.New("the Deployment specified was not found in this workspace. Your account or API Key may not have access to the deployment specified")
-	errInvalidDeploymentKey = errors.New("invalid Deployment selected")
+	ErrInvalidDeploymentKey = errors.New("invalid Deployment selected")
```
neel-astro, line 25: any reason to make this public? I don't see the variable being used outside the `deployment` package

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/astronomer/astro-cli/pkg/input"
+	"github.com/astronomer/astro-cli/pkg/printutil"
+)
+
+var (
+	ErrWorkerQueueDefaultOptions = errors.New("failed to get worker queue default options")
+	ErrInvalidWorkerQueueOption  = errors.New("worker queue option is invalid")
+	ErrCannotUpdateExistingQueue = errors.New("worker queue already exists")
+	ErrInvalidNodePool           = errors.New("node pool selection failed")
```
neel-astro, line 20:
```suggestion
	errWorkerQueueDefaultOptions = errors.New("failed to get worker queue default options")
	errInvalidWorkerQueueOption  = errors.New("worker queue option is invalid")
	errCannotUpdateExistingQueue = errors.New("worker queue already exists")
	errInvalidNodePool           = errors.New("node pool selection failed")
```

nitpick: ideally we would want to keep the variable local to the package unless specifically needed outside. Just to prevent them from being misused in the future

cloud/deployment/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,223 @@
...
+// If user did not request a workerType, then it prompts the user to pick one
+// An ErrInvalidNodePool is returned if a user chooses an option not on the list
+func selectNodePool(workerType string, nodePools []astro.NodePool, out io.Writer) (string, error) {
+	var (
+		nodePoolID, message string
+		pool                astro.NodePool
+		errToReturn         error
+	)
+
+	message = "No worker-type was specified.  Select the worker type to use"
```
neel-astro, line 179:
```suggestion
	message = "No worker-type was specified. Select the worker type to use"
```

cmd/cloud/deployment.go
```diff
@@ -52,6 +54,11 @@ var (
 		# Update a deployment variables from a file
 		$ astro deployment variable update --deployment-id <deployment-id> --load --env .env.my-deployment
 		`
+	wQueueConcurrency    int
+	wQueueMinWorkerCount int
+	wQueueMaxWorkerCount int
+	wQueueWorkerType     string
+	wQueueName           string
```
neel-astro, line 61: might want to keep these under `deployment_workerqueue.go` where they are actually being used

cmd/cloud/deployment_workerqueue_test.go
```diff
@@ -0,0 +1,238 @@
...
+		testUtil.InitTestConfig(testUtil.Initial)
+		expectedOut := "Usage:\n"
+		cmdArgs := []string{"worker-queue", "create"}
+		resp, err := execDeploymentCmd(cmdArgs...)
+		assert.Error(t, err)
+		assert.NotContains(t, resp, expectedOut)
+	})
+	// TODO When updating existing queues, pass in ID of the existing queues
+	// TODO When updating existing queues, is changing the name of the existing queue/s allowed?
+	// TODO any more error cases
```
neel-astro, line 237: are these still relevant?

neel-astro (approved):
reviewed it from a CLI code perspective, looks fine to me. But have left few minor nitpick might want to have a look at that.

sunkickr:
@jemishp yeah we should probably refactor deployment update if the function is going to be part of the worker queue commands. The deployment update command needs the table but the worker queue command does not. We probably need to abstract the table print out of deployment.Update. Up to you when to do this. I think it can be done when updating worker-au. up to you

cloud/deployment/deployment.go
jemishp: `ErrInvalidDeploymentKey` is exported for testing convenience as  `workerqueue`  test uses it for an error path test.

cloud/deployment/workerqueue/workerqueue.go
jemishp: These do not need to be exported as nothing outside uses them. Good catch.

cmd/cloud/deployment.go
jemishp: post breaking out `workerqueue` from deployment, I was debating if I should leave the vars in `deployment.go` or move them closer to where they are being used. Thanks for helping me decide :) Closer to where they are used wins!

cmd/cloud/deployment_workerqueue_test.go
jemishp: Yes leaving them here as they are relevant to #670 which is coming next!

PR #680, opened 2022-08-09
Fix user role update in Software 

neel-astro:
## Description
Changes:
- Moved from using `workspaceUpdateUserRole` to `workspaceUpsertUserRole` query to update user role, this new query would allow us to cover the case when a user is added to a workspace implicitly via team added to a workspace and then someone tries to update that user role. The original query won't work because it won't find the record which needs to be updated corresponding to that user, which would be needed to be created and is covered as part of the upsert query.

## 🎟 Issue(s)

Related https://github.com/astronomer/issues/issues/4913

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/houston/types.go
+++ b/houston/types.go
@@ -44,7 +44,7 @@ type ResponseData struct {
 	CancelUpdateDeploymentRuntime  *Deployment                 `json:"cancelRuntimeUpdate,omitempty"`
 	UpdateWorkspace                *Workspace                  `json:"updateWorkspace,omitempty"`
 	DeploymentLog                  []DeploymentLog             `json:"logs,omitempty"`
-	WorkspaceUpdateUserRole        string                      `json:"workspaceUpdateUserRole,omitempty"`
+	WorkspaceUpdateUserRole        string                      `json:"workspaceUpsertUserRole,omitempty"`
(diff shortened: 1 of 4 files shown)


bote795 (approved):

houston/types.go
```diff
@@ -44,7 +44,7 @@ type ResponseData struct {
 	CancelUpdateDeploymentRuntime  *Deployment                 `json:"cancelRuntimeUpdate,omitempty"`
 	UpdateWorkspace                *Workspace                  `json:"updateWorkspace,omitempty"`
 	DeploymentLog                  []DeploymentLog             `json:"logs,omitempty"`
-	WorkspaceUpdateUserRole        string                      `json:"workspaceUpdateUserRole,omitempty"`
+	WorkspaceUpdateUserRole        string                      `json:"workspaceUpsertUserRole,omitempty"`
```
andriisoldatenko, line 47: Can you please change it here too?

andriisoldatenko, line 47:
```suggestion
	workspaceUpsertUserRole        string                      `json:"workspaceUpsertUserRole,omitempty"`
```

andriisoldatenko (changes requested):

houston/types.go
neel-astro: updated

andriisoldatenko (approved):

Issue #677, opened 2022-08-05
Fix Settings File Extra Field


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

Better YAML structure for extras:  https://www.notion.so/astronomerio/Local-to-Cloud-Airflow-Object-Sync-File-Settings-File-Enhancements-369e418f667b4fc3a734233cc15238ae

**🧩 Describe the solution you'd like**

**🤔 Describe alternatives you've considered**

**💡 Additional context**

PR #697, opened 2022-08-23
Settings file enhancement

sunkickr:
## Description

Updates settings files functionality
- Automatically updates objects that already exists to reflect changes in the settings file
- Updates the settings file conn_extra field to be more intuitive

Adds new commands
- adds `astro dev object import` command
- adds `astro dev object export` command 

## 🎟 Issue(s)
-https://github.com/astronomer/astro-cli/issues/688
- https://github.com/astronomer/astro-cli/issues/687
- https://github.com/astronomer/astro-cli/issues/677

## 📋 Checklist


settings/settings.go
```diff
@@ -227,3 +276,312 @@ func objectValidator(bound int, args ...string) bool {
...
+	err := yaml.Unmarshal([]byte(yamlCons), &connections)
+	if err != nil {
+		return err
+	}
+	// add connections to settings file
+	for i := range connections {
+		port, err := strconv.Atoi(connections[i].ConnPort)
+		if err != nil {
+			fmt.Println("Issue with parsing port number: ")
+			fmt.Println(err)
```
neel-astro, line 460:
```suggestion
			fmt.Printf("Issue with parsing port number: %s", err.Error())
```

nitpick

settings/settings.go
```diff
@@ -227,3 +276,312 @@ func objectValidator(bound int, args ...string) bool {
...
+	viperSettings.Set("airflow", settings.Airflow)
+	err = viperSettings.WriteConfig()
+	if err != nil {
+		return err
+	}
+	fmt.Printf("successfully exported Connections\n\n")
+	return nil
+}
+
+func ExportVariables(id string, logs bool) error {
```
neel-astro, line 488: can we combine the common logic of this method &  `EnvExportVariables`, because most of it is the exact same, except where the output is stored

settings/settings.go
```diff
@@ -227,3 +276,312 @@ func objectValidator(bound int, args ...string) bool {
...
+		if parseErr {
+			return errors.New("there was an error during export")
+		}
+		return nil
+	}
+
+	return errors.New("Command must be used with Airflow 2.X")
+}
+
+func ExportConnections(id string, logs bool) error {
```
neel-astro, line 437: might be that I am missing something, but can I feel most of the logic of this method and `EnvExportConnections` are quite similar, can't we move that to common method let's say `GetConnections` and then this & `EnvExportConnections` task would be to just output the connections in desired format

settings/settings.go
```diff
@@ -227,3 +276,312 @@ func objectValidator(bound int, args ...string) bool {
...
+		}
+		rmCmd := "rm tmp.var"
+		_ = execAirflowCommand(id, rmCmd)
+		fmt.Printf("successfully exported variables\n\n")
+		return nil
+	}
+	return errors.New("variable export unsuccessful")
+}
+
+func ExportPools(id string, logs bool) error {
```
neel-astro, line 526: can we combine the common logic of this method & EnvExportVariables, because most of it is the exact same, except where the output is stored

settings/settings.go
```diff
@@ -227,3 +276,312 @@ func objectValidator(bound int, args ...string) bool {
...
+	return nil
+}
+
+func jsonString(connExtra map[interface{}]interface{}) string {
+	var extra_string string
+	i := 0
+	for k, v := range connExtra {
+		key, ok := k.(string)
+		value, ok := v.(string)
+		if ok {
```
neel-astro, line 574:
```suggestion
		key, keyOk := k.(string)
		value, valueOk := v.(string)
		if keyOk && valueOk {
```

nitpick: this might be the more correct way to express it

neel-astro (changes requested):
Have left few comments do check them out, apart from that:
- Can we add more unit test cases, the coverage is down quite a lot
- Can we enable linters for some of the files in `settings` folder and fix any listing issues which might come up
- Can we remove empty files like `settings/testfiles/test.env` & `settings/testfiles/airflow_settings_export.yaml`. Most likely they were added because of unit tests, might want to delete those files at the end of each test run

.golangci.yml
```diff
@@ -2,6 +2,10 @@
 run:
   # timeout for analysis, e.g. 30s, 5m
   timeout: 3m
+  skip-files:
+    - settings/old_settings.go
+    - settings/settings.go
+    - settings/types.go
```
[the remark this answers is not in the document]
sunkickr: I have to add "_" to all the type names so the export will work such as: Conn_Type. I can not figure out a way to get this to pass the linter. If the underscore is missing go changes the field name in the yaml file to conntype instead of conn_type

airflow/docker.go
```diff
@@ -519,7 +519,75 @@ func (d *DockerCompose) Bash(container string) error {
...
+
+	imageLabels, err := d.imageHandler.ListLabels()
+	if err != nil {
+		return err
+	}
+
+	airflowDockerVersion := defaultAirflowVersion
+	airflowVersion, ok := imageLabels[airflowVersionLabelName]
+	if ok {
+		if version := semver.MustParse(airflowVersion); version != nil {
```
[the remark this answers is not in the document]
sunkickr: okay

.golangci.yml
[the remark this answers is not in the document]
neel-astro: having yaml tags defined in the struct, and using `yaml.Marshal` should resolve such issues wrt exporting data to yaml file in the desired format.

Maybe we can sit on this and figure it out

cmd/airflow.go
```diff
@@ -328,6 +339,64 @@ func newAirflowBashCmd() *cobra.Command {
...
+			return nil
+		},
+		PreRunE: utils.EnsureProjectDir,
+		RunE:    airflowSettingsImport,
+	}
+	cmd.Flags().BoolVarP(&connections, "connections", "c", false, "Import connections from an Airflow Settings File")
+	cmd.Flags().BoolVarP(&variables, "variables", "v", false, "Import variables from an Airflow Settings File")
+	cmd.Flags().BoolVarP(&pools, "pools", "p", false, "Import pools from an Airflow Settings File")
+	cmd.Flags().StringVarP(&settingsFile, "settings-file", "s", "airflow_settings.yaml", "Settings or env file export objects too")
+	cmd.Flags().BoolVarP(&logs, "logs", "l", false, "Show logs from airflow commands used to import objects")
```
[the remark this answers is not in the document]
sunkickr: instead of logs?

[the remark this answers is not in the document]
neel-astro: yup, because whatever logs we were adding as part of code logic in case the `--logs` flag is enabled seems to me quite suitable as debug logs, that way there is only one flag from the user's POV to debug things.

from the developer's POV, one can simply add a log statement as `logrus.Debugf("whatever we want to log", args...)` instead of `fmt.Printf("whatever we want to log", args...)`, that way when users sets `--verbosity=debug` flag in the CLI command that these logs will be displayed.

Also to note is that we would need to enable `--verbosity` flag for Astro users as well, reference: https://github.com/astronomer/astro-cli/blob/main/cmd/root.go#L94

settings/settings.go
sunkickr: I actually think this function is rally different from EnvExportVariables. the parsing of the output is completely different which is most of the function

.env
```diff
@@ -0,0 +1 @@
+AIRFLOW__CORE__ENABLE_XCOM_PICKLING=true
```
neel-astro, line 1: I think this wasn't intentionally added, might want to remove this file

airflow/docker.go
```diff
@@ -522,14 +522,97 @@ func (d *DockerCompose) Bash(container string) error {
...
+	if err != nil {
+		return err
+	}
+
+	fileState, err := fileutil.Exists(settingsFile, nil)
+	if err != nil {
+		return errors.Wrap(err, errSettingsPath)
+	}
+	if !fileState {
+		return errors.New("file specified does not exist")
```
neel-astro, line 550: might want to make this a constant

airflow/docker.go
```diff
@@ -522,14 +522,97 @@ func (d *DockerCompose) Bash(container string) error {
...
```
neel-astro, line 595: same here, might want to make this a constant

airflow/docker.go
```diff
@@ -542,6 +625,24 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
 	return "", err
 }
 
+func (d *DockerCompose) CheckAiflowVersion() (uint64, error) {
```
neel-astro, line 628:
```suggestion
func (d *DockerCompose) checkAiflowVersion() (uint64, error) {
```

if you can, there is a similar logic in the same file, might want to replace that by this function call

cmd/airflow.go
```diff
@@ -328,6 +338,64 @@ func newAirflowBashCmd() *cobra.Command {
...
+			return nil
+		},
+		PreRunE: utils.EnsureProjectDir,
+		RunE:    airflowSettingsImport,
+	}
+	cmd.Flags().BoolVarP(&connections, "connections", "c", false, "Import connections from an Airflow Settings File")
+	cmd.Flags().BoolVarP(&variables, "variables", "v", false, "Import variables from an Airflow Settings File")
+	cmd.Flags().BoolVarP(&pools, "pools", "p", false, "Import pools from an Airflow Settings File")
+	cmd.Flags().StringVarP(&settingsFile, "settings-file", "s", "airflow_settings.yaml", "Settings or env file export objects too")
+	cmd.Flags().BoolVarP(&logs, "verbosity", "l", false, "Show logs from airflow commands used to import objects")
```
neel-astro, line 371: There is already a global `--verbosity` flag, can't we simply enable that for all our commands irrespective of Software/Cloud. It would mean an overall more consistent UX.

neel-astro (commented):
looks ok, have left a few minor suggestions, apart from them can we make errors defined in the `settings/settings.go` as constants, instead of directly adding them in the code

airflow/docker.go
sunkickr: okay

neel-astro (approved):
LGTM ❤️

airflow/testfiles/airflow_settings.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# For more information, refer to our docs: https://www.astronomer.io/docs/cloud/stable/develop/customize-image#configure-airflowsettingsyaml
+# For issues or questions, reach out to support@astronomer.io
+
+airflow:
+  connections:
+    - conn_id: 'local_postgres'
+      conn_type: 'postgres'
+      conn_host: 'postres.host'
+      conn_schema: 'postges.shcema'
```
kushalmalani, line 12: fix typo

airflow/testfiles/airflow_settings.yaml
```diff
@@ -0,0 +1,23 @@
+# This file allows you to configure Airflow Connections, Pools, and Variables in a single place for local development only.
+# NOTE: If putting a dict in conn_extra, please wrap in single quotes.
+
+# For more information, refer to our docs: https://www.astronomer.io/docs/cloud/stable/develop/customize-image#configure-airflowsettingsyaml
+# For issues or questions, reach out to support@astronomer.io
```
kushalmalani, line 5: Can we also add a link to the repo issues so they can create tickets directly in the repo

airflow/testfiles/airflow_settings.yaml
```diff
@@ -0,0 +1,23 @@
+# This file allows you to configure Airflow Connections, Pools, and Variables in a single place for local development only.
+# NOTE: If putting a dict in conn_extra, please wrap in single quotes.
+
+# For more information, refer to our docs: https://www.astronomer.io/docs/cloud/stable/develop/customize-image#configure-airflowsettingsyaml
```
kushalmalani, line 4: Is this the right link? its point to the astro architecture

cmd/airflow.go
```diff
@@ -328,6 +337,61 @@ func newAirflowBashCmd() *cobra.Command {
...
+		newObjectImportCmd(),
+		newObjectExportCmd(),
+	)
+	return cmd
+}
+
+func newObjectImportCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import",
+		Short: "Create and update local Airflow objects from an Aiflow Settings file",
```
kushalmalani, line 357:
```suggestion
		Short: "Create and update local Airflow objects from an Airflow Settings file",
```

kushalmalani (approved):
Left some nit comments

airflow/testfiles/airflow_settings.yaml
sunkickr: yeah this link is good

PR #710, opened 2022-09-06
Fix roles in deployment users list cmd

neel-astro:
## Description
Changes:
- Fixed how the role was picked for `astro deployment user list` command, earlier we used to pick the first deployment level role associated with that user, but as per the new logic, we would be picking up the first deployment level role associated with the user for the deployment id passed in the command.

Minor Changes:
- Renamed `NoneTeamRole` to `NoneRole` to better reflect the use of the variable, as it was used in both team & user use cases.

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


houston/queries.go
```diff
@@ -339,6 +339,9 @@ mutation UpdateDeployment($deploymentId: Uuid!, $payload: JSON!, $executor: Exec
       username
       roleBindings {
         role
+		deployment {
```
andriisoldatenko, line 342:
```suggestion
	deployment {
```

houston/queries.go
```diff
@@ -339,6 +339,9 @@ mutation UpdateDeployment($deploymentId: Uuid!, $payload: JSON!, $executor: Exec
...
+			id
+		}
```
andriisoldatenko, line 344:
```suggestion
		id
	}
```

neel-astro: I think I need to look at tabs vs spaces in my machine 🤔 , thanks for catching it

andriisoldatenko: may be add validation to this file somehow?

neel-astro: because it is a multi-line string, and we have spaces & tabs inside the string makes it difficult to format/validate. Maybe there might be a better way to store graphql query strings, will have a look at it

neel-astro:
@andriisoldatenko can we have a re-review at this PR, have addressed original comments, tackling linting of graphql query string isn't a straightforward way and the plan is to tackle it in a separate PR

andriisoldatenko (approved):

Issue #715, opened 2022-09-07
Organization Switch Command


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

Some users are apart of more than one organization. Users need a way to switch to a different organization once they are logged in

**🧩 Describe the solution you'd like**

https://www.notion.so/astronomerio/CLI-Organizations-787a53478edf4916b9b2f0b3b6cfdd10

**🤔 Describe alternatives you've considered**

**💡 Additional context**

https://www.notion.so/astronomerio/CLI-Organizations-787a53478edf4916b9b2f0b3b6cfdd10

PR #719, opened 2022-09-09
Organization switch

sunkickr:
## Description

Add organization list and switch commands

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/715

## 🧪 Functional Testing

Unit tests were created

## 📸 Screenshots
Screen Shot 2022-09-09 at 12 32 20 PM

## 📋 Checklist

- [X] Ran `make test` before taking out of draft
- [X] Added/updated applicable tests
- [X] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary).

### DIFF
--- a/cloud/organization/organization.go
+++ b/cloud/organization/organization.go
@@ -43,17 +43,17 @@ func newTableOut() *printutil.Table {
 	}
 }
 
-var ListOrganizations = func(c config.Context) ([]OrgRes, error) {
+func listOrganizations(c *config.Context) ([]OrgRes, error) {
 	orgDomain := "https://api." + c.Domain + "/v1alpha1/organizations"
 	authToken := c.Token
-	ctx := http_context.Background()
+	ctx := context.Background()
 	doOptions := &httputil.DoOptions{
 		Context: ctx,
 		Headers: map[string]string{"authorization": authToken},
 	}
 	res, err := httpClient.Do("GET", orgDomain, doOptions)
 	if err != nil {
-		return []OrgRes{}, fmt.Errorf("could not retrieve token: %w", err)
+		return []OrgRes{}, fmt.Errorf("could not retrieve organization list: %w", err)
 	}
 	var orgResponse []OrgRes
 	err = json.NewDecoder(res.Body).Decode(&orgResponse)
@@ -71,7 +71,7 @@ func List(out io.Writer) error {
 		return err
 	}
 
-	or, err := ListOrganizations(c)
+	or, err := listOrganizations(&c)
 	if err != nil {
 		return errors.Wrap(err, astro.AstronomerConnectionErrMsg)
 	}
@@ -88,7 +88,7 @@ func List(out io.Writer) error {
 		} else {
 			color = false
 		}
-		tab.AddRow([]string{name, OrganizationID}, color)
+		tab.AddRow([]string{name, organizationID}, color)
 	}
 
 	tab.Print(out)
@@ -110,7 +110,7 @@ func getOrganizationSelection(out io.Writer) (string, error) {
 		return "", err
 	}
 
-	or, err := ListOrganizations(c)
+	or, err := listOrganizations(&c)
 	if err != nil {
 		return "", err
 	}
@@ -149,7 +149,7 @@ func Switch(orgName string, client astro.Client, out io.Writer) error {
 			return err
 		}
 	} else {
-		or, err := ListOrganizations(c)
+		or, err := listOrganizations(&c)
 		if err != nil {
 			return errors.Wrap(err, astro.AstronomerConnectionErrMsg)
 		}
--- a/cloud/organization/organization.go
+++ b/cloud/organization/organization.go
@@ -1,29 +1,38 @@
 package organization
 
 import (
-	"io"
-	"encoding/json"
-	"strconv"
-	// "net/url"
-	"fmt"
 	http_context "context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strconv"
 
 	"github.com/pkg/errors"
 
 	astro "github.com/astronomer/astro-cli/astro-client"
 	"github.com/astronomer/astro-cli/cloud/auth"
 	"github.com/astronomer/astro-cli/config"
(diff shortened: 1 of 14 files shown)


sunkickr:
The coverage is decreasing because I do not have a test for the function `ListOrganizations`. This function is temporary and will be replaced once we had add support for the core API to the Astro CLI. Is it OK for now that we just mock this function and not actually test it since it is temporary and will be replaced in the future?

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
+package organization
+
+import (
+	http_context "context"
```
neel-astro, line 4:
```suggestion
     "context"
```

Not sure why we need to alias the import, but in case there is a need to better name it as `httpContext` so as to follow the golang naming convention

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
...
+func newTableOut() *printutil.Table {
+	return &printutil.Table{
+		Padding:        []int{44, 50},
+		DynamicPadding: true,
+		Header:         []string{"NAME", "ID"},
+		ColorRowCode:   [2]string{"\033[1;32m", "\033[0m"},
+	}
+}
+
+var ListOrganizations = func(c config.Context) ([]OrgRes, error) {
```
neel-astro, line 46: nitpick: not sure we need to make this function a variable, since the only logic inside is to make an API call, which can be mocked easily by overriding `httpClient` itself. Ex: https://github.com/astronomer/astro-cli/blob/main/cloud/auth/auth_test.go#L101-L107

Nothing wrong with making a function variable, but we should avoid it as much as possible so that folks in the future might not end up misusing it.

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
...
+
+	or, err := ListOrganizations(c)
+	if err != nil {
+		return errors.Wrap(err, astro.AstronomerConnectionErrMsg)
+	}
+
+	tab := newTableOut()
+	for i := range or {
+		name := or[i].Name
+		OrganizationID := or[i].ID
```
neel-astro, line 82:
```suggestion
		organizationID := or[i].ID
```

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
...
+		Headers: map[string]string{"authorization": authToken},
+	}
+	res, err := httpClient.Do("GET", orgDomain, doOptions)
+	if err != nil {
+		return []OrgRes{}, fmt.Errorf("could not retrieve token: %w", err)
+	}
+	var orgResponse []OrgRes
+	err = json.NewDecoder(res.Body).Decode(&orgResponse)
+	if err != nil {
+		return []OrgRes{}, fmt.Errorf("cannot decode response: %w", err)
```
kushalmalani, line 61:
```suggestion
		return []OrgRes{}, fmt.Errorf("cannot decode organization list response: %w", err)
```

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
...
+	orgDomain := "https://api." + c.Domain + "/v1alpha1/organizations"
+	authToken := c.Token
+	ctx := http_context.Background()
+	doOptions := &httputil.DoOptions{
+		Context: ctx,
+		Headers: map[string]string{"authorization": authToken},
+	}
+	res, err := httpClient.Do("GET", orgDomain, doOptions)
+	if err != nil {
+		return []OrgRes{}, fmt.Errorf("could not retrieve token: %w", err)
```
kushalmalani, line 56: Do we need to edit the error message here?

cloud/organization/organization.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	// get auth id
+	var id string
+	if orgName == "" {
+		_id, err := getOrganizationSelection(out)
+		if err != nil {
+			return err
+		}
+
+		id = _id
```
neel-astro, line 152:
```suggestion
		id, err = getOrganizationSelection(out)
		if err != nil {
			return err
		}
```

cmd/cloud/organization.go
```diff
@@ -0,0 +1,69 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return organizationList(cmd, out)
+		},
+	}
+	return cmd
+}
+
+func newOrganizationSwitchCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "switch [workspace_id]",
```
neel-astro, line 40:
```suggestion
		Use:     "switch [organization_id]",
```

cmd/cloud/organization.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/astronomer/astro-cli/cloud/organization"
+)
+
+func newOrganizationCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "organization",
+		Aliases: []string{"or"},
```
neel-astro, line 14:
```suggestion
		Aliases: []string{"org"},
```

nitpick: maybe `org` might be a better alias, leaving it up to you to decide

neel-astro (commented):
have left a few minor comments, otherwise looks fine. 

@sunkickr I am okay to not writing unit tests if the plan is to replace the ListOrganization function, but would recommend mocking the `httpClient` instead of `ListOrganization` to write tests for other functions

cloud/organization/organization.go
sunkickr: okay I updated the tests

cloud/organization/organization.go
```diff
@@ -0,0 +1,172 @@
...
+		name := or[i].Name
+		organizationID := or[i].ID
+
+		var color bool
+
+		if c.Organization == or[i].ID {
+			color = true
+		} else {
+			color = false
+		}
```
neel-astro, line 90:
```suggestion
		if c.Organization == or[i].ID {
			color = true
		}
```
`else` clause not needed as the default value of bool is `false`

neel-astro (approved):
LGTM 👍

PR #749, opened 2022-09-22
New command for exporting organization audit logs

maxdml:
## Description

Usage:
`astro organization export-audit-logs [organization_name] --output-file [output_file.json] [flags]`

**1) Refactor to use core API**
- `astro-client/client.go` exposes two helpers:
    * `DoPublicGraphQLQuery` (called by `DoWithPublicClient` from the existing `astro-client/astro.go` methods)
    * `DoPublicRESTQuery`
- Both `DoPublicGraphQLQuery` and `DoPublicRESTQuery` call `doPublic`, which is made more generic.
- Consolidate request options in the `httputil.DoOptions` struct (`Path`, `Method`)
- Add `GetPublicRESTAPIURL()`helper

**2) new REST query**
- Add code in `cmd/cloud/` and `cloud/organization/` to export audit logs (`/organizations/:orgShortName/audit-logs`)
- Notes:
  - Need to be connected with the target organization
  - Need to be an ORGANIZATION_OWNER
  - Today our context only has an orgId, but ideally it'd also have an organization short name: core API REST routes expect the short name.

`Note: adding new core API queries:`
- In `astro-client/client.go`, add a new `c *HTTPClient` member function
- Configure `httputil.DoOptions` with the `Method` and `Path` (the path must be the relative path, _e.g._, `/organizations`
- Call `c.DoPublicRESTQuery(doOpts)`

## 🎟 Issue(s)

https://app.zenhub.com/workspaces/managed-airflow---admin-625732835c23b0001411e270/issues/astronomer/astro/3781

## 🧪 Functional Testing

Tested locally against our dev environement and the Astronomer organization

## 📸 Screenshots

Export with default filename or specifying a filename:

Screen Shot 2022-11-11 at 10 34 48 AM

Min/max params are checked by the API:
param check

Usage:

usage

## 📋 Checklist


astro-client/build-astro-clients.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+# This script is used to generate a golang client for the Astro REST public API
+# TODO: for now it gets the specs from the prod server, but eventually we will publish the specs in a public repo
```
fredzhy, line 4: eventually we will release the entire clients as a public repo/package. so no need to generate clients anymore in cli

houston/houston.go
```diff
@@ -163,8 +164,10 @@ func (c *Client) DoWithContext(doOpts httputil.DoOptions, ctx *config.Context) (
 		doOpts.Headers["authorization"] = ctx.Token
 	}
 	newLogger.Debugf("Request Data: %v\n", string(doOpts.Data))
+	doOpts.Method = http.MethodPost
+	doOpts.Path = ctx.GetSoftwareAPIURL()
 	var response httputil.HTTPResponse
-	httpResponse, err := c.HTTPClient.Do("POST", ctx.GetSoftwareAPIURL(), &doOpts)
+	httpResponse, err := c.HTTPClient.Do(&doOpts)
```
kushalmalani, line 170: @neel-astro Can you review the software changes here

cmd/cloud/organization.go
```diff
@@ -54,6 +60,24 @@ func newOrganizationSwitchCmd(out io.Writer) *cobra.Command {
...
+		Aliases: []string{"eal"},
+		Short:   "Export your organization audit logs. Requires being an organization owner.",
+		Long:    "Export your organization audit logs. Requires being an organization owner.",
+		Args:    cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return organizationExportAuditLogs(cmd, out, args)
+		},
+	}
+	cmd.PersistentFlags().StringVarP(&auditLogsOutputFilePath, "output-file", "o", "", "Path to a file for storing exported audit logs")
+	cmd.PersistentFlags().IntVarP(
```
kushalmalani, line 75: Does the endpoint to limit the logcount...if yes, can we add that ability to be able to pull only last `n` logs

maxdml: The endpoint currently only allows to select a time period in `days`.

cmd/cloud/organization.go
```diff
@@ -54,6 +62,24 @@ func newOrganizationSwitchCmd(out io.Writer) *cobra.Command {
 	return cmd
 }
 
+func newOrganizationExportAuditLogs(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "export-audit-logs [organization_name] --output-file [output_file.json]",
```
neel-astro, line 67:
```suggestion
		Use:     "export-audit-logs [organization_name] --output-file [output_file.json]",
```
General command structure of most CLI commands has been `astro <resource> <action>` or `astro <resource> <subresource> <action>` should we try and adhere to that in order to keep things consistent? cc @kushalmalani @sunkickr

Also, we have been using `organization id` for switch command, so it would be ideal from user POV to be consistent and use id over name. 

maybe final command structure could be something like: `astro organization logs export --organization-id=< organization_id>` or `astro organization audit-logs export --organization-id=< organization_id>` because generally the argument is for the resource identifier. In this case, the resource is the log and the organization is more of a parent resource from the CLI command point of view.

cmd/root.go
```diff
@@ -45,7 +45,8 @@ func NewRootCmd() *cobra.Command {
 	}
 	houstonClient = houston.NewClient(httpClient)
 
-	astroClient := astro.NewAstroClient(httputil.NewHTTPClient())
+	// FIXME I think it is correct and recommended to share our HTTP client across all wrappers.
```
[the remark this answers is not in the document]
neel-astro: Generally, it's ideal to have separate HTTP transport for each client because each client can have varying transport configurations/requirements. 

In this case, we have kept it separate because the configuration for `httpClient` is specific to the Houston client, and using the same for Astro would create confusion for the users as the configuration name would be quite misleading.

neel-astro (changes requested):
looks fine overall, have left a few comments. Also, we might need to get the unit test coverage in green

cmd/cloud/organization.go
maxdml: Let's do `astro organization audit-logs export` then. I think this might fit better the future where we can create/update/delete orgs from the CLI?

config/config.go
```diff
@@ -57,6 +57,7 @@ var (
 		LocalAstro:           newCfg("local.astrohub", "http://localhost:8871/v1"),
 		LocalCore:            newCfg("local.core", "http://localhost:8888/v1alpha1"),
 		LocalPublicAstro:     newCfg("local.public_astrohub", "http://localhost:8871/graphql"),
+		LocalPublicRESTAstro: newCfg("local.public_rest_astro", "http://localhost:8888/v1alpha1"),
```
kushalmalani, line 60: This seems same as LocalCore on line 58. Does core-api have multiple endpoints?

maxdml: Not today. I missed LocalCore. Will update the PR to use that instead of adding a new entry.

maxdml:
@neel-astro updated coverage

cmd/cloud/organization.go
neel-astro: Hmm, in that case isn't there any API on the Astrohub API side to get the organization name from the organization id? If not then might want to at least take the organization name as a flag value instead of an argument to the command, since that's the parent resource identifier to be consistent with other command structure, i.e. `astro organization audit-logs export --organization-name=[organization_name]`

cmd/cloud/organization.go
```diff
@@ -54,6 +62,37 @@ func newOrganizationSwitchCmd(out io.Writer) *cobra.Command {
...
+	}
+	cmd.AddCommand(
+		newOrganizationExportAuditLogs(out),
+	)
+	return cmd
+}
+
+func newOrganizationExportAuditLogs(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "export [organization_name] --output-file [output_file.json]",
```
neel-astro, line 80: might want to make the organization name a flag to be consistent with other CLI commands, for example existing deployment variable command looks like: `astro deployment variable create FOO=BAR FOO2=BAR2 --deployment-id <deployment-id>`

neel-astro (commented):
LGTM 👍  thanks for raising the PR, have just left a minor comment about flag vs argument might want to address that

cmd/cloud/organization.go
maxdml: Sure. I also reorganized the flag so their scope is correct. Note organization name is mandatory.

Screen Shot 2022-09-29 at 10 08 35 AM

cmd/cloud/organization.go
```diff
@@ -54,6 +63,41 @@ func newOrganizationSwitchCmd(out io.Writer) *cobra.Command {
...
+		Short:   "Manage your organization audit logs.",
+		Long:    "Manage your organization audit logs.",
+	}
+	cmd.AddCommand(
+		newOrganizationExportAuditLogs(out),
+	)
+	cmd.PersistentFlags().StringVarP(&orgName, "organization-name", "n", "", "Name of the organization to manage audit logs for.")
+	err := cmd.MarkPersistentFlagRequired("organization-name")
+	if err != nil {
+		panic(err)
```
neel-astro, line 79:
```suggestion
		log.Fatalf("Error marking organization-name flag as required in astro organization audit-logs command: %s", err.Error())
```
Better to send a meaningful error to the user instead of code panic with an entire stack trace

neel-astro (approved):
LGTM, have left a minor suggestion

maxdml:
@kushalmalani @neel-astro looks like I am not authorized to merge.

cmd/cloud/organization.go
kushalmalani: @maxdml Does the user need to pass the organization name...If they are already logged in, the cli knows what org they are logged into. We can just get that info from context and find the orgName for it and send in api call rather than asking the user to input it.

`astro organization audit logs export <path>`, if no path is specified, we just default to export it to current directory

kushalmalani:
@maxdml Let's chat once the core API issue is fixed and i can help you get this PR merged.

cmd/cloud/organization.go
maxdml: My preference would be to avoid network calls if possible. Ideally the context would hold onto both ID and org name.

All organization-related core API endpoints take the organization short name in the `path`. We would essentially be adding one more network roundtrip + processing at the API for each of these commands. For short commands this can add a noticeable delay in the CLI responsiveness. In addition, eventually this will add unnecessary load to the API.

kushalmalani: Makes sense to avoid an api call. If we can save the organization name to context, then let's just re-use that rather than asking the user to input it.

maxdml: I think adding the name to the context should be in a different PR (I'm not super familiar with that part of the code yet).

kushalmalani:
@maxdml Is this ready for a final review? If yes, can you take out of draft state

maxdml:
Finalizing some changes on the core API side. Otherwise yes it is ready.

kushalmalani (approved):
Merge it whenever you are ready.

maxdml:
Thanks @kushalmalani, waiting for https://github.com/astronomer/astro/pull/4607/ to go out

jemishp (approved):
LGTM

PR #776, opened 2022-10-03
Astro Run Command

sunkickr:
## Description

new astro run command

inlcudes dags and changes to the docker file that are for testing purposes only. These changes will be removed once the command is merged

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/834

## 🧪 Functional Testing

Added tests for new functions

## 📸 Screenshots

## 📋 Checklist


airflow/docker.go
```diff
@@ -568,6 +574,58 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
...
+
+	err = d.imageHandler.Build(airflowTypes.ImageBuildConfig{Path: d.airflowHome, Output: true, NoCache: noCache})
+	// remove airflow db init
+	fileErr := fileutil.RemoveLineFromFile("./Dockerfile", "RUN airflow db init", "")
+	if fileErr != nil {
+		fmt.Printf("Removing line 'RUN airflow db init' from Dockerfile unsuccessful: %s\n", err.Error())
+	}
+	if err != nil {
+		return err
+	}
```
neel-astro, line 620:
```suggestion
	defer func() {
		// remove airflow db init
		fileErr := fileutil.RemoveLineFromFile("./Dockerfile", "RUN airflow db init", "")
		if fileErr != nil {
			fmt.Printf("Removing line 'RUN airflow db init' from Dockerfile unsuccessful: %s\n", err.Error())
		}
	}()
	err = d.imageHandler.Build(airflowTypes.ImageBuildConfig{Path: d.airflowHome, Output: true, NoCache: noCache})
	if err != nil {
		return err
	}
```
nit: IMO a cleaner way to write the same logic

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+		"/usr/local/airflow/variables.yaml",
+	}
+	args = append(args, cmdArgs...)
+
+	fmt.Println("\nStarting a DAG run for " + dagID + "...")
+	fmt.Println("\nLoading DAGS...")
+
+	cmdErr := cmdExec(DockerCmd, stdout, stderr, args...)
+	if cmdErr != nil {
+		log.Debug(cmdErr)
```
neel-astro, line 344: shouldn't we return the error in case we are not able to run the command? or do we want to hide any/all docker-related errors in this command from the users?

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+		}
+	}
+	// docker exec
+	if containerName == "" {
+		// convert Settings file to variables and connection yaml
+		err = settings.FileToConnectionYAML(settingsFile)
+		if err != nil {
+			log.Debug(err)
+		}
+		args = []string{
```
neel-astro, line 301:
```suggestion
		}
		defer func() {
			// delete files
	                os.Remove("./variables.yaml")
	                os.Remove("./connections.yaml")
		}()
		args = []string{
```
always better to defer & delete right after the creation, instead of adding it at the end of the function

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+	// delete container
+	err = cmdExec(DockerCmd, nil, nil, "rm", "astro-run")
+	if err != nil {
+		log.Debug(err)
+	}
+
+	// delete files
+	os.Remove("./variables.yaml")
+	os.Remove("./connections.yaml")
+
```
neel-astro, line 359:
```suggestion
```
moved to defer right after `settings.FileToConnectionYAML` call

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+	cmdErr := cmdExec(DockerCmd, stdout, stderr, args...)
+	if cmdErr != nil {
+		log.Debug(cmdErr)
+	}
+
+	if err != nil {
+		log.Debug(err)
+	}
+	// delete container
+	err = cmdExec(DockerCmd, nil, nil, "rm", "astro-run")
```
neel-astro, line 351: I don't think so we need to run this in case `containerName != ""` and also can we `astro-run` a constant please

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+	if err != nil {
+		log.Debug(err)
+	}
+
+	// delete files
+	os.Remove("./variables.yaml")
+	os.Remove("./connections.yaml")
+
+	fmt.Println("\nSee the output of this command for errors. To view task logs, use the --task-logs` flag.")
```
neel-astro, line 360: quick question, is the `task-logs` flag being used anywhere? I can only see it being passed through but never used

cloud/deploy/deploy.go
```diff
@@ -473,7 +473,11 @@ func buildImageWithoutDags(path string, imageHandler airflow.ImageHandler) error
 
 		dagsIgnoreSet = true
 	}
-
+	// remove astro-run-dag from requirments.txt
```
neel-astro, line 476: I don't think it is a good idea to be doing this in `deploy.go`, always better to handle it where it is being used/created makes the code easier to understand and less complex

cloud/deploy/deploy_test.go
```diff
@@ -66,7 +66,7 @@ func TestDeployWithoutDagsDeploySuccess(t *testing.T) {
 
 	mockImageHandler := new(mocks.ImageHandler)
 	airflowImageHandler = func(image string) airflow.ImageHandler {
-		mockImageHandler.On("Build", mock.Anything).Return(nil)
+		mockImageHandler.On("Build", mock.Anything, mock.Anything).Return(nil)
```
neel-astro, line 69: why do we need to add one more argument 🤔 , I don't see any change in the `Build` function arguments in the `docker_image.go`

cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+var (
+	dagID     string
+	startDate string
+	taskLogs  bool
+)
+
+func newRunCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run DAG-ID",
+		Short: "Run a local DAG withpython by running its tasks sequentially",
```
neel-astro, line 18:
```suggestion
		Short: "Run a local DAG with python by running its tasks sequentially",
```

cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+			return nil
+		},
+		PreRunE: utils.EnsureProjectDir,
+		RunE:    run,
+	}
+	cmd.Flags().StringVarP(&envFile, "env", "e", ".env", "Location of file containing environment variables")
+	cmd.Flags().BoolVarP(&noCache, "no-cache", "", false, "Do not use cache when building container image")
+	cmd.Flags().BoolVarP(&taskLogs, "task-logs", "", false, "Show task logs while the DAG is running")
+	cmd.Flags().StringVarP(&settingsFile, "settings-file", "s", "airflow_settings.yaml", "Settings or env file export objects too")
+	cmd.Flags().StringVarP(&startDate, "start-date", "d", "", "The start date for the DAG this is the current date and time by default")
```
neel-astro, line 31: is `startDate` being used anywhere? can't see any reference of it being used

cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+	startDate string
+	taskLogs  bool
+)
+
+func newRunCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run DAG-ID",
+		Short: "Run a local DAG withpython by running its tasks sequentially",
+		Long:  "Run a local DAG by running its tasks sequentially. This command will spin up a docker airflow environment and execute your DAG code",
+		Args:  cobra.MaximumNArgs(1),
```
neel-astro, line 20:
```suggestion
		Args:  cobra.ExactArgs(1),
```
`ExactArgs` will also ensure case when someone fails to pass the dag-id as well as case when more than one argument is passed

pkg/fileutil/files.go
```diff
@@ -173,3 +173,45 @@ func GetFilesWithSpecificExtension(folderPath, ext string) []string {
...
+	f, err := os.OpenFile(filePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, os.ModePerm) //nolint:gomnd
+	if err != nil {
+		return err
+	}
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(string(content), lineText) {
+		_, err = f.WriteString("\n" + lineText + " " + commentText)
```
neel-astro, line 187:
```suggestion
		_, err = f.WriteString(lineText + " " + commentText + "\n")
```
should it end with an end of the line?

settings/settings.go
```diff
@@ -590,3 +590,70 @@ func jsonString(conn *Connection) string {
 	}
 	return extraString
 }
+
+func FileToConnectionYAML(settingsFile string) error {
```
neel-astro, line 594:
```suggestion
func WriteAirflowSettingstoYAML(settingsFile string) error {
```
nit: can we get to a more appropriate name for the function, as of now the name is somewhat misleading

settings/types.go
```diff
@@ -60,3 +60,33 @@ type AirflowPools []struct {
...
+type ConnYAML struct {
+	ConnID   string      `mapstructure:"conn_id" yaml:"conn_id"`
+	ConnType string      `mapstructure:"conn_type" yaml:"conn_type"`
+	Host     string      `mapstructure:"host" yaml:"host"`
+	Schema   string      `mapstructure:"schema" yaml:"schema"`
+	Login    string      `mapstructure:"login" yaml:"login"`
+	Password string      `mapstructure:"password" yaml:"password"`
+	Port     int         `mapstructure:"port" yaml:"port"`
+	Extra    interface{} `mapstructure:"extra" yaml:"extra"`
+}
```
neel-astro, line 92: I don't think so we need `mapstructure` tag for any of these fields

airflow/docker.go
```diff
@@ -568,6 +574,58 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
...
+
+	fmt.Println("Building image... For a faster 'astro run' experince run this command while Airflow is running with 'astro dev start'\n ")
+	// add astro-run-dag
+	err = fileutil.AddLineToFile("./requirements.txt", "astro-run-dag", "# This package is needed for the astro run command. It will be removed before a deploy")
+	if err != nil {
+		fmt.Printf("Adding 'astro-run-dag' package to requirements.txt unsuccessful: %s\nManually add package to requirements.txt", err.Error())
+	}
+	// add airflow db init
+	err = fileutil.AddLineToFile("./Dockerfile", "RUN airflow db init", "")
```
jemishp, line 607: Is there a way to avoid adding/removing lines to the dockerfile all together? Instead is it possible to run the command directly after logging into the container or via an env var?

airflow/docker.go
```diff
@@ -568,6 +574,58 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
...
+	if err != nil {
+		fmt.Printf("Adding line 'RUN airflow db init' to Dockerfile unsuccessful: %s\nYou may need to manually add this line for 'astro run' to work", err.Error())
+	}
+
+	err = d.imageHandler.Build(airflowTypes.ImageBuildConfig{Path: d.airflowHome, Output: true, NoCache: noCache})
+	// remove airflow db init
+	fileErr := fileutil.RemoveLineFromFile("./Dockerfile", "RUN airflow db init", "")
```
jemishp, line 614: should we check if the build succeeded before removing the line?

airflow/docker.go
jemishp: We could also refactor all of the file edits, image build and run the new image into a smaller helper methods to keep the flow small and easy to read.

airflow/docker_image.go
jemishp: +1

airflow/docker_image.go
```diff
@@ -276,6 +275,92 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+
+	fmt.Println("\nStarting a DAG run for " + dagID + "...")
+	fmt.Println("\nLoading DAGS...")
+
+	cmdErr := cmdExec(DockerCmd, stdout, stderr, args...)
+	if cmdErr != nil {
+		log.Debug(cmdErr)
+	}
+
+	if err != nil {
```
jemishp, line 347: why are we checking this err and logging it? Where is it coming from?

cloud/deploy/deploy.go
jemishp: +1

cloud/deploy/deploy.go
```diff
@@ -534,6 +538,11 @@ func buildImage(c *config.Context, path, currentVersion, deployImage, imageName
 				return "", err
 			}
 		} else {
+			// remove astro-run-dag from requirments.txt
```
jemishp, line 541: since this is the case for when `!dagDeployEnabled` why not call `buildImageWithoutDags()` as it is removing the line and building image too. Or else re-use the same code from a separate helper method.

jemishp (changes requested):

airflow/docker.go
```diff
@@ -568,6 +574,58 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
 	return "", err
 }
 
+func (d *DockerCompose) RunTest(dagID, settingsFile, startDate string, noCache, taskLogs bool) error {
+	// Get project containers
+	psInfo, err := d.composeService.Ps(context.Background(), d.projectName, api.PsOptions{
+		All: true,
+	})
+	if err != nil {
+		return errors.Wrap(err, composeCreateErrMsg)
```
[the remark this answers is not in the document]
sunkickr: I agree

airflow/docker_image.go
```diff
@@ -22,6 +21,9 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	airflowTypes "github.com/astronomer/astro-cli/airflow/types"
+	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/settings"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
sunkickr: okay

airflow/docker.go
```diff
@@ -568,6 +574,58 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
...
+					if err != nil {
+						return err
+					}
+					return nil
+				}
+			}
+		}
+	}
+
+	fmt.Println("Building image... For a faster 'astro run' experince run this command while Airflow is running with 'astro dev start'\n ")
```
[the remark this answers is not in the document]
sunkickr: got it

airflow/docker.go
sunkickr: with this method it gets cached so there is some upside. Id like a way to run it without adding to the Dockerfile. Right now I am just running `docker run <python command>` because this gets the fastest results. If we can find another way to do it that doesn't slow the command down without adding a line that would be great. I think this is an improvement we should plan to add on later

airflow/docker.go
sunkickr: My thought was to remove the line no matter what so the line doesn't confuse the user. If the build fails the line would still be in the Dockerfile. It always be removed

airflow/docker.go
sunkickr: okay

sunkickr: @jemishp which file edit functions should we refactor?

airflow/docker_image.go
sunkickr: okay

airflow/docker_image.go
sunkickr: This error "cmdErr" is returned at the end of the command I just want to make sure that the container is deleted if the command fails

airflow/docker_image.go
sunkickr: sorry this is a typo I'll remove

airflow/docker_image.go
sunkickr: okay

cloud/deploy/deploy_test.go
sunkickr: I changed build but went back. weird that these tests aren't failing

sunkickr: I'll fix it

cmd/run.go
sunkickr: startDate feature doesn't exist yet

sunkickr: I'll remove since it seems it will be awhile till we have it

pkg/fileutil/files.go
sunkickr: not sure but it definitely needs a `"\n" +` up front if the users file doesn't end their file with a new line this won't work

sunkickr: in this particular use case didn't think we needed to add a new line at the end since we are going to delete it later

settings/settings.go
sunkickr: okay I'll change

settings/types.go
sunkickr: not sure

cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+		PreRunE: utils.EnsureProjectDir,
+		RunE:    run,
+	}
+	cmd.Flags().StringVarP(&envFile, "env", "e", ".env", "Location of file containing environment variables")
+	cmd.Flags().BoolVarP(&noCache, "no-cache", "", false, "Do not use cache when building container image")
+	cmd.Flags().BoolVarP(&taskLogs, "task-logs", "", false, "Show task logs while the DAG is running")
+	cmd.Flags().StringVarP(&settingsFile, "settings-file", "s", "airflow_settings.yaml", "Settings or env file export objects too")
```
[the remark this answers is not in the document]
sunkickr: yes

airflow/docker_image.go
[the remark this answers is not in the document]
sunkickr: fixed

cloud/deploy/deploy.go
sunkickr: okay I fixed it

airflow/docker.go
```diff
@@ -171,6 +171,12 @@ func (d *DockerCompose) Start(imageName, settingsFile string, noCache, noBrowser
 
 	// Build this project image
 	if imageName == "" {
+		// add astro-run-dag package
+		// add astro-run-dag
+		err = fileutil.AddLineToFile("./requirements.txt", "astro-run-dag", "# This package is needed for the astro run command. It will be removed before a deploy")
```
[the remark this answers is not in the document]
sunkickr: fixed

airflow/docker.go
```diff
@@ -171,10 +171,21 @@ func (d *DockerCompose) Start(imageName, settingsFile string, noCache, noBrowser
...
+		}
 		err = d.imageHandler.Build(airflowTypes.ImageBuildConfig{Path: d.airflowHome, Output: true, NoCache: noCache})
 		if err != nil {
 			return err
 		}
+		// remove astro-run-dag from requirments.txt
+		err = fileutil.RemoveLineFromFile("./requirements.txt", "astro-run-dag", " # This package is needed for the astro run command. It will be removed before a deploy")
+		if err != nil {
+			fmt.Printf("Removing line 'astro-run-dag' package from requirements.txt unsuccessful: %s\n", err.Error())
+		}
```
neel-astro, line 187:
```suggestion
		if err != nil {
			fmt.Printf("Adding 'astro-run-dag' package to requirements.txt unsuccessful: %s\nManually add package to requirements.txt", err.Error())
		}
		defer func() {
		        // remove astro-run-dag from requirments.txt
		        err = fileutil.RemoveLineFromFile("./requirements.txt", "astro-run-dag", " # This package is needed for the astro run command. It will be removed before a deploy")
		        if err != nil {
			        fmt.Printf("Removing line 'astro-run-dag' package from requirements.txt unsuccessful: %s\n", err.Error())
		        }
		}
		err = d.imageHandler.Build(airflowTypes.ImageBuildConfig{Path: d.airflowHome, Output: true, NoCache: noCache})
		if err != nil {
			return err
		}
```
to cover all the edge cases always better to have cleanup things as part of defer right after creation

airflow/docker.go
```diff
@@ -568,6 +579,59 @@ func (d *DockerCompose) getWebServerContainerID() (string, error) {
...
+					}
+					return nil
+				}
+			}
+		}
+	}
+
+	fmt.Println("Building image... For a faster 'astro run' experience run this command while Airflow is running with 'astro dev start'\n ")
+	// add astro-run-dag
+	err = fileutil.AddLineToFile("./requirements.txt", "astro-run-dag", "# This package is needed for the astro run command. It will be removed before a deploy")
```
neel-astro, line 606: will `astro-run-dag` package be installed in the scheduler container if the local project is already running?

airflow/docker_image.go
```diff
@@ -276,6 +276,91 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
 	return nil
 }
 
+func (d *DockerImage) Run(dagID, envFile, settingsFile, containerName string, taskLogs bool) error {
+	// delete container
```
neel-astro, line 280:
```suggestion
```
nit: is it still relevant, or else we can drop it?

airflow/docker_image.go
```diff
@@ -276,6 +276,91 @@ func (d *DockerImage) TagLocalImage(localImage string) error {
...
+	}
+	if containerName == "" {
+		// delete container
+		err = cmdExec(DockerCmd, nil, nil, "rm", astroRunContainer)
+		if err != nil {
+			log.Debug(err)
+		}
+	}
+
+	fmt.Println("\nSee the output of this command for errors. To view task logs, use the --task-logs` flag.")
```
neel-astro, line 360:
```suggestion
	fmt.Println("\nSee the output of this command for errors. To view task logs, use the '--task-logs' flag.")
```
nit

settings/types.go
```diff
@@ -60,3 +60,33 @@ type AirflowPools []struct {
 	PoolSlot        string `yaml:"slots"`
 	PoolDescription string `yaml:"description"`
 }
+
+// types for creating variables and connections yaml files
+
+type VariablesYAML struct {
```
neel-astro, line 66:
```suggestion
type DAGRunVariables struct {
```
I am just trying to figure out a better name for this since there are already two similar structs in the same file, it would be tough for a newcomer just to read the code and understand the use case of all of these structs if you have a better name then do go for it.

settings/types.go
```diff
@@ -60,3 +60,33 @@ type AirflowPools []struct {
...
+}
+
+type VarYAMLs []VarYAML
+
+type VarYAML struct {
+	Key   string `mapstructure:"key" yaml:"key"`
+	Value string `mapstructure:"value" yaml:"value"`
+}
+
+type ConnectionsYAML struct {
```
neel-astro, line 77:
```suggestion
type DAGRunConnections struct {
```
same as last comment

neel-astro (commented):
overall LGTM ❤️, this would add a lot of value for our users so thanks for all the effort. Have left a few minor nitpicks, will leave my approval once those are addressed and the `--task-logs` logic has been added.

airflow/docker.go
sunkickr: no this would be done when someone runs astro dev start so doesn't need to happen here

airflow/docker_image.go
sunkickr: yes we need to delete the container here just in case on exists and wasn't deleted before for some reason

settings/types.go
sunkickr: sure

airflow/docker_image.go
neel-astro: yeah, but the comment over here does not fit any purpose to have it at the start of `Run` function, makes sense to have it where we are deleting the container but that is already there

neel-astro (approved):
LGTM 👍 , but we would still need to fix the breaking unit tests before merging

sunkickr:
@neel-astro I moved the comment

neel-astro (approved):

jemishp (approved):
looks good but need to bump coverage.

Issue #787, opened 2022-10-05
Test enable dag deploy on `astro deploy` to improve UX


kushalmalani:
Test enable dag deploy on `astro deploy` to improve UX

PR #788, opened 2022-10-05
Adding dag deploy toggle to astro deploy

kushalmalani:
## Description

Adding dag deploy toggle to astro deploy

## 🎟 Issue(s)

Related #787

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/include/astronomermonitoringdag.go
+++ b/airflow/include/astronomermonitoringdag.go
@@ -1,0 +1,33 @@
+package include
+
+import "strings"
+
+var MonitoringDag = strings.TrimSpace(`
+"""A Monitoring DAG used by Astronomer to alert users via Control Plane when tasks aren't executing"""
+
+import os
+from datetime import timedelta
+
+from airflow import DAG
+from airflow.operators.bash import BashOperator
+from airflow.utils.dates import days_ago
+
+with DAG(
+    dag_id="astronomer_monitoring_dag",
+    schedule_interval=os.environ.get("AIRFLOW_MONITORING_DAG_SCHEDULE_INTERVAL", "*/5 * * * *"),
+    start_date=days_ago(2),
+    catchup=False,
+    is_paused_upon_creation=os.environ.get("PAUSE_ASTRONOMER_MONITORING_DAG", "False").lower()=="true",
+    dagrun_timeout=timedelta(minutes=30),
+    description=__doc__,
+    doc_md=__doc__,
+    tags=["monitoring"],
+) as dag:
+    hello = BashOperator(
+        task_id="hello",
+        bash_command="echo Hello from Astronomer!",
+        depends_on_past=False,
+        priority_weight=2 ** 31 - 1,
+        do_xcom_push=False,
+    )
+`)
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -128,7 +128,7 @@ func deployDags(path, runtimeID string, client astro.Client) error {
 }
 
 // Deploy pushes a new docker image
-func Deploy(path, runtimeID, wsID, pytest, envFile, imageName, deploymentName string, prompt, dags bool, client astro.Client) error { //nolint: gocognit, gocyclo
+func Deploy(path, runtimeID, wsID, pytest, envFile, imageName, deploymentName, dagDeploy string, prompt, dags bool, client astro.Client) error { //nolint: gocognit, gocyclo
 	// Get cloud domain
 	c, err := config.GetCurrentContext()
 	if err != nil {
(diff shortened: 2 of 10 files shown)


airflow/include/astronomermonitoringdag.go
```diff
@@ -0,0 +1,33 @@
+package include
+
+import "strings"
+
+var MonitoringDag = strings.TrimSpace(`
```
kushalmalani, line 5: We add this monitoring DAG on runtime when dag deploy is enabled to make sure its consistent across all deployments.

cloud/deploy/deploy.go
```diff
@@ -117,6 +125,7 @@ func deployDags(path, runtimeID string, client astro.Client) error {
 	// Delete the tar file
 	defer func() {
 		dagFile.Close()
+		os.Remove(monitoringDagPath)
```
neel-astro, line 128: nitpick: not directly related to this, but shouldn't the defer call be defined right after the operation we want to revert at the end of function execution, else in case the function fails because of an error in `ReportDagDeploymentStatus` or `azureUploader` the tar file as well as the monitoring dag both won't be deleted i.e.

```Golang
        dagDeployment, err := deployment.Initiate(runtimeID, client)
	if err != nil {
		return err
	}

	// Check the dags directory
	dagsPath := path + "/dags"
	monitoringDagPath := filepath.Join(dagsPath, "astronomer_monitoring_dag.py")

	// Create monitoring dag file
	err = fileutil.WriteStringToFile(monitoringDagPath, include.MonitoringDag)
	if err != nil {
		return err
	}

	// Generate the dags tar
	err = fileutil.Tar(dagsPath, path)
	if err != nil {
		return err
	}

         // Delete the tar file
	defer func() {
		os.Remove(monitoringDagPath)
		err = os.Remove(dagFile.Name())
		if err != nil {
			fmt.Println("\nFailed to delete dags tar file: ", err.Error())
			fmt.Println("\nPlease delete the dags tar file manually from path: " + dagFile.Name())
		}
	}()
			
	dagsFilePath := path + "/dags.tar"
	dagFile, err := os.Open(dagsFilePath)
	if err != nil {
		return err
	}
	defer dagFile.Close()
	versionID, err := azureUploader(dagDeployment.DagURL, dagFile)
	if err != nil {
		return err
	}
	var status string
	if versionID != "" {
		status = "SUCCEEDED"
	} else {
		status = "FAILED"
	}
	_, err = deployment.ReportDagDeploymentStatus(dagDeployment.ID, runtimeID, action, versionID, status, message, client)
	if err != nil {
		return err
	}
```

cloud/deploy/deploy.go
```diff
@@ -167,6 +176,52 @@ func Deploy(path, runtimeID, wsID, pytest, envFile, imageName, deploymentName st
...
+		if err != nil {
+			return err
+		}
+
+		runtimeID = currentDeployment.ID
+		scheduler := astro.Scheduler{}
+		scheduler.AU = currentDeployment.DeploymentSpec.Scheduler.AU
+		scheduler.Replicas = currentDeployment.DeploymentSpec.Scheduler.Replicas
+		spec := astro.DeploymentCreateSpec{
+			Executor:  "CeleryExecutor",
```
neel-astro, line 190: nitpick: might want to make this a constant, and is this okay to hardcode, I don't have much idea about how Astro works but can deployment have any other executor type?

cmd/cloud/deploy.go
```diff
@@ -108,5 +109,5 @@ func deploy(cmd *cobra.Command, args []string) error {
 	// Silence Usage as we have now validated command input
 	cmd.SilenceUsage = true
 
-	return deployImage(config.WorkingPath, deploymentID, ws, pytestFile, envFile, imageName, deploymentName, forcePrompt, dags, astroClient)
+	return deployImage(config.WorkingPath, deploymentID, ws, pytestFile, envFile, imageName, deploymentName, dagDeploy, forcePrompt, dags, astroClient)
```
neel-astro, line 112: nitpick: should we validate the value of `dagDeploy` before passing it to the deployImage, just curious on what would happen in case someone passes an invalid value.

neel-astro (commented):
looks fine, have left few minor comments

cloud/deploy/deploy.go
kushalmalani: Not for now. By default all deployments use Celery Executors

cloud/deploy/deploy.go
kushalmalani: Will need to do it post uploading the file.

cmd/cloud/deploy.go
kushalmalani: Right now, it would maintain the current state and not deploy it. But i think it makes sense to validate the incoming value and throw an error if the value is not one among the possible values.

sunkickr (approved):
LGTM

Issue #615, opened 2022-06-21
Using `astro deploy` with colima is not working


sunkickr:
**Describe the bug**

The issue seems to be that docker login is not working anymore. for this to be working with colima it must be done with 

```
All Pytests passed!
Pushing image to Astronomer registry
The push refers to repository [images.astronomer-dev.cloud/ckmzjm22937931d35puw3tcdo/cl4ogkybf14562s1q96omhr04]
9e933a9c0799: Preparing 
41cef26ad0d2: Preparing 
bace4fa01f7e: Preparing 
f0361c874bc0: Preparing 
8e8b528c1008: Preparing 
5f70bf18a086: Waiting 
7aa1460a63a9: Waiting 
6ddbfdfa2663: Waiting 
2d4836d02795: Waiting 
670df0ae3eaf: Waiting 
c9448c284e24: Waiting 
5928455b8faa: Waiting 
5d00a2aaa537: Waiting 
8bf0625426fd: Waiting 
7033ce60c163: Waiting 
1f372117d631: Waiting 
13c79935a81d: Waiting 
7fbe7620d085: Waiting 
d1910a49fce7: Waiting 
db6d3f8a8114: Waiting 
68472b43fff8: Waiting 
a5e1b13d17e7: Waiting 
2d795c040b71: Waiting 
0c12ab093adb: Waiting 
1008b6c73ce5: Waiting 
b02d7c827250: Waiting 
3670db09d21e: Waiting 
b52b205045a6: Waiting 
284a6c64b82c: Waiting 
388eedeb736e: Waiting 
2feece0964b8: Waiting 
e8b689711f21: Waiting 
unauthorized: authentication required

```

**What CLI Version did you experience this bug?**

v1.1.0

**What Operating System is the above CLI installed on?**

**🪜 Steps To Reproduce**

**📸 Screenshots**
<img width="809" alt="Screen Shot 2022-06-21 at 11 01 34 AM" src="https://user-images.githubusercontent.com/63181127/174868262-f506c036-7d1b-4269-9e33-f9fd5be7ea21.png">

sunkickr:
Haven't been able to figure out what is causing this solution on colima. I am able to confirm that if you login to the astro registry with `docker login images.astronomer.cloud -u cli -p <auth-token>` you would get the auth token from cloud.astronomer.io/token

yannrouillard:
[...]

PR #814, opened 2022-10-13
fix colima deploy part2

sunkickr:
## Description

seems like the login command is still not working. This update get it to work and prints the result of running the command ( in this case "login succeeded")

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/615

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -300,7 +300,7 @@ func useBash(authConfig *cliTypes.AuthConfig, image string) error {
 	var err error
 	if authConfig.Username != "" { // Case for cloud image push where we have both registry user & pass, for software login happens during `astro login` itself
 		cmd := "echo \"" + authConfig.Password + "\"" + " | docker login " + authConfig.ServerAddress + " -u " + authConfig.Username + " --password-stdin"
-		err = cmdExec("bash", os.Stdout, os.Stderr, "-c", cmd) // This command will only work on machines that have bash
+		err = cmdExec("bash", os.Stdout, os.Stderr, "-c", cmd) // This command will only work on machines that have bash. If users have issues we will revist
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


airflow/docker_image.go
```diff
@@ -299,7 +299,8 @@ var cmdExec = func(cmd string, stdout, stderr io.Writer, args ...string) error {
 func useBash(authConfig *cliTypes.AuthConfig, image string) error {
 	var err error
 	if authConfig.Username != "" { // Case for cloud image push where we have both registry user & pass, for software login happens during `astro login` itself
-		err = cmdExec(EchoCmd, nil, os.Stderr, fmt.Sprintf("%q", authConfig.Password), "|", DockerCmd, "login", authConfig.ServerAddress, "-u", authConfig.Username, "--password-stdin")
+		cmd := "echo \"" + authConfig.Password + "\"" + " | docker login " + authConfig.ServerAddress + " -u " + authConfig.Username + " --password-stdin"
+		err = cmdExec("bash", os.Stdout, os.Stderr, "-c", cmd)
```
neel-astro, line 303: will this work on a machine that does not have bash, let's say windows powershell?
what is the root cause for the original command not working with Colima?

sunkickr: I don't think that colima works with windows powershell

sunkickr: with the old command I was getting `Error: Cannot perform an interactive login from a non TTY device`

sunkickr: for some reason running the echo command with bash fixes it: https://stackoverflow.com/questions/52280905/golang-exec-command-bash-command-not-working

neel-astro (approved):

airflow/docker_image.go
neel-astro: let's leave a comment in the code on the drawback of the new logic

Issue #790, opened 2022-10-06
Propogate errors to the users that prevent webserver or scheduler to start


sunkickr:
**Describe the bug**

If the  webserver or scheduler get stuck in a starting state the user doesn't know

```
Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
Env file ".env" found. Loading...
[+] Running 4/4
 ⠿ Container example-repo-update_e4e3a9-postgres-1   Started                                                                                                                                           1.2s
 ⠿ Container example-repo-update_e4e3a9-scheduler-1  Started                                                                                                                                           2.4s
 ⠿ Container example-repo-update_e4e3a9-triggerer-1  Started                                                                                                                                           2.4s
 ⠿ Container example-repo-update_e4e3a9-webserver-1  Started                                                                                                                                           3.2s


Airflow is starting up! This might take a few minutes…
```

**What CLI Version did you experience this bug?**

1.6

**What Operating System is the above CLI installed on?**

**🪜 Steps To Reproduce**

1. use astro-sdk in dag
2. set AIRFLOW__CORE__ENABLE_XCOM_PICKLING=FALSE
3. astro dev start

**📸 Screenshots**

sunkickr:
A few ideas here:
- 5 min timeout on the webserver health status check
- have a max number of pod retries for the scheduler and webserver (maybe 3 retries)

PR #815, opened 2022-10-13
return error if webserver is unhealthy

jemishp:
## Description

This PR implements a timeout of 1 minute after which an error gets returned to the user when webserver fails to start. Previously `astro dev start` would wait indefinitely and if there are issues we would never return. 

Additionally we check if CLI is running on M1 architecture and use a longer wait ( 5 minutes ) before declaring a startup failure. 

Users can also se `--wait 2m` to over ride default timeout all together and use their custom timeout. 

It also has an additional print `Project is not running. Run 'astro dev logs --webserver | --scheduler' for details.`

## 🎟 Issue(s)

Related #790 

## 🧪 Functional Testing

```bash
$ astro dev start
...
Env file ".env" found. Loading...
[+] Running 4/4
 ⠿ Container astro-local_005fa4-postgres-1   Started                       0.5s
 ⠿ Container astro-local_005fa4-triggerer-1  Started                       1.2s
 ⠿ Container astro-local_005fa4-scheduler-1  Started                       1.2s
 ⠿ Container astro-local_005fa4-webserver-1  Started                       1.8s


Airflow is starting up! This might take a few minutes…

Project is not running. Run 'astro dev logs --webserver | --scheduler' for details.
Error: webserver failed to start: timed out after 5m0s
```
```bash
$ astro dev start --wait 2m // this will wait 2 minutes regardless of architecture
```

## 📋 Checklist


airflow/docker.go
```diff
@@ -656,7 +657,11 @@ var checkWebserverHealth = func(settingsFile string, project *types.Project, com
 					}
 				}
 				return errComposeProjectRunning
+			} else if string(marshal) == `{"action":"health_status: unhealthy"}` {
```
neel-astro, line 660: currently on M1 mac, the webserver will always go to an unhealthy state and then become healthy, this is because it takes about 4-5 minutes of start-up time, and our current health check configuration isn't enough to wait for such long. So with these changes on M1 mac, it will always go to this else clause even if the webserver most likely would come up after a few minutes.

So we might want to increase the health check retries over here in this PR: https://github.com/astronomer/astro-cli/blob/main/airflow/include/composeyml.go#L99-L104

neel-astro (changes requested):

airflow/docker.go
sunkickr: what should we increase the health check retries too?

jemishp: I would say 4-5 minutes is a long time to wait especially if you are watching the screen. It currently waits 30 seconds, so how does bumping the retries to 45 sound? That gives us 1 and a half minutes.

neel-astro: But the problem with having 1:30 minutes is that on M1 mac it is never going to be enough at this point in time, thus always sending back the error response that the project is not running, which means we would be conveying incorrect information back to the user as the project is running and trying to start up and cause confusion.

Hence in trying to solve one issue, we might end up causing another issue from the user's POV

neel-astro: Or if we can be somewhat smarter, then we can have separate health check logic for separate machine types, i.e. if a user is on a darwin arm64 machine then we have a health check of 5 minutes, else we have a health check of 60 seconds. wdyt? cc @sunkickr @jemishp

kushalmalani (approved):

sunkickr:
This does not seem to be working the scenario described in this issue: https://github.com/astronomer/astro-cli/issues/790

jemishp:
@sunkickr @neel-astro You are both correct and I have refactored it to use the context with a timer. We wait for the webserver to get to a healthy state for 1 minute by default and 5 mins on a M1. If it times out, we return the user an error and point them to the logs. Please try it out and let me know if you see other issues.

pkg/util/util.go
```diff
@@ -76,3 +76,13 @@ func CheckEnvBool(envBool string) bool {
 	}
 	return false
 }
+
+// IsM1 returns true if running on M1 architecture
+// returns false if not running on M1 architecture
+// We use this to setup longerHealthCheck
+func IsM1(myOS, myArch string) bool {
```
andriisoldatenko, line 83: based on this article https://www.yellowduck.be/posts/detecting-apple-silicon-via-go 
proper way to check it by calling:
```
sysctl -in sysctl.proc_translated
```
I just tested on my macs and it returns "" for Intel and != "" (`0`) for Intel chips.

Since we are using go.1.18 your approach should work I susspect:

> Go 1.17 [implemented](https://tip.golang.org/doc/go1.17#compiler) a new way of passing function arguments and 
results using registers instead of the stack on 64-bit x86 architecture on selected operating systems. Go 1.18 expands the supported platforms to include 64-bit ARM (GOARCH=arm64), big- and little-endian 64-bit PowerPC (GOARCH=ppc64, ppc64le), as well as 64-bit x86 architecture (GOARCH=amd64) on all operating systems. On 64-bit ARM and 64-bit PowerPC systems, benchmarking shows typical performance improvements of 10% or more.

https://tip.golang.org/doc/go1.18

jemishp: @andriisoldatenko it works provided that the user has installed the correct golang for their architecture.

sunkickr (approved):
Everything works as expected. Wondering if we need a time-out flag here so users can increase the time-out if needed

neel-astro (approved):

Issue #802, opened 2022-10-12
Check if ports are available before dev start


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

Sometime a user my start airflow but not have their ports available. In this case they may wait a long time before knowing that their start won't work due to unavailable ports 

**🧩 Describe the solution you'd like**

Check if the ports needed are available before running the command

**🤔 Describe alternatives you've considered**

**💡 Additional context**

frankcash:
@sunkickr  should we also check for ports defined in docker-compose override or just the ports for the webserver and db?

PR #818, opened 2022-10-14
Check Ports Before Start

sunkickr:
## Description

Check if ports are already in use before running dev start. this will save users time so they don't have to go through a whole start befoer knowing if it's going to break due to already allocated ports 

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

- [X] Ran `make test` before taking out of draft


cmd/airflow.go
```diff
@@ -753,3 +763,13 @@ func prepareDefaultAirflowImageTag(airflowVersion string, httpClient *airflowver
 	}
 	return defaultImageTag
 }
+
+func checkPort(port string) error {
```
jemishp, line 767: Does this have to be a string or can it be an int? Also what if the string was `""` or if int was 0?

cmd/airflow.go
```diff
@@ -753,3 +763,13 @@ func prepareDefaultAirflowImageTag(airflowVersion string, httpClient *airflowver
...
+	host := "localhost:" + port
```
jemishp, line 768: Why are we hard coding localhost here? `checkPort()` is being used to check webserver port and postgres port. There is a `config.CFG.PostgresHost` which could have a value that is different than `localhost`.

jemishp (changes requested):
Please add missing tests.

Also left some comments/questions.

cmd/airflow.go
sunkickr: net.Listen() takes a string so this would have to be a string

cmd/airflow.go
sunkickr: I tried using config.CFG.PostgresHost instead of localhost and that did not work. Only applying "localhost" directly works. config.CFG.PostgresHost is set to postgres by default not sure what that is

cmd/airflow.go
```diff
@@ -500,7 +502,16 @@ func airflowStart(cmd *cobra.Command, args []string) error {
 	if len(args) > 0 {
 		envFile = args[0]
 	}
-
+	// check if ports are allocated
+	err := checkPort(config.CFG.PostgresPort.GetString())
+	if err != nil {
+		return errors.Wrap(err, "the Postgres port "+config.CFG.PostgresPort.GetString()+" is already in use. Either Airflow is already running in this Astro project, another Astro project, or another service is using this port")
```
neel-astro, line 508: IMO if the project is already running, then the `cannot start, project already running` error should take precedence over this error, so ideally both these checks should be in the `airflow` package after `d.composeService.Ps` in the `Start` method

cmd/airflow.go
neel-astro: It is set to "postgres" as default because that value is used by webserver & scheduler container to connect to postgres container, in which case localhost or 0.0.0.0 or 127.0.0.0 does not work for config.CFG.PostgresHost.

neel-astro (commented):
left a minor comment in terms of logic placement

cmd/airflow.go
sunkickr: d.composeService.Ps doesn't run till after the image is built. This can be very ignoring for users who's project takes 5+min to build. They have to wait all that time just to learn that the ports are already allocated

cmd/airflow_test.go
```diff
@@ -1155,3 +1189,6 @@ func TestPrepareDefaultAirflowImageTag(t *testing.T) {
 		assert.Equal(t, airflowversions.DefaultRuntimeVersion, resp)
 	})
 }
+
+func TestCheckPort(t *testing.T) {
```
kushalmalani, line 1193: Why do we have a empty test case?

cmd/airflow.go
neel-astro: `d.composeService.Ps` check runs upfront, before building any images or any other operation.

Reference: https://github.com/astronomer/astro-cli/blob/main/airflow/docker.go#L159-L172

Screenshot 2022-11-22 at 10 12 42 AM

frankcash:
#802

PR #819, opened 2022-10-17
Sql cli integration

pankajkoti:
## Description

Integrate SQL CLI with Astro CLI

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


Dockerfile.sql_cli
```diff
@@ -0,0 +1,9 @@
+ARG PYTHON_MAJOR_MINOR_VERSION="3.9"
+# ARG PYTHON_BASE_IMAGE="python:${PYTHON_MAJOR_MINOR_VERSION}-slim-bullseye"
+ARG PYTHON_BASE_IMAGE="python:${PYTHON_MAJOR_MINOR_VERSION}"
```
kaxil, line 3: Why commented code?

pankajkoti: I used a non-slim image so that it could have required libraries that are needed for python-sdk installation. But, figured out now required gcc installation and the slim base image works now. Removed the comment.

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+		fmt.Println(err)
+		return ""
+	}
+	defer res.Body.Close()
+
+	type response struct {
+		Info struct {
+			Version string `json:"version"`
+		} `json:"info"`
+	}
```
neel-astro, line 42: please move this out of the function logic, it is always better to keep struct definition global

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+	var resp response
+	err = json.NewDecoder(res.Body).Decode(&resp)
+	if err != nil {
+		fmt.Println(err)
+		return ""
```
neel-astro, line 47: same here

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/docker/docker/pkg/archive"
+	"github.com/docker/docker/pkg/stdcopy"
+)
+
+func getPypiVersion() string {
+	url := "https://pypi.org/pypi/astro-sql-cli/json"
+	method := "GET"
+
+	httpClient := &http.Client{}
+	req, err := http.NewRequest(method, url, http.NoBody)
```
neel-astro, line 26:
```suggestion
	req, err := http.NewRequest(http.MethodGet, url, http.NoBody)
```

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+func CommonDockerUtil(cmd []string, flags map[string]string) error {
+	ctx := context.Background()
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		panic(err)
+	}
+
+	astroSQLCliVersion := getPypiVersion()
+	opts := types.ImageBuildOptions{
+		Dockerfile: "Dockerfile.sql_cli",
```
neel-astro, line 67: can we leave a TODO here to fix this if we are not fixing it in this PR, as now users would need the Dockerfile in their local machine alongside where the CLI binary resides to make this work

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if astroSQLCliVersion != "" {
+		opts.BuildArgs = map[string]*string{"VERSION": &astroSQLCliVersion}
+	}
+
+	_, currentfilePath, _, ok := runtime.Caller(0)
+	if !ok {
+		fmt.Println("Failed to get current file path and hence cannot locate Dockerfile for building the image")
+	}
+	dockerfilePath := path.Join(path.Dir(currentfilePath), "../Dockerfile.sql_cli")
```
neel-astro, line 79: I feel it is extremely wrong to assume that Docerfile will always have a fixed relative location from the CLI binary. Can we at least set an environment variable whose value is the location of the Dockerfile, and then read the value of that env variable over here?

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+	buf := new(strings.Builder)
+	_, err = io.Copy(buf, body.Body)
+	if err != nil {
+		fmt.Printf("Image build response read failed %s", err)
+	}
+
+	for key, value := range flags {
+		cmd = append(cmd, []string{fmt.Sprintf("--%s", key), value}...)
+	}
+	fmt.Println(cmd)
```
neel-astro, line 94: is this required?

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+	resp, err := cli.ContainerCreate(ctx, &container.Config{
+		Image: "sql_cli",
```
neel-astro, line 96: can the image name be constant

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+		Cmd:   cmd,
+		Tty:   false,
+	}, nil, nil, nil, "")
+	if err != nil {
+		panic(err)
```
neel-astro, line 101: handle error pls instead of panicking

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+
+	_, currentfilePath, _, ok := runtime.Caller(0)
+	if !ok {
+		fmt.Println("Failed to get current file path and hence cannot locate Dockerfile for building the image")
+	}
+	dockerfilePath := path.Join(path.Dir(currentfilePath), "../Dockerfile.sql_cli")
+	fmt.Println("Building image...")
+	body, err := cli.ImageBuild(ctx, getContext(dockerfilePath), opts)
+	if err != nil {
+		fmt.Printf("Image building failed %v ", err)
```
neel-astro, line 83: we should return the error from here instead of printing and moving forward, since I don't think so rest of the logic matters in case image build itself fail 🤷‍♂️

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+func getContext(filePath string) io.Reader {
+	ctx, _ := archive.TarWithOptions(filePath, &archive.TarOptions{})
+	return ctx
+}
+
+func CommonDockerUtil(cmd []string, flags map[string]string) error {
+	ctx := context.Background()
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		panic(err)
```
neel-astro, line 62: handle errors pls instead of panicking

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+		Image: "sql_cli",
+		Cmd:   cmd,
+		Tty:   false,
+	}, nil, nil, nil, "")
+	if err != nil {
+		panic(err)
+	}
+
+	if err := cli.ContainerStart(ctx, resp.ID, types.ContainerStartOptions{}); err != nil {
+		panic(err)
```
neel-astro, line 105: handle error pls instead of panicking

sql/flow.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	statusCh, errCh := cli.ContainerWait(ctx, resp.ID, container.WaitConditionNotRunning)
+	select {
+	case err := <-errCh:
+		if err != nil {
+			panic(err)
+		}
+	case <-statusCh:
```
neel-astro, line 114: nit: we should at least log the status because AFAIK in case the container failed to start that would be captured in `statusCh`

neel-astro (changes requested):
Good work considering this is your first time with Astro CLI and Golang ❤️. At a high level, it looks fine, but have left a bunch of comments, I feel we need to handle errors better in our logic, let it flow back correctly, and also have a bit cleaner setup around using the dockerfile

sql/flow.go
pankajkoti: make sense, made the needed change.

sql/flow.go
pankajkoti: not required, removed.

sql/flow.go
pankajkoti: We are logging error in case something fails while the container is in wait state. Otherwise it in an okay body. We're have an error for container failing to start above. I think this log would just be an additional log as part of the command output and we might not need it. What do you think?

cmd/sql/flow.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func runCmd() error {
+	vars := make(map[string]string)
+	if environment != "" {
+		vars["environment"] = environment
+	}
+	if connection != "" {
+		vars["connection"] = connection
+	}
```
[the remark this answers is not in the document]
pankajkoti: We're still in the initial stages and not sure what flags would be needed for each command. Is it okay if we refine this later once we have more clarity?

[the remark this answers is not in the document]
pankajkoti: I have moved this to a common function now.

sql/flow.go
kaxil: Can we not replicate what we do with `astro dev init` which creates a `Dockerfile`.

Only difference being we can put it in a specific folder/path.

Asking users to set an environment variable, makes the experience poor and we don't users to know about the Dockerfile at all.

andriisoldatenko:
could you please add missing tests? cc @neel-astro

feluelle:
I have added missing tests for all commands except for unknown errors coming from docker, which just get propagated.

cmd/root.go
```diff
@@ -92,6 +94,12 @@ Welcome to the Astro CLI, the modern command line interface for data orchestrati
 		newAuthCommand(),
 	)
 
+	if config.CFG.SQLCLI.GetBool() {
+		rootCmd.AddCommand(
+			sql.NewFlowCommand(),
+		)
+	}
```
andriisoldatenko, line 101: you can find missing tests by these messages from codecov ^^

feluelle: Yes working on those already. Completely missed the other files, sorry.

tatiana:
> could you please add the missing tests? cc @neel-astro

@andriisoldatenko, my understanding is that @sunkickr was happy for us to have this PR merged if the only pending thing is coverage - so we can have this feature in the 1.7 alpha release. We'll be making a separate PR to cover the bits which do not have tests yet - if that is okay

My understanding is that we addressed all the requested changes, I hope we can get this merged soon :crossed_fingers:

andriisoldatenko:
> > could you please add the missing tests? cc @neel-astro
> 
> @andriisoldatenko, my understanding is that @sunkickr was happy for us to have this PR merged if the only pending thing is coverage - so we can have this feature in the 1.7 alpha release. We'll make a separate PR to cover the bits that do not have tests yet - if that is okay.
> 
> Current status: `codecov/patch — 80.09% of diff hit (target 87.15%)`
> 
> Would it be okay if we temporarily reduce the lower bound of code coverage to 80 in https://github.com/astronomer/astro-cli/blob/main/codecov.yml#L3, and we'll rollback this change in the next PR? https://github.com/astronomer/astro-cli/blob/main/codecov.yml#L3
> 
> My understanding is that we addressed all the requested changes. I hope we can get this merged soon 🤞

sounds good to me.

sql/flow.go
feluelle: Hey @neel-astro, I have re-added panic'ing as I think there is nothing the caller can do to fix it.

Please let me know what you think i.e. why you think we should `return` the error.

cc @tatiana @pankajkoti

sunkickr (approved):
LGTM

sql/flow.go
feluelle: @jemishp @kushalmalani I would be interested in what you think of using `panic` in this case.

feluelle:
@sunkickr thanks, but I just have added a new commit (with discussion in https://github.com/astronomer/astro-cli/pull/819#discussion_r1006885335). I am waiting now for the devs to review once more.

cmd/sql/flow.go
```diff
@@ -0,0 +1,295 @@
...
+
+	vars := buildCommonVars()
+	mountDirs, err := getBaseMountDirs(projectDir)
+	if err != nil {
+		return err
+	}
+
+	projectDir, err = getAbsolutePath(projectDir)
+	if err != nil {
+		return err
```
kushalmalani, line 180: These lines of code are used across multiple functions in this file...Any chance we move it to a common function and re-use and not duplicate code

sql/flow.go
kushalmalani: Can you return the error and handle gracefully?

kushalmalani (approved):

sql/flow.go
jemishp: Instead of panicking, please return the `err` from `CommonDockerUtil()` so that when `executeCmd()` calls it; it can check for err and log it or return it back to the user for immediate feedback.

sql/flow.go
```diff
@@ -0,0 +1,161 @@
...
+func getContext(filePath string) io.Reader {
+	ctx, _ := archive.TarWithOptions(filePath, &archive.TarOptions{})
+	return ctx
+}
+
+func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []string) {
+	ctx := context.Background()
+	cli, err := newDockerClient()
+	if err != nil {
+		err = fmt.Errorf("docker client initialisation failed for command %v: %w", cmd, err)
```
jemishp, line 80: fix typo for initialization

jemishp (changes requested):
Left some comments around panicking but other than that looks fine.

sql/flow.go
tatiana: That's the British spelling - not necessarily a typo :see_no_evil: Do we use American spelling in the Astro CLI?

jemishp: I was wondering about that British / American too... :)

jemishp (approved):

sql/flow.go
kushalmalani: Yeah, American English is what we use 😅

PR #847, opened 2022-10-31
cli keeps calling non paginated endpoint even when interactive flag is on

ajayy004:
## Description

_**Issue**_: CLI makes a workspace list query to identify if there is 1 or more workspace

_**Fix**_: If the interactive flag is set to true, CLI will call the paginated workspace query with page size 2 to make the same decision to auto-select the first workspace if it returns 1 record if it returns more than 1. The user will be prompted to select a workspace 

_**Enhancement**_: Added new query to validate `workspaceId`, which does not pull `roleBindings` since it's not required to validate `workspaceId`.

## 🎟 Issue(s)

Related https://github.com/astronomer/issues/issues/5166

## 🧪 Functional Testing

with-interactive-flag-set-to-true

> with-interactive-flag-set-to-false
with-interactive-flag-set-to-false

## 📋 Checklist

### DIFF
--- a/software/auth/auth.go
+++ b/software/auth/auth.go
@@ -54,20 +54,24 @@ func basicAuth(username, password string, ctx *config.Context, client houston.Cl
 	return client.AuthenticateWithBasicAuth(username, password, ctx)
 }
 
-var switchToLastUsedWorkspace = func(c *config.Context, workspaces []houston.Workspace) bool {
+var switchToLastUsedWorkspace = func(client houston.ClientInterface, c *config.Context) bool {
 	if c.LastUsedWorkspace == "" {
 		return false
 	}
 
-	for idx := range workspaces {
-		if c.LastUsedWorkspace == workspaces[idx].ID {
-			if err := c.SetContextKey("workspace", workspaces[idx].ID); err != nil {
-				return false
-			}
-			return true
-		}
+	// validate workspace
+	workspace, err := client.ValidateWorkspaceId(c.LastUsedWorkspace)
+	if err != nil || workspace != nil && workspace.ID != c.LastUsedWorkspace {
+		log.Debugf("last used workspace id is not valid: %s", err.Error())
+		return false
 	}
-	return false
+
+	if err := c.SetContextKey("workspace", workspace.ID); err != nil {
+		log.Debugf("unable to set workspace context: %s", err.Error())
+		return false
+	}
+
+	return true
 }
 
 // oAuth handles oAuth with houston api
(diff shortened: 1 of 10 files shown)


software/auth/auth.go
```diff
@@ -173,7 +175,12 @@ func Login(domain string, oAuthOnly bool, username, password string, client hous
 		return err
 	}
 
-	workspaces, err := client.ListWorkspaces()
+	if interactive {
+		workspaces, err = client.PaginatedListWorkspaces(2, 0)
```
ajayy004, line 179: Since we need to identify if the number of workspaces returned 1 or more, calling workspace pagination with page size 2 if the interactive flag is set to true

software/auth/auth.go
```diff
@@ -127,6 +127,20 @@ func registryAuth(client houston.ClientInterface, out io.Writer) error {
 	return nil
 }
 
+func getWorkspaces(client houston.ClientInterface, interactive bool) ([]houston.Workspace, error) {
+	var workspaces []houston.Workspace
+	var err error
+
+	if interactive {
+		workspacePageSize := 2
```
neel-astro, line 135: why not use the `PageSize` value from `config.CFG.PageSize` or else `defaultPageSize`?
Reference: https://github.com/astronomer/astro-cli/blob/main/software/auth/auth.go#L135

neel-astro (changes requested):

software/auth/auth.go
ajayy004: Since we need to identify if the number of workspaces returned 1 or more, calling workspace pagination with page size 2 if the interactive flag is set to true

neel-astro: That's not entirely true, because the list of workspaces returned from this method is used to switch to `last_used_workspace`, and by just pulling two workspaces, that functionality will almost always be broken in case interactive is set to true.

So from a UX point of view, if interactive is set to true, then whenever a user tries to log in, they would be asked to select a particular workspace every single time, and not automatically set the last used workspace from the previous login as the current workspace.

We should document this in our release note if we are going with this approach, and also add comments in the code around the logic of choosing the page size as 2 for future reference.

Might want to check once on this functional change with @cmarteepants if not already done.

cmarteepants: The trade-off (selecting the workspace every single time a user logs in) is too expensive from a UX standpoint. I'm not following how the change helps solve the initial problem. Is there another approach we can use to solve the problem, while retaining the login UX?

ajayy004: As discussed, I'll change the implementation of `switchToLastUsedWorkspace` to use `getWorkspace` API to identify if the user still has access/available  `last_used_workspace ` workspace. That way will have the same user experience.

software/auth/auth.go
```diff
@@ -127,6 +131,20 @@ func registryAuth(client houston.ClientInterface, out io.Writer) error {
...
```
neel-astro, line 139: nit: can we just leave a comment here in the code on why we just need it to be 2 and nothing else

neel-astro (approved):
LGTM, have left in a minor nitpick

software/auth/auth.go
ajayy004: sure

ajayy004: Done

neel-astro (approved):

PR #849, opened 2022-11-01
Sql cli 0.1 tests

pankajkoti:
## Description

Add tests for sql/flow.go 

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


sql/mocks/flow.go
```diff
@@ -0,0 +1,143 @@
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
neel-astro, line 1: Can we add the mockery command used to generate this mock part of the `make mock` command
Reference: https://github.com/astronomer/astro-cli/blob/main/Makefile#L23-L40

sql/pypi_client.go
```diff
@@ -14,6 +14,8 @@ type pypiVersionResponse struct {
 
 const astroSQLCliProjectURL = "https://pypi.org/pypi/astro-sql-cli/json"
 
+var getPypiVersion = GetPypiVersion
+
 func GetPypiVersion(projectURL string) (string, error) {
```
neel-astro, line 19:
```suggestion
var getPypiVersion = func (projectURL string) (string, error) {
```
nit: cleaner way to do the same thing

neel-astro (changes requested):
Looks fine, can we also move these tests to use mocks: https://github.com/astronomer/astro-cli/blob/main/cmd/sql/flow_test.go#L58-L109

sql/pypi_client.go
pankajkoti: We need the function to reassign it to the variable in the tests after having monkey-patched, because the successive tests use the same monkey-patched version otherwise. https://github.com/astronomer/astro-cli/pull/849/files#diff-0d310e499467646455c981ef796fc947f43f23205c90e7806505c0541fe295c5R77

neel-astro: syntactically both ways represent & achieve the same thing, but I am fine with whatever you might be comfortable with

neel-astro (approved):
LGTM

codecov.yml
```diff
@@ -1,4 +1,6 @@
 ---
+ignore:
+  - "sql/docker_interface.go"
```
feluelle, line 3: 🤔 Why does codecov complain about this? Aren't we using this in our tests and code?

feluelle (approved):
LGTM (minor comment)

codecov.yml
pankajkoti: No, we don't use them since we're mocking those method implementations. So, they don't get called and those lines miss out of coverage.

feluelle: Alright, thanks 👍

sunkickr (approved):

kushalmalani (approved):

PR #851, opened 2022-11-01
Add global debug option to flow command and correct flags

feluelle:
## Description

In the sql-cli we are adding a new global `--debug` flag which we need to register to cobra, too. So that astro-cli users can use the flag in flow.

## 🎟 Issue(s)

Related https://github.com/astronomer/astro-sdk/issues/1040

## 📋 Checklist


kaxil:
@kushalmalani @andriisoldatenko Could you review this PR please?

neel-astro (approved):
LGTM

cmd/sql/flow_test.go
```diff
@@ -130,7 +130,7 @@ func TestFlowGenerateCmd(t *testing.T) {
 	err := execFlowCmd([]string{"init", projectDir}...)
 	assert.NoError(t, err)
 
-	err = execFlowCmd([]string{"generate", "example_basic_transform", "--project-dir", projectDir}...)
+	err = execFlowCmd([]string{"generate", "example_basic_transform", "--project-dir", projectDir, "--verbose"}...)
```
tatiana, line 133: Do we have tests for both when we enable `--verbose` and not? Do they have different assertions?

cmd/sql/flow_test.go
```diff
@@ -170,6 +170,15 @@ func TestFlowRunCmd(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestDebugFlowRunCmd(t *testing.T) {
+	projectDir := t.TempDir()
+	err := execFlowCmd([]string{"init", projectDir}...)
+	assert.NoError(t, err)
+
+	err = execFlowCmd([]string{"--debug", "run", "example_templating", "--env", "dev", "--project-dir", projectDir}...)
+	assert.NoError(t, err)
```
tatiana, line 179: Is checking if the command didn't raise an exception good enough? Would there be any other meaningful assertions we could make here?

cmd/sql/flow_test.go
feluelle: Those tests are only testing that the inputs work with cobra. For example, if the flag is unknown, it will raise an error.

feluelle: We can add all flags at once to cover all scenarios of flags added to cobra.

cmd/sql/flow_test.go
feluelle: I think that this is currently enough as cobra will raise an error if the command is incorrect e.g. has too many args or unknown flags, etc.

kushalmalani (approved):

cmd/sql/flow.go
```diff
@@ -149,6 +143,14 @@ func executeValidate(cmd *cobra.Command, args []string) error {
 	projectDirAbsolute := mountDirs[0]
 	args = []string{projectDirAbsolute}
 
+	if environment != "" {
+		flags["env"] = environment
+	}
+
```
pankajkoti, line 149: We also need to add the connection flag for the Validate command.

go.mod
```diff
@@ -24,10 +24,10 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/viper v1.9.0
 	github.com/stretchr/testify v1.7.1
-	golang.org/x/crypto v0.0.0-20220511200225-c6db032c6c88 // indirect
-	golang.org/x/net v0.0.0-20220624214902-1bab6f366d9e
-	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a
-	golang.org/x/text v0.3.7 // indirect
+	golang.org/x/crypto v0.1.0 // indirect
```
pankajkoti, line 27: Do we need changes in this file to be checked in?

go.sum
```diff
@@ -1101,8 +1101,8 @@ golang.org/x/crypto v0.0.0-20201117144127-c1f2f97bffc9/go.mod h1:jdWPYTVW3xRLrWP
 golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83/go.mod h1:jdWPYTVW3xRLrWPugEBEK3UY2ZEsg3UU495nc5E+M+I=
 golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2/go.mod h1:T9bdIzuCu7OtxOm1hfPfRQxPLYneinmdGuTeoZ9dtd4=
 golang.org/x/crypto v0.0.0-20210817164053-32db794688a5/go.mod h1:GvvjBRRGRdwPK5ydBHafDWAxML/pGHZbMvKqRZ5+Abc=
-golang.org/x/crypto v0.0.0-20220511200225-c6db032c6c88 h1:Tgea0cVUD0ivh5ADBX4WwuI12DUd2to3nCYe2eayMIw=
-golang.org/x/crypto v0.0.0-20220511200225-c6db032c6c88/go.mod h1:IxCIyHEi3zRg3s0A5j5BB6A9Jmi73HwBIUl50j+osU4=
+golang.org/x/crypto v0.1.0 h1:MDRAIl0xIo9Io2xV565hzXHw3zVseKrJKodhohM5CjU=
```
pankajkoti, line 1104: Same question, do we need these changes to be checked in?

cmd/sql/flow.go
feluelle: Oh good catch @pankajkoti !

go.mod
feluelle: Yes, [this](https://faun.pub/understanding-go-mod-and-go-sum-5fd7ec9bcc34) can be useful to read. I am reading it right now.

feluelle: Why I had to run `go mod tidy` is because I added a new dependency `golang.org/x/exp/slices`.

pankajkoti (approved):

PR #860, opened 2022-11-07
Print flow installation commands

feluelle:
## Description

Currently, `flow` users are not seeing any progress made during the very first installation. This PR is addressing this by showing the commands for installing it - but only when the layers are not cached yet.

## 🎟 Issue(s)

Related https://github.com/astronomer/astro-sdk/issues/1143

## 📸 Screenshots

First time running `astro flow` command:

Screenshot 2022-11-07 at 16 23 31

...

Screenshot 2022-11-07 at 16 25 39

When installed, it won't show the output as all of it is _cached_:

Screenshot 2022-11-07 at 16 22 07

## 📋 Checklist

### DIFF
--- a/sql/flow.go
+++ b/sql/flow.go
@@ -26,7 +26,6 @@ var (
 	DockerClientInit   = NewDockerClient
 	IoCopy             = io.Copy
 	PrintBuildingSteps = printBuildingSteps
-	Println            = fmt.Println
 )
 
 func getContext(filePath string) io.Reader {
@@ -59,14 +58,10 @@ func printBuildingSteps(r io.Reader) error {
 		//  ---> Running in 0afb2e0c5ad7
 		if strings.HasPrefix(prevStream, "Step ") && strings.HasPrefix(currStream, " ---> Running in ") {
 			if firstMessage {
-				if _, err := Println("Installing flow.. This might take some time."); err != nil {
-					return err
-				}
+				fmt.Println("Installing flow.. This might take some time.")
 				firstMessage = false
 			}
-			if _, err := Println(prevStream); err != nil {
-				return err
-			}
+			fmt.Println(prevStream)
 		}
 		prevStream = currStream
 	}
--- a/sql/flow.go
+++ b/sql/flow.go
@@ -54,6 +54,9 @@ func printBuildingSteps(r io.Reader) error {
 			continue
 		}
 		currStream = jsonMessage.Stream
+		// We only print steps which are actually running, e.g.
+		// Step 2/4 : ENV ASTRO_CLI Yes
+		//  ---> Running in 0afb2e0c5ad7
 		if strings.HasPrefix(prevStream, "Step ") && strings.HasPrefix(currStream, " ---> Running in ") {
 			if firstMessage {
 				if _, err := Println("Installing flow.. This might take some time."); err != nil {
--- a/sql/flow.go
+++ b/sql/flow.go
@@ -15,11 +15,6 @@ import (
 	"github.com/docker/docker/pkg/jsonmessage"
 )
 
-type MountVolume struct {
-	SourceDirectory string
-	TargetDirectory string
-}
-
 const (
 	SQLCliDockerfilePath      = ".Dockerfile.sql_cli"
 	SQLCLIDockerfileWriteMode = 0o600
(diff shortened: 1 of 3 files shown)


sql/flow_test.go
```diff
@@ -54,9 +57,44 @@ func TestCommonDockerUtilSuccess(t *testing.T) {
 		mockDockerBinder.On("ContainerLogs", mock.Anything, mock.Anything, mock.Anything).Return(sampleLog, nil)
 		return mockDockerBinder, nil
 	}
+	PrintBuildingSteps = func(r io.Reader) error {
+		return nil
+	}
```
pankajkoti, line 62: Can we define this nil return function as some mockedPrintBuildingSteps function and use that here? Because the same 3 lines are repeated in most test cases.

sql/flow.go
```diff
@@ -26,21 +23,59 @@ const (
...
+			return err
+		}
+		if err := jsonMessage.Error; err != nil {
+			return err
+		}
+		if jsonMessage.Stream == "\n" {
+			continue
+		}
+		currStream = jsonMessage.Stream
+		if strings.HasPrefix(prevStream, "Step ") && strings.HasPrefix(currStream, " ---> Running in ") {
```
pankajkoti, line 57: Can we please add a comment here to describe what the conditions mean here?

pankajkoti (commented):
Looks great to me, a couple of minor inline comments.

feluelle:
@pankajkoti I have addressed your comments. PTAL

pankajkoti (approved):

sql/flow.go
```diff
@@ -26,21 +23,62 @@ const (
...
+			continue
+		}
+		currStream = jsonMessage.Stream
+		// We only print steps which are actually running, e.g.
+		// Step 2/4 : ENV ASTRO_CLI Yes
+		//  ---> Running in 0afb2e0c5ad7
+		if strings.HasPrefix(prevStream, "Step ") && strings.HasPrefix(currStream, " ---> Running in ") {
+			if firstMessage {
+				if _, err := Println("Installing flow.. This might take some time."); err != nil {
+					return err
```
neel-astro, line 63:
```suggestion
				fmt.Println("Installing flow.. This might take some time.")
```
I don't think it is important to check the error for `fmt.Println`, even if `fmt.Println` most likely have an error, then the returned error from here might also not be logged

sql/flow.go
```diff
@@ -26,21 +23,62 @@ const (
...
+				}
+				firstMessage = false
+			}
+			if _, err := Println(prevStream); err != nil {
+				return err
```
neel-astro, line 68:
```suggestion
			fmt.Println(prevStream)
```

sql/flow_test.go
```diff
@@ -54,9 +69,42 @@ func TestCommonDockerUtilSuccess(t *testing.T) {
...
+Step 2/4 : ENV ASTRO_CLI Yes
+`
+	// Capture output from PrintBuildingSteps Println calls
+	actualOutput := ""
+	Println = func(a ...any) (n int, err error) {
+		for _, arg := range a {
+			actualOutput += fmt.Sprintln(arg)
+		}
+		return 0, nil
+	}
```
neel-astro, line 90: a simpler way to do the same thing would be to listen to `os.Stdout` (Reference: https://github.com/astronomer/astro-cli/blob/main/airflow/docker_test.go#L673-L691) and capture everything. IMO ideally, monkey-patching the code to write unit tests should be the last resort, it makes code less readable and prone to misuse.

sql/flow.go
```diff
@@ -26,21 +23,62 @@ const (
...
+	PrintBuildingSteps = printBuildingSteps
+	Println            = fmt.Println
 )
 
 func getContext(filePath string) io.Reader {
 	ctx, _ := archive.TarWithOptions(filePath, &archive.TarOptions{})
 	return ctx
 }
 
+func printBuildingSteps(r io.Reader) error {
```
neel-astro, line 37:
```suggestion
	Println            = fmt.Println
)

func getContext(filePath string) io.Reader {
	ctx, _ := archive.TarWithOptions(filePath, &archive.TarOptions{})
	return ctx
}

var PrintBuildingSteps = func(r io.Reader) error {
```
nit: a simpler way to do the same thing

neel-astro (commented):
looks fine, have left minor nitpicks

sql/flow_test.go
feluelle: That's exactly what I (and @pankajkoti) were looking for. Thanks.

sql/flow.go
feluelle: Yes, but I think it looks cleaner having it as separate function - not as inline function in vars.

feluelle:
Hey @neel-astro, thanks for your review. I have addressed your feedback. PTAL

neel-astro (approved):
LGTM

tatiana:
@feluelle, thanks for getting this done. It certainly improves the user experience.
I also liked the refactorings and tidying ups - but it may be better to those as part of a separate PR in future, since we end up squashing them :D

PR #873, opened 2022-11-14
cli core services setup and migration

missing1984:
### This PR migrated 3 endpoints to consume core services like UI does today.

1.  listOrganizations
2. getSelfUser
3. createUserInvite 

### 3 commands are impacted by this change
1. `astro organization list`. - consume listOrganizations endpoint
2. `astro user invite`. - consume createUserInvite endpoint
3. `astro login` - consume both listOrganizations and getSelfUser endpoints

### Openapi codegen and core service client at `astro-client-core/*`
* `github.com/deepmap/oapi-codegen` is brought in to generate the openapi client for core services. all generated code are in `astro-client-core/api.gen.go`.  
* `astro-client-core/client.go` wrap up the integration of generated client and our context (setting auth headers)

### `cmd/*` layer change
* initialized  `astrocore.CoreClient` from root and pass down to all cloud commands

### `cloud/*` layer change.
* modified organization package to consume core service only.
* modified user package to consume core service only.
* modified auth package to consume core service getSelf and list organization.

### Other notable change
* `orgShortName` is a required argument in order to call core services. currently we don't have it persisted in config. a `migration.go` module is introduced that backfill the `organization_short_name` field. the migration function will run at the end of the cloud setup function.
* `astro-client-core` came with generated mocks. All relevant unit tests are switched over to use the new mocked client.

also, this is my first golang PR so don't hesitate to comment out dumb mistake..

## 🎟 Issue(s)

https://github.com/astronomer/astro/issues/2247

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


cmd/cloud/migration.go
```diff
@@ -0,0 +1,41 @@
...
+var ErrorShortName = errors.New("cannot find organization shortname")
+
+// migrate config
+func migrateCloudConfig(coreClient astrocore.CoreClient) error {
+	ctx, err := context.GetCurrentContext()
+	if err != nil {
+		return err
+	}
+	// backfill OrganizationShortName
+	if (ctx.OrganizationShortName == "") && (ctx.Organization != "") {
```
missing1984, line 20: here is the code that backfill organization_short_name in config

cmd/cloud/setup.go
```diff
@@ -269,17 +278,18 @@ func checkAPIKeys(astroClient astro.Client, args []string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-
-	organizations, err := astroClient.GetOrganizations()
+	orgs, err := organization.ListOrganizations(coreClient)
```
missing1984, line 281: The current code uses GQL services to fetch org list (which differ from organization package.. don't know why). i've switched to consume core services as well.

cloud/auth/auth.go
```diff
@@ -301,25 +302,42 @@ func switchToLastUsedWorkspace(c *config.Context, workspaces []astro.Workspace)
...
+	orgs := *orgsResp.JSON200
+	// default to first one in case something crazy happen lol
+	activeOrg := orgs[0]
+	for i := range orgs {
+		if orgs[i].Id == authOrgID {
+			activeOrg = orgs[i]
+			break
 		}
 	}
+	err = c.SetOrganizationContext(activeOrg.Id, activeOrg.ShortName)
```
missing1984, line 333: This is guard against a rare use case where user's login org is different than the org he can access. We always treat AuthZ is the source of truth of user org relationship

missing1984:
This is ready for review. i am adding new unit tests now.

Makefile
```diff
@@ -16,6 +18,9 @@ lint:
 build:
 	go build -o ${OUTPUT} -ldflags "${LDFLAGS_VERSION}" main.go
 
+core_api_gen:
+	oapi-codegen -include-tags=User,Organization,Invite -generate=types,client -package=astrocore "${CORE_OPENAPI_SPEC}" > ./astro-client-core/api.gen.go
```
jemishp, line 22: could the makefile test if `oapi-codegen` exists and install if it does not exist. Similar to the linter on line 14.

cloud/auth/auth.go
```diff
@@ -289,25 +290,46 @@ func switchToLastUsedWorkspace(c *config.Context, workspaces []astro.Workspace)
 	return astro.Workspace{}, false, nil
 }
 
-// checkToken requests a users rolebindings and sets the workspace to make sure that token works
-// TODO check orgID is in the token
-func checkToken(c *config.Context, client astro.Client, out io.Writer) error {
-	self, err := client.GetUserInfo()
+// check client status after a successfully login
+func checkUserSession(c *config.Context, client astro.Client, coreClient astrocore.CoreClient, out io.Writer) error {
+	// fetch self user base on token
```
jemishp, line 295:
```suggestion
	// fetch self user based on token
```

cloud/auth/auth.go
```diff
@@ -289,25 +290,46 @@ func switchToLastUsedWorkspace(c *config.Context, workspaces []astro.Workspace)
...
 		return err
 	}
-	organizationID := self.AuthenticatedOrganizationID
-
-	if organizationID != "" {
-		err = c.SetContextKey("organization", organizationID)
-		if err != nil {
-			return err
+	activeOrgID := c.Organization
+	// OrganizationId is optional, it may not returned by getSelf api
```
jemishp, line 306:
```suggestion
	// OrganizationId is optional, it may not be returned by getSelf api
```

cloud/auth/auth.go
```diff
@@ -289,25 +290,46 @@ func switchToLastUsedWorkspace(c *config.Context, workspaces []astro.Workspace)
...
+	}
+	// fetch all orgs that the user can access
+	orgsResp, err := coreClient.ListOrganizationsWithResponse(http_context.Background())
+	err = astrocore.NormalizeAPIError(orgsResp.HTTPResponse, orgsResp.Body, err)
+	if err != nil {
+		return err
+	}
+	orgs := *orgsResp.JSON200
+	// default to first one in case something crazy happen lol
+	activeOrg := orgs[0]
```
jemishp, line 318: do we need to pick the first org? 
It is already set to `activeOrgID := c.Organization` or `*selfResp.JSON200.OrganizationId`

Makefile
missing1984: should be doable, let me fix that

cloud/auth/auth.go
```diff
@@ -289,25 +290,46 @@ func switchToLastUsedWorkspace(c *config.Context, workspaces []astro.Workspace)
...
-// checkToken requests a users rolebindings and sets the workspace to make sure that token works
-// TODO check orgID is in the token
-func checkToken(c *config.Context, client astro.Client, out io.Writer) error {
-	self, err := client.GetUserInfo()
+// check client status after a successfully login
+func checkUserSession(c *config.Context, client astro.Client, coreClient astrocore.CoreClient, out io.Writer) error {
+	// fetch self user base on token
+	// we set CreateIfNotExist to true so we always create astro user when a successfully login
+	createIfNotExist := true
+	selfResp, err := coreClient.GetSelfUserWithResponse(http_context.Background(), &astrocore.GetSelfUserParams{
```
jemishp, line 298: why is it using `http_context` versus  `c *config.Context` which is being passed in?

cloud/auth/auth.go
missing1984: 1. selfResp.JSON200.OrganizationId is an optional field.
2. c.Organization is the existing org from config but user may have been removed from the org.

This is a fallback scenario that in case 1 and 2 happen, we still log user in with the first org that he can access.

cloud/auth/auth.go
missing1984: they are different thing right?  i think `c *config.Context` is our context from config file.

jemishp: you are correct.

cloud/user/user.go
```diff
@@ -1,26 +1,29 @@
...
 import (
+	http_context "context"
 	"fmt"
 	"io"
 
+	astrocore "github.com/astronomer/astro-cli/astro-client-core"
 	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/context"
 
-	"github.com/astronomer/astro-cli/astro-client"
```
jemishp, line 10: yay completely moved over :)

cloud/user/user.go
```diff
@@ -29,45 +34,57 @@ func CreateInvite(email, role string, out io.Writer, client astro.Client) error
 	if err != nil {
 		return err
 	}
-	derivedOrganizationID, err := getOrganizationID(client)
+	ctx, err = context.GetCurrentContext()
```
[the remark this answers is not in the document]
jemishp: Would it have an issue where the inviter's CLI does not have the correct OrgID in its org as the inviter may have been removed from the org?

cloud/user/user_test.go
```diff
@@ -28,88 +28,89 @@ func (t testWriter) Write(p []byte) (n int, err error) {
...
 	})
-	t.Run("error path when CreateUserInvite returns an error", func(t *testing.T) {
+	t.Run("error path when CreateUserInviteWithResponse returns an error", func(t *testing.T) {
 		expectedOutMessage := "failed to create invite: test-inv-error"
 		out := new(bytes.Buffer)
-		mockClient := new(astro_mocks.Client)
-		mockClient.On("GetWorkspace", mock.Anything).Return(expectedWorkspace, nil).Once()
-		mockClient.On("CreateUserInvite", expectedInput).Return(astro.UserInvite{}, errorInvite).Once()
+		mockClient := new(astrocore_mocks.ClientWithResponsesInterface)
+		mockClient.On("CreateUserInviteWithResponse", mock.Anything, mock.Anything, mock.Anything).Return(&createInviteResponseError, nil).Once()
```
jemishp, line 64: We loose the ability to test that our input to the API call was exactly what we expected. For e.g. if we want/need to test that email is always lower case, etc we can do that easily if we didn't have a `mock.Anything` but had the `expectedInput` instead.

missing1984: yep, we pass it directly to api so i thought there's not too much value of testing this ( or if we did normalize the email it can be tested purposely ). I am fine to switch back to use input compare here though

cmd/cloud/migration.go
jemishp: just curious why would `organization_short_name` not be filled in because we already call `c.SetOrganizationSetup()` with it if the user comes in via [checkUserSession() or checkToken()](https://github.com/astronomer/astro-cli/pull/873/files#diff-69d508fdcd15361e006cbb3a7f512757cfafb962e5e3ccad1d7071033bb8b713R325), or via [checkAPIKeys()](https://github.com/astronomer/astro-cli/pull/873/files#diff-d961dc2a8e5c959835ea7f2bd6dbb204a39431e38091c81df1bc813bc14164ebR303)

cmd/cloud/setup.go
jemishp: Nice catch...could be legacy code that was just forgotten. Thanks for migrating it over.

cloud/user/user.go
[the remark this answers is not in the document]
missing1984: core api will error out in this case, also i think this is more align with user's exceptions which invite is sent on behalf of current org. We are good here.

cmd/cloud/migration.go
missing1984: because the code in those place are optional. 
1. checkAPIKeys will only run in the context of api key.
2. checkToken will not backfill if the token is valid as we only backfill short name in login path.

so a "catch all" migration step feels safer. I think in the future we will run into migration scenario as well so having a migration step will be helpful.

cloud/auth/auth_test.go
```diff
@@ -379,80 +405,94 @@ func TestSwitchToLastUsedWorkspace(t *testing.T) {
...
+		mockCoreClient.On("GetSelfUserWithResponse", mock.Anything, mock.Anything).Return(&mockGetSelfResponse, nil).Once()
+		mockCoreClient.On("ListOrganizationsWithResponse", mock.Anything).Return(&mockOrganizationsResponse, nil).Once()
 		mockClient.On("ListWorkspaces", "test-org-id").Return([]astro.Workspace{{ID: "test-id-1"}, {ID: "test-id-2"}}, nil).Once()
 		mockClient.On("ListWorkspaces", "test-org-id").Return([]astro.Workspace{}, errMock).Once()
 		ctx := config.Context{Domain: "test-domain", Organization: "test-org-id"}
 		buf := new(bytes.Buffer)
-		err := checkToken(&ctx, mockClient, buf)
+		err := checkUserSession(&ctx, mockClient, mockCoreClient, buf)
 		assert.NoError(t, err)
 		mockClient.AssertExpectations(t)
```
kushalmalani, line 497: We need to add mockCoreClient.AssertExpectations(t)

kushalmalani: This should be in every test block

cloud/auth/auth.go
missing1984: good catch, so in this case we just display a message in console and ask user reach out to org owner, does it sound right?

kushalmalani (approved):
Left a couple of comments..Code looks ok otherwise. We will be waiting to merge this in until 1.8 pre release is cut.

cloud/auth/auth.go
missing1984: Fixed, throw an error with message `no Organization found. Please contact your Astro Organization Owner to be invited to the organization`

cloud/user/user.go
```diff
@@ -1,26 +1,29 @@
 package user
 
 import (
+	http_context "context"
```
neel-astro, line 4:
```suggestion
	httpContext "context"
```
in order to keep the naming convention consistent

Makefile
```diff
@@ -36,6 +45,9 @@ mock_airflow:
 mock_astro:
 	mockery --filename=Client.go --output=astro-client/mocks --dir=astro-client --outpkg=astro_mocks --name Client
 
+mock_astro_core:
```
neel-astro, line 48: could we also add this to [`mock`](https://github.com/astronomer/astro-cli/pull/873/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R32) as well, so that mocking astro core is also run when someone does `make mock` because `mock` is added so that all the mocks get refreshed via a single command.

Makefile
```diff
@@ -4,6 +4,8 @@ VERSION ?= SNAPSHOT-${GIT_COMMIT_SHORT}
 LDFLAGS_VERSION=-X github.com/astronomer/astro-cli/version.CurrVersion=${VERSION}
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 
+CORE_OPENAPI_SPEC=../astro/apps/core/docs/public/public_v1alpha1.yaml
```
neel-astro, line 7: could we add GitHub URL instead? It might not be necessary for everyone to have that file locally, and more importantly, in case someone has an older version locally, we might risk overwriting it with stale code.

neel-astro (commented):
left few minor comments, otherwise lgtm

astro-client-core/client.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/url"
+
+	"github.com/astronomer/astro-cli/context"
+	"github.com/astronomer/astro-cli/pkg/httputil"
+)
+
+var (
+	ErrorRequest = errors.New("failed to perform request")
+	ErrorBaseURL = errors.New("invalid baseurl")
+	ErrorServer  = errors.New("")
```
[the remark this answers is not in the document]
missing1984: good catch, this is no longer used, removed.

astro-client-core/client.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// a shorter alias
+type CoreClient = ClientWithResponsesInterface
```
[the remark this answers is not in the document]
missing1984: yep, i used Client initially and found out Client is already taken by the generated code. therefore named it CoreClient, i let's keep it for now

Makefile
missing1984: astro is a private repo so this file can not be read publicly.  any idea how to do it in an easy way?

sunkickr:
could you make sure that the refresh tokens code still works. Code is located here: https://github.com/astronomer/astro-cli/blob/e6b73ff9d57df3949a0675eab650f1337782bc82/cmd/cloud/setup.go#L149

missing1984:
> could you make sure that the refresh tokens code still works. Code is located here:
> 
> https://github.com/astronomer/astro-cli/blob/e6b73ff9d57df3949a0675eab650f1337782bc82/cmd/cloud/setup.go#L149

How to test it? I modified the exp date in config and rerun command -- a new token is fetched and saved.

but i ran into an issue that the command which triggers the renew does not return anything.

```
➜  astro-cli git:(main) ./astro workspace list  ->>> this command triggers renew but nothing is output.
 NAME     ID     
➜  astro-cli git:(main) ./astro workspace list  ->>> rerun displayed the right information.
 NAME                                                 ID                              
 Registry Daedalus                                    cl9egxx2b303242cx16ey2eyt6      
 MatthiasSecTesting                                   cla6wzgkd137147ahygq1v2p7nd     
 new ws                                               ckyblko4p57591f35ln2s56lo       
 Workspace 1                                          ckv19qvd2390411dzb71wxbbzn  
```

this is happening on main as well so i don't think it's introduced by my change.

Makefile
neel-astro: I don't think there might be a straightforward way to do it without exposing tokens, since this repo is public. So let's leave it as it is

neel-astro (approved):

kushalmalani (approved):

PR #902, opened 2022-11-24
Respect non-zero exit codes from docker command in flow command

feluelle:
## Description

Currently, we don't respect the docker commands exit code / status code.

With the proposed change, if the docker commands exit code / status code is non-zero we raise an error with the status code.

## 🎟 Issue(s)

Related https://github.com/astronomer/astro-sdk/issues/1234

## 📸 Screenshots

**Before:**

Screenshot 2022-11-24 at 11 50 42

**After:**

Screenshot 2022-12-01 at 11 39 04

## 📋 Checklist


sql/flow.go
```diff
@@ -73,17 +73,18 @@ func displayMessages(r io.Reader) error {
...
 
-func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []string) error {
+func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []string) (int, error) {
+	var statusCode int
 	ctx := context.Background()
 
 	cli, err := DockerClientInit()
 	if err != nil {
-		return fmt.Errorf("docker client initialization failed %w", err)
+		return statusCode, fmt.Errorf("docker client initialization failed %w", err)
```
pankajkoti, line 82: where is the value of statusCode set?

feluelle: The default int is 0 :)

feluelle:
@neel-astro I am having issues getting codecov to recognise the tested lines. I have written tests for all cases codecov is currently reporting on. There shouldn't be even one line not being tested here. In my opinion all of them are false-positives. Any idea why it is doing this to me? 🙄

sql/flow.go
tatiana: @feluelle, what if we set it explicitly, to help the readability? Does lint complain?

feluelle: Yes, it complains.
```
sql/flow.go:78:23: var-declaration: should drop = 0 from declaration of var statusCode; it is the zero value (revive)
        var statusCode int = 0
                             ^
make: *** [lint] Error 1
```

feluelle:
FYI, I have changed the implementation slightly to error out instead of explicitly calling os.exit as with the previous implementation we would kind of overwrite the exit code. Also, the status code from docker containers is _int64_ whereas the exit code for os commands is _int_. So, we would also have to convert and therefore could loose information.

Makefile
```diff
@@ -52,7 +52,9 @@ mock_pkg:
 	mockery --filename=Azure.go --output=pkg/azure/mocks --dir=pkg/azure --outpkg=azure_mocks --name Azure
 
 mock_sql_cli:
-	mockery --filename="flow.go" --output="sql/mocks" --dir=sql/ --outpkg=mocks --name DockerBind
```
kushalmalani, line 55: You might wanna delete the mocks/flow.go file if its not longer going to be used

feluelle: @kushalmalani it has been renamed.

feluelle: @kushalmalani please merge the PR if it looks good to you. Thanks 🙏

sunkickr (approved):

Issue #747, opened 2022-09-22
Users getting wrong error on astro login 


sunkickr:
When users login in the astro context and use an invalid domain they are  getting:

```
Error: HTTP DO Failed: Post "[https://houston.astronomer.feasf:443/v1](https://houston.astronomer.feasf/v1)": dial tcp: lookup houston.astronomer.feasf on 208.67.222.222:53: no such host
```

They should be getting the error:

```
Error! Invalid domain. You are attempting to login into Astro.
Are you trying to authenticate to Astronomer Software? If so, please change your current context with 'astro context switch'
```

Reproduce:
run `astro login astronomer-devf.io` in software context

jemishp:
@sunkickr What should the error message be in this case? For e.g.

PR #912, opened 2022-11-29
improve login flow

jemishp:
## Description

This PR improves UX when a user provides an incorrect cloud domain when logging in. Before this, the invalid domain would incorrectly try and login into a software context as shown below:

```bash
$ astro login fail.astronomer.io
 fail.astronomer.io                  N/A
Error: HTTP DO Failed: Post "https://houston.fail.astronomer.io:443/v1": dial tcp: lookup houston.fail.astronomer.io on 192.168.86.1:53: no such host
```
With this PR we do an extra check on the user's context and print an error if their context was previously set to log into a valid cloud domain. Additionally we give the user an option to login to astronomer software if they want.

## 🎟 Issue(s)

Related #747 

## 🧪 Functional Testing

```bash
$ astro login fail.astronomer.io
Error: fail.astronomer.io is an invalid domain to login into Astro.
Are you trying to authenticate to Astronomer Software? (y/n) n
Canceling login...
```

## 📋 Checklist

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -1,14 +1,15 @@
 package cmd
 
 import (
+	"fmt"
 	"io"
-
-	"github.com/astronomer/astro-cli/pkg/domainutil"
 
 	astro "github.com/astronomer/astro-cli/astro-client"
 	astrocore "github.com/astronomer/astro-cli/astro-client-core"
 	cloudAuth "github.com/astronomer/astro-cli/cloud/auth"
 	"github.com/astronomer/astro-cli/context"
+	"github.com/astronomer/astro-cli/pkg/domainutil"
+	"github.com/astronomer/astro-cli/pkg/input"
 	softwareAuth "github.com/astronomer/astro-cli/software/auth"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/auth.go
```diff
@@ -60,7 +63,14 @@ func login(cmd *cobra.Command, args []string, astroClient astro.Client, coreClie
...
 
 	if len(args) == 1 {
+		// check if user provided a valid cloud domain
 		if !context.IsCloudDomain(args[0]) {
+			// get the domain from context as an extra check
+			ctx, _ := context.GetCurrentContext()
+			if context.IsCloudDomain(ctx.Domain) {
+				// return error if context domain is a valid cloud domain
+				return fmt.Errorf("%w: %s. %s", errInvalidDomain, args[0], "You are attempting to login into Astro.\n Are you trying to authenticate to Astronomer Software? If so, please change your current context with 'astro context switch'")
+			}
```
neel-astro, line 73: why can't we ask for user input at this point in time, a yes/no question? because this error would be hit also for a valid case when a user is logged into Astro and is then trying to log in to a valid Software domain (a valid scenario for almost all migrating customers). This means two extra steps for users which could be avoided I feel.

```
Error: invalid domain: fail.astronomer.io. You are attempting to login into Astro.
 Are you trying to authenticate to Astronomer Software? (y/n)
```

If a user says `yes` at that point in time, then we move on to call `softwareLogin(args[0], oAuth, "", "", houstonVersion, houstonClient, out)` else return out of the function

jemishp: I like the idea. Thanks!

kushalmalani (approved):

PR #924, opened 2022-12-02
Add config command and call config command to mount airflow dirs

pankajkoti:
## Description

Currently, if Airflow home and DAGs directory are set other than the default, those
directories are needed for the execution of commands like generate and run, but those
do not get mounted in the Docker container. SQL CLI now exposes a new command - config; 
to get those values. We now use that config command from SQL CLI internally while running 
generate & run command to get values of the global directories to be mounted.
Additional, we also are adding a cobra command to expose the SQL CLI config command.

## 🎟 Issue(s)

related: https://github.com/astronomer/astro-sdk/issues/1230

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/sql/flow.go
+++ b/cmd/sql/flow.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/astronomer/astro-cli/sql"
 	"github.com/spf13/cobra"
@@ -20,6 +21,7 @@ var (
 	noGenerateTasks   bool
 	verbose           bool
 	debug             bool
+	key               string
 )
 
 func getAbsolutePath(path string) (string, error) {
@@ -59,7 +61,7 @@ func getBaseMountDirs(projectDir string) ([]string, error) {
 	return mountDirs, nil
 }
 
-func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, setAirflowDagsFolder bool) (flags map[string]string, mountDirs []string, err error) {
+func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, setAirflowDagsFolder, mountGlobalDirs bool) (flags map[string]string, mountDirs []string, err error) {
 	flags = make(map[string]string)
 	mountDirs, err = getBaseMountDirs(projectDir)
 	if err != nil {
@@ -72,6 +74,17 @@ func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, se
 			return nil, nil, err
 		}
 		flags["project-dir"] = projectDir
+	}
+
+	if mountGlobalDirs {
+		configFlags := make(map[string]string)
+		configFlags["project-dir"] = projectDir
+		configFlags["key"] = "airflow_home"
+		_, airflowHomeAbs, _ := sql.CommonDockerUtil([]string{"flow", "config"}, nil, configFlags, mountDirs, true)
+		mountDirs = append(mountDirs, strings.TrimSpace(airflowHomeAbs))
+		configFlags["key"] = "airflow_dags_folder"
+		_, airflowDagsFolderAbs, _ := sql.CommonDockerUtil([]string{"flow", "config"}, nil, configFlags, mountDirs, true)
+		mountDirs = append(mountDirs, strings.TrimSpace(airflowDagsFolderAbs))
 	}
 
 	if setAirflowHome && airflowHome != "" {
@@ -100,7 +113,7 @@ func executeCmd(cmd *cobra.Command, args []string, flags map[string]string, moun
 	if debug {
 		cmdString = slices.Insert(cmdString, 1, "--debug")
 	}
-	exitCode, err := sql.CommonDockerUtil(cmdString, args, flags, mountDirs)
+	exitCode, _, err := sql.CommonDockerUtil(cmdString, args, flags, mountDirs, false)
 	if err != nil {
 		return fmt.Errorf("error running %v: %w", cmdString, err)
 	}
@@ -112,7 +125,7 @@ func executeCmd(cmd *cobra.Command, args []string, flags map[string]string, moun
 }
 
 func executeBase(cmd *cobra.Command, args []string) error {
-	flags, mountDirs, err := buildFlagsAndMountDirs(projectDir, false, false, false)
+	flags, mountDirs, err := buildFlagsAndMountDirs(projectDir, false, false, false, false)
 	if err != nil {
 		return err
 	}
@@ -124,7 +137,7 @@ func executeInit(cmd *cobra.Command, args []string) error {
 		projectDir = args[0]
 	}
 
-	flags, mountDirs, err := buildFlagsAndMountDirs(projectDir, false, true, true)
+	flags, mountDirs, err := buildFlagsAndMountDirs(projectDir, false, true, true, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


sql/flow.go
```diff
@@ -141,34 +142,57 @@ func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []s
...
 	if err != nil {
-		return fmt.Errorf("docker container logs fetching failed %w", err)
+		return "", fmt.Errorf("docker container logs fetching failed %w", err)
 	}
 
-	if _, err := IoCopy(os.Stdout, cout); err != nil {
-		return fmt.Errorf("docker logs forwarding failed %w", err)
+	outputLogs := ""
+	if returnStdout {
+		rescueStdout := os.Stdout
```
dimberman, line 168: Could this be separated into a helper function?

dimberman (dismissed):

sql/flow.go
pankajkoti: moved it to a function.

kushalmalani (approved):

cmd/sql/flow.go
```diff
@@ -74,6 +76,17 @@ func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, se
 		flags["project-dir"] = projectDir
 	}
 
+	if mountGlobalDirs {
+		configFlags := make(map[string]string)
+		configFlags["project-dir"] = projectDir
+		configFlags["key"] = "airflow_home"
+		_, airflowHomeAbs, _ := sql.CommonDockerUtil([]string{"flow", "config"}, nil, configFlags, mountDirs, true)
```
feluelle, line 83: We should handle the error as well as status/exit code.

cmd/sql/flow.go
```diff
@@ -74,6 +76,17 @@ func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, se
...
+		mountDirs = append(mountDirs, strings.TrimSpace(airflowHomeAbs))
+		configFlags["key"] = "airflow_dags_folder"
+		_, airflowDagsFolderAbs, _ := sql.CommonDockerUtil([]string{"flow", "config"}, nil, configFlags, mountDirs, true)
```
feluelle, line 86: Same here.

sql/flow.go
```diff
@@ -143,35 +162,43 @@ func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []s
...
+	outputLogs := ""
+	if returnStdout {
+		outputLogs, err = captureStdout(cout)
+		if err != nil {
+			return statusCode, "", err
+		}
+	} else {
+		if _, err := Io().Copy(os.Stdout, cout); err != nil {
+			return statusCode, "", fmt.Errorf("docker logs forwarding failed %w", err)
+		}
```
feluelle, line 196: Since we already copy the containers stdout to the os stdout, I would suggest we read `os.Stdout` outside of the this function.

feluelle (changes requested):

sql/flow.go
pankajkoti: Returning cout instead now and dropping the usage of capturing stdout as discussed. Also, readcloser (cout) can be read only once as it acts as a stream and hence we do not copy it to stdout when we intend to return it for further usage in the calling function.

cmd/sql/flow.go
```diff
@@ -264,6 +315,21 @@ func initCommand() *cobra.Command {
...
+		Use:          "config",
+		Args:         cobra.MaximumNArgs(1),
+		RunE:         executeConfig,
+		SilenceUsage: true,
+	}
+	cmd.SetHelpFunc(executeHelp)
+	cmd.Flags().StringVar(&projectDir, "project-dir", ".", "")
+	cmd.Flags().StringVar(&environment, "env", "default", "")
+	cmd.Flags().StringVar(&key, "key", "", "")
+	cmd.MarkFlagRequired("key") //nolint:errcheck
```
feluelle, line 329: We should ideally let this through and let sql_cli handle it.
```suggestion
```

sql/flow.go
```diff
@@ -143,35 +152,37 @@ func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []s
...
 	if err != nil {
-		return statusCode, fmt.Errorf("docker container logs fetching failed %w", err)
+		return statusCode, nil, fmt.Errorf("docker container logs fetching failed %w", err)
 	}
 
-	if _, err := Io().Copy(os.Stdout, cout); err != nil {
-		return statusCode, fmt.Errorf("docker logs forwarding failed %w", err)
+	if !returnOutput {
+		if _, err := Io().Copy(os.Stdout, cout); err != nil {
+			return statusCode, nil, fmt.Errorf("docker logs forwarding failed %w", err)
```
feluelle, line 179: Same here

sql/flow.go
```diff
@@ -143,35 +152,37 @@ func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []s
...
+		}
 	}
 
 	if err := cli.ContainerRemove(ctx, resp.ID, types.ContainerRemoveOptions{}); err != nil {
-		return statusCode, fmt.Errorf("docker remove failed %w", err)
+		return statusCode, nil, fmt.Errorf("docker remove failed %w", err)
```
feluelle, line 184: and here

sql/flow.go
feluelle: Makes sense 👍

cmd/sql/flow.go
pankajkoti: yes, makes sense. Previously we had it as a flag, but have converted it to a required argument now in commit https://github.com/astronomer/astro-sdk/pull/1338/commits/b1d35465ced1484974c40abc076fec01096f72e3

sql/flow.go
```diff
@@ -143,35 +152,37 @@ func CommonDockerUtil(cmd, args []string, flags map[string]string, mountDirs []s
...
 	case err := <-errCh:
 		if err != nil {
-			return statusCode, fmt.Errorf("docker container wait failed %w", err)
+			return statusCode, nil, fmt.Errorf("docker container wait failed %w", err)
 		}
 	case status := <-statusCh:
 		statusCode = status.StatusCode
 	}
 
 	cout, err := cli.ContainerLogs(ctx, resp.ID, types.ContainerLogsOptions{ShowStdout: true, ShowStderr: true})
```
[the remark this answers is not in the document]
pankajkoti: yes, done this.

cmd/sql/flow.go
```diff
@@ -74,6 +75,40 @@ func buildFlagsAndMountDirs(projectDir string, setProjectDir, setAirflowHome, se
...
+
+		args = []string{"airflow_dags_folder"}
+		exitCode, output, err = sql.CommonDockerUtil([]string{"flow", "config"}, args, configFlags, mountDirs, true)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error running %v: %w", configCommandString, err)
+		}
+		if exitCode != 0 {
+			return nil, nil, sql.DockerNonZeroExitCodeError(exitCode)
+		}
+		airflowDagsFolderAbs, err := sql.ConvertReadCloserToString(output)
```
tatiana, line 105: @pankajkoti I noticed a couple of parts o the code we're not covering with tests. Would it make sense to extract these into smaller functions and increase the test coverage? Do you have any other ideas?

pankajkoti: yes, good idea. Refactored and we have a good coverage now.

sql/flow.go
pankajkoti: we no longer use this.

jemishp (approved):
LGTM

PR #973, opened 2022-12-20
As sysadmin not able to create the deployment from CLI without airflow version tag

ajayy004:
## Description
 Houston has a logic to determine default trigger count based on airflow version, so no need to provide default trigger count

## 🎟 Issue(s)

Related https://github.com/astronomer/issues/issues/5265

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


cmd/software/deployment.go
```diff
@@ -143,7 +143,7 @@ func newDeploymentCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	if triggererEnabled {
-		cmd.Flags().IntVarP(&createTriggererReplicas, "triggerer-replicas", "", 1, "Number of replicas to use for triggerer airflow component, valid 0-2")
+		cmd.Flags().IntVarP(&createTriggererReplicas, "triggerer-replicas", "", -1, "Number of replicas to use for triggerer airflow component, valid 0-2")
```
neel-astro, line 146: The default value set here is also added in the help function, which could mislead users to believe that by default triggerer won't get created. A cleaner approach to do the same thing would be to use keep default here as `1` and then have the logic to check if the a value has been passed by the user or not in `deploymentCreate` function.

The logic could look like this:
```
if !cmd.Flags().Changed("triggerer-replicas") {
        createTriggererReplicas = -1
}
```

Reference: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed

neel-astro (commented):
left a minor comment, looks fine otherwise

cmd/software/deployment.go
```diff
@@ -143,7 +143,11 @@ func newDeploymentCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	if triggererEnabled {
-		cmd.Flags().IntVarP(&createTriggererReplicas, "triggerer-replicas", "", 1, "Number of replicas to use for triggerer airflow component, valid 0-2")
+		cmd.Flags().IntVarP(&createTriggererReplicas, "triggerer-replicas", "", 0, "Number of replicas to use for triggerer airflow component, valid 0-2")
+	}
+
+	if !cmd.Flags().Changed("triggerer-replicas") {
+		createTriggererReplicas = -1
 	}
```
neel-astro, line 151: This logic needs to be in `deploymentCreate` function in this file, over here `cmd.Flags().Changed("triggerer-replicas")` will always evaluate to `false`

neel-astro (changes requested):

software/deployment/deployment.go
```diff
@@ -140,15 +138,8 @@ func CheckTriggererEnabled(client houston.ClientInterface) bool {
...
-	// reset triggerer count to zero in case airflowVersion < 2.2.0
-	if airflowVersion != "" {
-		if version, _ := semver.NewVersion(airflowVersion); version != nil && version.LessThan(triggererAllowedAirflowVersion) {
-			triggererReplicas = 0
-		}
-	}
-
-	if CheckTriggererEnabled(client) {
+func addTriggererReplicasArg(vars map[string]interface{}, client houston.ClientInterface, triggererReplicas int) {
+	if CheckTriggererEnabled(client) && triggererReplicas != -1 {
```
neel-astro, line 142: not exactly related to this function, but if you could add a unit test case for the createDeployment logic where triggererReplicas is -1 that would be great.

neel-astro (approved):
functionally LGTM, left a minor nitpick to add a unit test to cover the newly added scenario.

neel-astro (approved):

Issue #904, opened 2022-11-28
Tell user about new CLI release when they run a command


sunkickr:
- [X] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

Users need to know when the latest release of the CLI is avalible and if they are on an old release

**🧩 Describe the solution you'd like**

Have a process run before every command runs that checks if the user is on the latest version of the CLI. If they are not give them the following message
```
A new release of astro is available: <new release> To upgrade, run: brew upgrade astro <link to latest release on github>
```

**🤔 Describe alternatives you've considered**

**💡 Additional context**

https://astronomer.productboard.com/feature-board/2193567-step-1-feature-backlog/features/15729999/detail

PR #985, opened 2022-12-22
New release message

sunkickr:
## Description

This PR makes it so users will be told if a newer release is available

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/904

## 🧪 Functional Testing

tested manually but it needs to be tested as pre-release

## 📸 Screenshots

Screenshot 2022-12-22 at 4 42 42 PM

## 📋 Checklist

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -18,7 +18,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"github.com/google/go-github/v28/github"
+	"github.com/google/go-github/v48/github"
 
 )
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -14,9 +14,12 @@ import (
 	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/pkg/ansi"
 	"github.com/astronomer/astro-cli/pkg/httputil"
+	"github.com/astronomer/astro-cli/version"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/google/go-github/v28/github"
+
 )
 
 var (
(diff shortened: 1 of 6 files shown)


version/version.go
```diff
@@ -15,3 +21,63 @@ func PrintVersion() {
...
+		}
+		if latestRelease == nil || release.GetTagName() > latestRelease.GetTagName() {
+			latestRelease = release
+		}
+	}
+
+	return latestRelease, nil
+}
+
+func CompareVersions(client *github.Client, owner, repo string) error {
```
neel-astro, line 49: ideally, we shouldn't be returning errors from here, because in case there is some issue with the check we should still be executing the actual command, rather than failing here and blocking everything for the user.

go.mod
```diff
@@ -86,7 +86,10 @@ require (
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/mock v1.6.0 // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/google/go-cmp v0.5.8 // indirect
+	github.com/google/go-cmp v0.5.9 // indirect
+	github.com/google/go-github/v28 v28.1.1 // indirect
```
neel-astro, line 90: can you run `go mod tidy` please to remove unused dependencies

neel-astro (commented):
love this change, have left minor comments

version/version.go
sunkickr: can we log the error?

go.mod
sunkickr: fixed

version/version.go
sunkickr: fixed

cmd/root.go
```diff
@@ -63,6 +65,16 @@ func NewRootCmd() *cobra.Command {
...
 Welcome to the Astro CLI, the modern command line interface for data orchestration. You can use it for Astro, Astronomer Software, or Local Development.`,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Check for latest version
+			if config.CFG.UpgradeMessage.GetBool() {
+				// create github client
+				githubClient := github.NewClient(nil)
+				// compare current version to latest
+				err = version.CompareVersions(githubClient, "astronomer", "astro-cli")
+				if err != nil {
+					logrus.Debug(err)
```
neel-astro, line 75: you would need to add this log to the `InitDebugLogs`, directly writing logrus.Debug won't work.
```suggestion
					softwareCmd.InitDebugLogs = append(softwareCmd.InitDebugLogs, fmt.Errorf("Error comparing CLI versions: %s", err.Error()))
```

neel-astro (approved):
Have left a minor comment. looks good otherwise

PR #990, opened 2022-12-29
add user list and update commands

sunkickr:
## Description

add user list and update commands

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/992
- https://github.com/astronomer/astro-cli/issues/991

## 🧪 Functional Testing

tested localy

## 📸 Screenshots

astro user update
Screenshot 2022-12-29 at 2 14 02 PM

astro user list
Screenshot 2022-12-29 at 2 14 29 PM

## 📋 Checklist


cloud/user/user.go
```diff
@@ -55,6 +62,53 @@ func CreateInvite(email, role string, out io.Writer, client astrocore.CoreClient
...
+		return err
+	}
+	ctx, err := context.GetCurrentContext()
+	if err != nil {
+		return err
+	}
+	if ctx.OrganizationShortName == "" {
+		return ErrNoShortName
+	}
+	// Get all org users. Setting limit to 1000 for now
```
jemishp, line 78: Is 1000 users too much? What order are they coming in?

cmd/cloud/user.go
```diff
@@ -41,6 +47,36 @@ func newUserInviteCmd(out io.Writer) *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"inv"},
+		Short:   "List all the users in your Astro Organization",
+		Long:    "List all the users in your Astro Organization",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listUsers(cmd, out)
+		},
+	}
+	cmd.Flags().IntVarP(&limit, "limit", "l", limitDefault, "Maximum number of organization users listed")
```
jemishp, line 60:
```suggestion
	cmd.Flags().IntVarP(&limit, "limit", "l", limitDefault, "Maximum number of organization users listed. By default it lists 100 users.")
```

jemishp (changes requested):
Left a few comments and please add missing tests.

cloud/user/user.go
sunkickr: I think they are coming in a random order right now. Not sure what order they should be coming in

cmd/cloud/user.go
neel-astro, line 60: I could be wrong, but I think the implementation of the limit is slightly misleading because as of now we would list all the users in the organization irrespective of the limit passed. This limit value as of now is only being used in the API call pagination, but we always end up fetching all the users in the end.

neel-astro (changes requested):
left some comments, looks fine otherwise

cmd/cloud/user.go
sunkickr: so should we not expose the limit to the user? what to you think the limit should be?

neel-astro: Either we can only output the number of records equal to the limit passed by the user and exit or we can use limit as page size and paginate the whole response, i.e. if the limit is 10, then we list the first 10 users, then ask for user input for next 10 sets of users or exit like a web page. The software side has something similar, reference: https://github.com/astronomer/astro-cli/pull/662
If we are not comfortable with either then we shouldn't allow users to pass the limit at all.

sunkickr: I would like to just list all the users by default. I don't see any organization have too many users for that. How would we implement that? what should we set the limit too?

neel-astro: well in that case we could have a limit specified inside the code only (could be 100 to start with), and don't expose it to the users via flag since it is more of an internal CLI logic rather than a UX option.

cloud/user/user.go
```diff
@@ -55,6 +62,53 @@ func CreateInvite(email, role string, out io.Writer, client astrocore.CoreClient
...
+		email = user.Username
+		if err != nil {
+			return err
+		}
+	} else {
+		for i := range users {
+			if users[i].Username == email {
+				userID = users[i].Id
+			}
+		}
```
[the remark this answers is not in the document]
jemishp: +1 on new endpoint to get query `userID` directly.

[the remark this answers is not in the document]
sunkickr: The API can do this with a user ID but can't do this with an email yet. We don't expect users to know the user id of the user they are updating

cloud/user/user.go
```diff
@@ -55,6 +62,53 @@ func CreateInvite(email, role string, out io.Writer, client astrocore.CoreClient
...
+	}
```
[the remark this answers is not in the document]
sunkickr: good idea

cmd/cloud/organization.go
```diff
@@ -107,6 +111,67 @@ func newOrganizationExportAuditLogs(_ io.Writer) *cobra.Command {
 	return cmd
 }
 
+func neOrganizationwUserRootCmd(out io.Writer) *cobra.Command {
```
kushalmalani, line 114:
```suggestion
func newOrganizationUserRootCmd(out io.Writer) *cobra.Command {
```

cloud/user/user.go
```diff
@@ -67,3 +129,99 @@ func IsRoleValid(role string) error {
...
+
+	userMap := map[string]astrocore.User{}
+	for i := range users {
+		index := i + 1
+		table.AddRow([]string{
+			strconv.Itoa(index),
+			users[i].FullName,
+			users[i].Username,
+			users[i].Id,
+			*users[i].OrgRole,
```
kushalmalani, line 150: Why is this a pointer, since the value is just a constant string?

kushalmalani (approved):
Left a couple of comments. Otherwise code looks ok

cloud/user/user.go
sunkickr: because it's a pointer in API gen go

PR #994, opened 2022-12-30
add workspace user add, list, update, and remove commands

sunkickr:
## Description

add workspace user list, update, and remove commands

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/995

## 🧪 Functional Testing

Testing manually 

## 📸 Screenshots

workspace user list
Screenshot 2022-12-30 at 9 11 28 AM

workspace user update
Screenshot 2022-12-30 at 9 12 14 AM

workspace user remove
Screenshot 2022-12-30 at 9 12 54 AM

## 📋 Checklist


cloud/user/user.go
```diff
@@ -225,3 +240,227 @@ func ListOrgUsers(out io.Writer, client astrocore.CoreClient) error {
...
+	}
+	resp, err := client.DeleteWorkspaceUserWithResponse(httpContext.Background(), ctx.OrganizationShortName, workspace, userID)
+	if err != nil {
+		return err
+	}
+	err = astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(out, "The workspace user %s was successfully removed from the workspace\n", email)
```
jemishp, line 464: nit 
```suggestion
	fmt.Fprintf(out, "The user %s was successfully removed from the workspace\n", email)
```

cmd/cloud/workspace.go
```diff
@@ -71,6 +147,47 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
 	return workspace.Switch(id, astroClient, out)
 }
 
+func addWorkspaceUsers(cmd *cobra.Command, args []string, out io.Writer) error {
+	var email string
+
+	// if an email was provided in the args we use it
+	if len(args) > 0 {
+		email = args[0]
```
jemishp, line 155: should this be lower cased and validated as an email before using it?

cmd/cloud/workspace.go
```diff
@@ -71,6 +147,47 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
...
+	cmd.SilenceUsage = true
+	return user.ListWorkspaceUsers(out, astroCoreClient, "")
+}
+
+func updateWorkspaceUsers(cmd *cobra.Command, args []string, out io.Writer) error {
+	var email string
+
+	// if an email was provided in the args we use it
+	if len(args) > 0 {
+		email = args[0]
```
jemishp, line 172: should this be lower cased and validated as an email before using it?

cmd/cloud/workspace.go
```diff
@@ -71,6 +147,47 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
...
+	cmd.SilenceUsage = true
+	return user.UpdateWorkspaceUserRole(email, role, "", out, astroCoreClient)
+}
+
+func removeWorkspaceUsers(cmd *cobra.Command, args []string, out io.Writer) error {
+	var email string
+
+	// if an email was provided in the args we use it
+	if len(args) > 0 {
+		email = args[0]
```
jemishp, line 184: should this be lower cased and validated as an email before using it?

cmd/cloud/workspace.go
```diff
@@ -52,6 +54,80 @@ func newWorkspaceSwitchCmd(out io.Writer) *cobra.Command {
...
+}
+
+func newWorkspaceUserAddCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add [email]",
+		Short: "Add a user to an Astro Workspace with a specific role",
+		Long: "Add a user to an Astro Workspace with a specific role\n$astro workspace user add [email] --role [WORKSPACE_MEMBER, " +
+			"WORKSPACE_BILLING_ADMIN, WORKSPACE_OWNER].",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return addWorkspaceUsers(cmd, args, out)
```
kushalmalani, line 81: Should these function name be singular..`addWorkspaceUser`, `updateWorkspaceUser`, `removeWorkspaceUser`

cloud/user/user.go
```diff
@@ -225,3 +240,227 @@ func ListOrgUsers(out io.Writer, client astrocore.CoreClient) error {
...
+	if email == "" {
+		user, err := selectUser(users, false)
+		userID = user.Id
+		email = user.Username
+		if err != nil {
+			return err
+		}
+	} else {
+		for i := range users {
+			if users[i].Username == email {
```
jemishp, line 274: Does case matter? Or should we lowercase `users[i].Username` before comparing it to `email`?

cloud/user/user.go
```diff
@@ -225,3 +240,227 @@ func ListOrgUsers(out io.Writer, client astrocore.CoreClient) error {
...
+				userID = users[i].Id
+			}
+		}
+	}
```
jemishp, line 455: Can we refactor this into its own method as it is duplicated in all three `AddWorkspaceUser`,  `UpdateWorkspaceUserRole` & `DeleteWorkspaceUser`?

cmd/cloud/workspace.go
```diff
@@ -71,6 +147,47 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
...
+	return user.ListWorkspaceUsers(out, astroCoreClient, "")
+}
+
+func updateWorkspaceUsers(cmd *cobra.Command, args []string, out io.Writer) error {
+	var email string
+
+	// if an email was provided in the args we use it
+	if len(args) > 0 {
+		email = args[0]
+	}
```
kushalmalani, line 173: Do we need to throw an error for else case. in a scenario where they do not pass email address?

cloud/user/user.go
```diff
@@ -130,27 +131,41 @@ func IsRoleValid(role string) error {
 	return ErrInvalidRole
 }
 
-func selectUser(users []astrocore.User) (astrocore.User, error) {
+func selectUser(users []astrocore.User, workspace bool) (astrocore.User, error) {
+	roleColumn := "ORGANIZATION ROLE"
+	if workspace {
+		roleColumn = "WORKSPACE ROLE"
+	}
```
jemishp, line 138: Do we need this header to be different for Org vs. workspace role when we print a list? What if the header was `ROLE` and the value has `ORGANIZATION_` or `WORKSPACE_` in it.

cmd/cloud/workspace.go
```diff
@@ -71,6 +147,47 @@ func workspaceSwitch(cmd *cobra.Command, out io.Writer, args []string) error {
...
+func removeWorkspaceUsers(cmd *cobra.Command, args []string, out io.Writer) error {
+	var email string
+
+	// if an email was provided in the args we use it
+	if len(args) > 0 {
+		email = args[0]
+	}
+
+	cmd.SilenceUsage = true
+	return user.DeleteWorkspaceUser(email, "", out, astroCoreClient)
```
kushalmalani, line 188: Call this function RemoveWorkspaceUser, since we are not deleting them from the system, but just removing them from a particular workspace

jemishp (changes requested):
Left some suggestions and comments. Looks good otherwise.

One question as I read through this PR: Would a `WORKSPACE_BILLING_ADMIN` be able to run `astro deploy` or `astro deployment` commands? Do we need to change other commands based on a user's Role?

cmd/cloud/workspace.go
sunkickr: It will ask you to pick from a list of users if an email is not provided. Like in deployment commands

cmd/cloud/workspace.go
sunkickr: okay sounds good

cmd/cloud/workspace.go
sunkickr: works for me

cloud/user/user.go
sunkickr: That makes sense

cloud/user/user.go
sunkickr: I thought specifying the type of role would be nice. Especially since we are already changing the table in other ways.

cmd/cloud/workspace.go
sunkickr: It seems that emails are always stored as lower case. Not sure if we need to validate as an email. if someone gives a non email the API will give an error

sunkickr:
@jemishp WORKSPACE_BILLING_ADMIN is actually the wrong role. I'll change it to WORKSPACE_EDITOR. Which commands work with which roles should all be controlled by the RBAC layer in the CLI

cloud/user/user.go
sunkickr: I think we should lowercase email but from what Iv seen everything in users[i].Username is already lower cased. I think the API does this when the user signs up

kushalmalani (approved):

jemishp (approved):
LGTM.

PR #1007, opened 2023-01-06
Expose the deployment of SQL CLI workflows in the Astro CLI

pankajkoti:
## Description

>Implement the command astro flow deploy in the Astro CLI. As a starting point, this command will work on Astro Cloud and it will deploy using the image-based approach.
Also, Identify the Python SDK availability/version compatibility and promote the correct version of Python SDK.
Creates environment specific dags folders `<project-dir>/.deploy/<env>/dags` and deploys DAGs from those directory to Astro Cloud by successively calling the `astro deploy` command post taking care of SQL CLI related deploy implementation.

**WIP**: Unit tests and tidying up the code.

## 🎟 Issue(s)

closes: https://github.com/astronomer/astro-sdk/issues/1473
closes: https://github.com/astronomer/astro-sdk/issues/1520
closes: https://github.com/astronomer/astro-sdk/issues/1602

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


cmd/sql/flow.go
```diff
@@ -317,6 +350,57 @@ func executeCmd(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func executeDeployCmd(cmd *cobra.Command, args []string) error {
+	astroRuntimeVersion, err := getAstroDockerfileRuntimeVersion()
+	if err != nil {
+		return err
+	}
+	SQLCLIVersion, err := sql.GetPypiVersion(sql.AstroSQLCLIProjectURL)
```
tatiana, line 358: Is there any scenario where retrieving the latest version in PyPI would not match the local version of the SQL CLI?

sql/flow_config.go
```diff
@@ -18,6 +18,15 @@ type configResponse struct {
 	BaseDockerImage string `json:"baseDockerImage"`
 }
 
+type compatibilityVersions struct {
+	AstroRuntime   string `json:"astro-runtime"`
+	AstroSdkPython string `json:"astro-sdk-python"`
```
tatiana, line 23: @pankajkoti what do you think if we check both for the lower and upper bound versions?
https://github.com/astronomer/astro-sdk/issues/1473

```
            "astroSDKPythonMaxVersion": "",
            "astroSDKPythonMinVersion": "1.3.1",
            "astroRuntimeMinVersion": "6.1.0",
            "astroRuntimeMaxVersion": "7.1.0"
```

This would be helpful if we need to introduce breaking changes in the future.

cmd/sql/flow.go
```diff
@@ -317,6 +350,57 @@ func executeCmd(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	requiredRuntimeVersion, requiredPythonSDKVersion, err := sql.GetPythonSDKComptability(sql.AstroSQLCLIConfigURL, SQLCLIVersion)
```
tatiana, line 362: To be future-proof, we can have a lower and upper threshold of supported versions, as mentioned in the comment
https://github.com/astronomer/astro-cli/pull/1007/files#r1063343197

cmd/sql/flow.go
```diff
@@ -317,6 +350,57 @@ func executeCmd(cmd *cobra.Command, args []string) error {
...
+	}
+	runtimeVersionMet, err := utils.IsRequiredVersionMet(astroRuntimeVersion, requiredRuntimeVersion)
+	if err != nil {
+		return err
+	}
+
+	if !runtimeVersionMet {
+		pythonSDKPromptContent := promptContent{
+			"Please say yes/no.",
+			fmt.Sprintf("Would you like to add the required version %s of Python SDK dependency to requirements.txt?", requiredPythonSDKVersion),
```
tatiana, line 374: We should be explicit that if users do not add this dependency, the deployment will not proceed.

cmd/sql/flow.go
```diff
@@ -317,6 +350,57 @@ func executeCmd(cmd *cobra.Command, args []string) error {
...
+		if result == "no" {
+			return ErrPythonSDKVersionNotMet
+		}
+		requiredPythonSDKDependency := "\nastro-sdk-python==" + requiredPythonSDKVersion
+		b, err := os.ReadFile(astroRequirementsfilePath)
+		if err != nil {
+			return err
+		}
+		existingRequirements := string(b)
+		if !strings.Contains(existingRequirements, requiredPythonSDKDependency) {
```
tatiana, line 389: What happens if the user has an older version of the Python SDK? Should we prompt again and delete the old version?

cmd/sql/flow.go
pankajkoti: The local version of SQL CLI is always the latest at the moment.

sql/flow_config.go
feluelle: Maybe we can have sth. like `>6.1.0,<7.1.0` syntax as in python. Maybe go-version supports that?

cloud/deploy/deploy.go
```diff
@@ -102,9 +103,8 @@ func getRegistryURL(domain string) string {
 	return registry
 }
 
-func deployDags(path, runtimeID string, client astro.Client) error {
+func deployDags(path, dagsPath, runtimeID string, client astro.Client) error {
```
tatiana, line 106: WDYT of naming this as `dagsFolderPath`?

cmd/cloud/deploy.go
```diff
@@ -64,6 +65,7 @@ func newDeployCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&pytestFile, "test", "t", "", "Location of Pytests or specific Pytest file. All Pytest files must be located in the tests directory")
 	cmd.Flags().StringVarP(&imageName, "image-name", "i", "", "Name of a custom image to deploy")
 	cmd.Flags().BoolVarP(&dags, "dags", "d", false, "Push only DAGs to your Astro Deployment")
+	cmd.Flags().StringVar(&dagsPath, "dags-path", "", "If set deploy dags from this path instead of the dags from working directory")
```
tatiana, line 68: Should we hide this flag?
What happens if users run `astro deploy --help`? Are we displaying it?

cmd/sql/flow.go
```diff
@@ -451,6 +593,39 @@ func runCommand() *cobra.Command {
...
+//nolint:dupl
+func deployCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          deployCmdName,
+		RunE:         executeDeployCmd,
+		SilenceUsage: true,
+	}
+	cmd.Flags().StringVar(&env, "env", "default", "")
+	cmd.Flags().StringVar(&workflowName, "workflow-name", "", "")
+	cmd.Flags().StringVar(&projectDir, "project-dir", ".", "")
```
neel-astro, line 625: please check once the output for `astro flow deploy --help`, I don't think so flow library which handles help output for other commands will work for this one, in which case we would need the default help out with description of these flags being set

go.sum
```diff
@@ -648,6 +649,8 @@ github.com/imdario/mergo v0.3.12/go.mod h1:jmQim1M+e3UYxmgPu/WyfjB3N3VflVyUjjjwH
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/inconshreveable/mousetrap v1.0.1 h1:U3uMjPSQEBMNp1lFxmllqCPM6P5u/Xq7Pgzkat/bFNc=
 github.com/inconshreveable/mousetrap v1.0.1/go.mod h1:vpF70FUmC8bwa3OWnCshd2FqLfsEA9PFc4w1p2J65bw=
+github.com/inconshreveable/mousetrap v1.1.0 h1:wN+x4NVGpMsO7ErUn/mUI3vEoE6Jt13X2s0bqwp9tc8=
+github.com/inconshreveable/mousetrap v1.1.0/go.mod h1:vpF70FUmC8bwa3OWnCshd2FqLfsEA9PFc4w1p2J65bw=
```
neel-astro, line 653: please run `go mod tidy` to clean up the `go.sum` file of unused dependencies

sql/flow.go
```diff
@@ -17,12 +17,12 @@ import (
...
 
 const (
-	astroSQLCLIProjectURL     = "https://pypi.org/pypi/astro-sql-cli/json"
-	astroSQLCLIConfigURL      = "https://raw.githubusercontent.com/astronomer/astro-sdk/astro-cli/sql-cli/config/astro-cli.json"
-	SQLCliDockerfilePath      = ".Dockerfile.sql_cli"
-	SQLCLIDockerfileWriteMode = 0o600
-	SQLCliDockerImageName     = "sql_cli"
-	PythonVersion             = "3.9"
+	AstroSQLCLIProjectURL = "https://pypi.org/pypi/astro-sql-cli/json"
+	AstroSQLCLIConfigURL  = "https://raw.githubusercontent.com/astronomer/astro-sdk/1473-check-python-sdk-version/sql-cli/config/astro-cli.json"
```
neel-astro, line 21: I am guessing this is temporary, leaving a comment for our reference to remove it before merging the PR

cmd/sql/flow.go
```diff
@@ -317,6 +370,94 @@ func executeCmd(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func executeDeployCmd(cmd *cobra.Command, args []string) error {
```
neel-astro, line 373: nit: ideally I would expect most of the logic inside this function to be part of `sql` package instead of `cmd/sql`. In my mind this is how I see the SQL code structure at a high level:
`cmd/sql` -> anything related to command declaration for all the SQL CLI commands
`sql` -> all the business logic outside the `cobra` command declaration for all the SQL CLI commands

sql/flow_config.go
```diff
@@ -79,3 +88,25 @@ func GetBaseDockerImageURI(configURL string) (string, error) {
 
 	return resp.BaseDockerImage, nil
 }
+
+func GetPythonSDKComptability(configURL, sqlCliVersion string) (astroRuntimeVersion, astroSDKPythonVersion string, err error) {
+	httpClient := &http.Client{}
```
neel-astro, line 93: nit: this should go into the global var block, and be reused in all the methods of the package. There is no point in creating a new HTTP client for each individual method.

cmd/sql/flow.go
```diff
@@ -317,6 +370,94 @@ func executeCmd(cmd *cobra.Command, args []string) error {
...
+	}
+
+	astroDeployCmd := cloud.NewDeployCmd()
+	astroDeployArgs := []string{"--dags-path", dagsPath}
+	astroDeployCmd.SetArgs(astroDeployArgs)
+
+	_, err = astroDeployCmd.ExecuteC()
+	if err != nil {
+		return err
+	}
```
neel-astro, line 456: nit: should we be calling `cloud.Deploy` instead of executing the deploy command, I feel that to be more correct implementation in my mind 🤔, will leave it up to you to take a final call

neel-astro (commented):
left bunch of nitpicks, looks fine otherwise

sql/flow_config.go
pankajkoti: Handling as per Felix's suggestions.

cloud/deploy/deploy.go
pankajkoti: keeping it to dagsPath as it being referred to so in the existing code for `astro deploy` and would not like to rename it in this PR.

cmd/sql/flow.go
pankajkoti: Since we append the required version to the end of file, currently it may not be an issue. Let's discuss this over a call together with Felix. Felix and I had a preliminary discussion on this and need to conclude together with you.

cmd/sql/flow.go
pankajkoti: @feluelle and I discussed this and feel that keeping the existing approach is suitable since by initialising the cloud.NewDeployCmd() we have it's default flag values and need not worry about constructing the deploy command otherwise that would need to be pass to cloud.Deploy. And the ExecuteC() method would know what to call it with once we have the cloud.NewDeployCmd() initialised.

cmd/sql/flow.go
tatiana: @pankajkoti I had a discussion with @feluelle last week, which will make this no longer true - shall we update the ticket description for https://github.com/astronomer/astro-sdk/issues/1582, so we improve this in that ticket?

cmd/sql/flow.go
pankajkoti: Log a ticket to enhance this post 0.3

cmd/sql/flow.go
feluelle: Thank you for the heads-up. This https://github.com/astronomer/astro-sdk/pull/1615 should handle that :)

pkg/input/input.go
```diff
@@ -39,3 +40,31 @@ func Password(promptText string) (string, error) {
...
+
+func GetYesNoSelector(pc PromptContent) *promptui.Select {
+	return &promptui.Select{
+		Label: pc.Label,
+		Items: []string{"y", "n"},
+	}
+}
+
+// Gets a y/n confirmation from the user for the given prompt content using the promptui library and returns a boolean accordingly
+func PromptGetConfirmation(runner PromptRunner) (bool, error) {
```
neel-astro, line 61: we should be adding unit tests for these functions since it has been added to `pkg` which is a shared package.

neel-astro (commented):
overall looks fine to me, have left few minor comments

cmd/sql/flow.go
```diff
@@ -66,6 +81,8 @@ func initGlobalCmdArgs() []string {
 }
 
 // Initialize specific cmd args by setting the cmd flags, resolving filepaths and overwriting args
+//
+//nolint:gocognit
```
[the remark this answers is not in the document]
pankajkoti: @feluelle ☝🏽 :)

pkg/input/input.go
pankajkoti: @neel-astro would appreciate you help here, quite unsure how to add unit tests for these components.

sql/flow.go
```diff
@@ -183,3 +190,75 @@ var ExecuteCmdInDocker = func(cmd, mountDirs []string, returnOutput bool) (exitC
...
+	}
+	if !runtimeVersionMet {
+		result, err := input.PromptGetConfirmation(promptRunner)
+		if err != nil {
+			return err
+		}
+		if !result {
+			return ErrPythonSDKVersionNotMet
+		}
+		requiredPythonSDKDependency := "\nastro-sdk-python" + requiredPythonSDKVersion
```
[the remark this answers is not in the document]
pankajkoti: This will be controlled by the compatibility matrix. It may not be always pinned to a specific version. It could also be of the format `>=2.1.0, <3.0.0` which is configured here: https://github.com/astronomer/astro-sdk/blob/astro-cli/sql-cli/config/astro-cli.json

cmd/sql/flow.go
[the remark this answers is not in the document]
feluelle: In my opinion cognitive complexity can't always be measured based on number of lines hence the `nolint`. It is quite straightforward code. Refactoring this in my opinion only adds more boilerplate code.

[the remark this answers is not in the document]
feluelle: @neel-astro I'm open to suggestions for refactoring this.

sql/flow.go
```diff
@@ -183,3 +190,75 @@ var ExecuteCmdInDocker = func(cmd, mountDirs []string, returnOutput bool) (exitC
...
+	defer file.Close()
+
+	scanner := BufIo().NewScanner(file)
+	scanner.Scan()
+	text := scanner.Text()
+	if !strings.Contains(text, runtimeImagePrefix) {
+		return "", ErrNoBaseAstroRuntimeImage
+	}
+
+	runtimeVersion := astroRuntimeVersionRegex.FindStringSubmatch(text)[1]
```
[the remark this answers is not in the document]
pankajkoti: checking that the array has minimum 2 items.

pkg/input/input.go
```diff
@@ -39,3 +40,31 @@ func Password(promptText string) (string, error) {
...
+	_, result, err := runner.Run()
+	if err != nil {
+		return false, err
+	}
+	if result == "y" {
```
kushalmalani, line 66: Does this have to be case sensitive? Also, should there be a default response if user doesnt enter anything. 

Since this a common package, we should allow for those options to be configured

pkg/input/input.go
```diff
@@ -39,3 +40,31 @@ func Password(promptText string) (string, error) {
 	fmt.Print("\n")
 	return string(bytePassword), nil
 }
+
+// Structure to hold content required for displayig prompts required for promptui library functions
```
kushalmalani, line 44: nit: typo change displayig to displaying

kushalmalani (approved):
Dag deploy part of the code looks good to me. Left a comment for the common package (input) function. Will Lean on SQL team folks to review SQL code changes.

pkg/input/input.go
neel-astro:
```Golang
func TestPromptGetConfirmation(t *testing.T) {
	runner := GetYesNoSelector(PromptContent{Label: "test label, enter y/n"})
	runner.Keys = &promptui.SelectKeys{Next: promptui.Key{Code: rune('S')}, Prev: promptui.Key{Code: rune('W')}, PageUp: promptui.Key{Code: rune('D')}, PageDown: promptui.Key{Code: rune('A')}}
	tests := []struct {
		name        string
		inputString string
		want        bool
		wantErr     bool
	}{
		{
			name:        "basic yes case",
			inputString: "\n",
			want:        true,
			wantErr:     false,
		},
		{
			name:        "basic no case",
			inputString: "S\n",
			want:        false,
			wantErr:     false,
		},
		{
			name:        "no input case",
			inputString: "",
			want:        false,
			wantErr:     true,
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			runner.Stdin = io.NopCloser(strings.NewReader(tt.inputString))
			got, err := PromptGetConfirmation(runner)
			if (err != nil) != tt.wantErr {
				t.Errorf("PromptGetConfirmation() error = %v, wantErr %v", err, tt.wantErr)
				return
			}
			if got != tt.want {
				t.Errorf("PromptGetConfirmation() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

cmd/sql/flow.go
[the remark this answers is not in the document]
feluelle: Thank you. I have pushed the changes. PTAL again.

pkg/input/input.go
feluelle: Added in b3bd5e2

pkg/input/input.go
feluelle:
```suggestion
// Structure to hold content required for displaying prompts required for promptui library functions
```

pkg/input/input.go
feluelle: That's right, but we tried to be consistent with `Confirm`:
```golang
// Confirm requests a user to confirm their input
func Confirm(promptText string) (bool, error) {
	reader := bufio.NewReader(os.Stdin)
	fmt.Printf("%s (y/n) ", promptText)

	text, _ := reader.ReadString('\n')
	return strings.Trim(text, "\r\n") == "y", nil
}
```

I would say we can improve that separately sometime - and consistently for all inputs.

feluelle: We are also planning to add a non-prompt way separately for the prompt in our flow deploy command.

neel-astro (approved):

PR #1017, opened 2023-01-13
wq update command works with KE

jemishp:
## Description

> This PR enables the `astro deployment worker-queue update` command to create correct queues based on the executor for the existing deployment being requested to update. `KubernetesExecutor` only supports one default queue and so if a user changes their worker type, we request the existing deployment to be updated with a new default queue and the resources be re-calculated based on the new worker-type.

## 🎟 Issue(s)

Related #931

## 🧪 Functional Testing

> update a default queue for a KE deployment
```bash
$ astro deployment worker-queue update --deployment-name jp-KE-test -t m5.xlarge -f

 #     WORKER QUEUE     ISDEFAULT     ID
 1     default          true          clcthey1d08452zvz1xop6bd3

> 1
worker queue default for jp-KE-test in cl0v1p6lc728255byzyfs7lw21 workspace updated
```
> update can be canceled
```bash
$ astro deployment worker-queue update --deployment-name jp-KE-test -n default -t m5.xlarge

Are you sure you want to update the default worker queue? If there are any tasks in your DAGs assigned to this worker queue, the tasks might get stuck in a queued state and fail to execute (y/n) n
Canceling worker queue update
```
> error when input is not valid 
```bash
jpatel@jemishs-mbp astro-cli % ./astro deployment worker-queue update --deployment-name jp-KE-test -n another-queue -t m5.xlarge
Error: KubernetesExecutor does not support a non default worker queue in the request. Rename the queue to default
```
## 📋 Checklist

### DIFF
--- a/cloud/deployment/fromfile/fromfile.go
+++ b/cloud/deployment/fromfile/fromfile.go
@@ -198,8 +198,8 @@ func getCreateOrUpdateInput(deploymentFromFile *inspect.FormattedDeployment, clu
 			// check if more than one queue is requested
 			if len(listQueues) > 1 {
 				return astro.CreateDeploymentInput{}, astro.UpdateDeploymentInput{},
-					fmt.Errorf("%w more than one worker queue. (%d) were requested",
-						workerqueue.ErrNotSupported, len(listQueues))
+					fmt.Errorf("%s %w more than one worker queue. (%d) were requested",
+						deployment.KubeExecutor, workerqueue.ErrNotSupported, len(listQueues))
 			}
 			for i := range listQueues {
 				err = workerqueue.IsKubernetesWorkerQueueInputValid(&listQueues[i])
--- a/cloud/deployment/workerqueue/workerqueue.go
+++ b/cloud/deployment/workerqueue/workerqueue.go
@@ -109,7 +109,7 @@ func CreateOrUpdate(ws, deploymentID, deploymentName, name, action, workerType s
 				}
 			}
 			// user requested an update and queueToCreateOrUpdate exists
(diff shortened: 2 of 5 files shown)


cloud/deployment/fromfile/fromfile.go
```diff
@@ -198,8 +198,8 @@ func getCreateOrUpdateInput(deploymentFromFile *inspect.FormattedDeployment, clu
 			// check if more than one queue is requested
 			if len(listQueues) > 1 {
 				return astro.CreateDeploymentInput{}, astro.UpdateDeploymentInput{},
-					fmt.Errorf("%w more than one worker queue. (%d) were requested",
-						workerqueue.ErrNotSupported, len(listQueues))
+					fmt.Errorf("%s %w more than one worker queue. (%d) were requested",
+						"KubernetesExecutor", workerqueue.ErrNotSupported, len(listQueues))
```
kushalmalani, line 202: You can use `deployment.KubeExecutor` instead of defining it as string

kushalmalani: There are few more such occurence in this PR. Can you update the same?

Thinking forward, we should have a constant file something like `airflow/constants.go` so we could reference there instead of importing and referencing across different packages. Doesnt need to be addressed in this PR, but eventually we should be more consistent so the code is more readable

neel-astro (approved):
LGTM apart from Kushal's comment

kushalmalani (approved):

cloud/deployment/fromfile/fromfile.go
jemishp: @kushalmalani changed all the strings to use constants.

PR #1033, opened 2023-01-20
Persist deployment and workspace IDs for flow command

feluelle:
## Description

This PR fixes the following issues:

* whitespace issues in flow dockerfile and deploy cmd
* reset of deployment config values
* e2e tests to run on latest version

## 📋 Checklist

### DIFF
--- a/cmd/sql/e2e/flow_test.go
+++ b/cmd/sql/e2e/flow_test.go
@@ -33,8 +33,9 @@ func chdir(t *testing.T, dir string) func() {
 }
 
 func execFlowCmd(args ...string) error {
+	// Set the astro-cli version to a specific version we want to test for
+	version.CurrVersion = "1.10.0"
 	cmd := sql.NewFlowCommand()
-	version.CurrVersion = "1.8"
 	cmd.SetArgs(args)
 	_, err := cmd.ExecuteC()
 	return err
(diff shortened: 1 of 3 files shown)


tatiana (approved):

cmd/sql/e2e/flow_test.go
```diff
@@ -33,8 +33,9 @@ func chdir(t *testing.T, dir string) func() {
 }
 
 func execFlowCmd(args ...string) error {
+	// Set the astro-cli version to a version compatible with the latest sql-cli version for e2e tests
+	version.CurrVersion = "1.10.0"
```
neel-astro, line 37: nit: is there a way to not need to update this on all releases? isn't it always equal to the latest CLI release version as per semver?

neel-astro (approved):

cmd/sql/e2e/flow_test.go
feluelle: Yes, there is a way, but currently, our config is allowing specific versions to be used for sql-cli versions. See [here](https://github.com/astronomer/astro-sdk/blob/astro-cli/sql-cli/config/astro-cli.json).

What we probably want is simply to have an entry for `latest` where we don't set any upper bound. Currently, the `1.10.0` entry is such an entry, but we might have to upper-bound it in the future.

We have the same issue if we are on a branch. The airflow version is a `SNAPSHOT-*`. We probably want to skip the compatibility matrix in these cases, maybe also for e2e tests? WDYT?

neel-astro: hmmm, anything with `SNAPSHOT-*` can be treated as the latest CLI version from SDK POV

feluelle: Sorry, I meant the astro-cli version - not airflow version.

PR #1035, opened 2023-01-23
Use installed flow version to get compatibiltiies from matrix

pankajkoti:
## Description

> Currently, we get the constraint of the SQL CLI python package that is used for installing flow and use the same constraint as a key to match in the compatibility matrix. However the compatibility matrix is meant to contain specific Pypi versions as key and not the constraints itself. We therefore need to get the installed flow version and use that flow version to find the compatibilities. For this reason, we call the `flow version` command which gives us the installed version of flow and we use the same to match it in the matrix.

## 🎟 Issue(s)

closes: https://github.com/astronomer/astro-sdk/issues/1642

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/sql/flow.go
+++ b/cmd/sql/flow.go
@@ -29,15 +29,15 @@ var (
 
 // All cmd names
 const (
-	flowCmdName            = "flow"
-	aboutCmdName           = "about"
-	configCmdName          = "config"
-	generateCmdName        = "generate"
-	initCmdName            = "init"
-	runCmdName             = "run"
-	validateCmdName        = "validate"
-	versionCmdName         = "version"
-	deployCmdName          = "deploy"
+	flowCmdName     = "flow"
+	aboutCmdName    = "about"
+	configCmdName   = "config"
+	generateCmdName = "generate"
+	initCmdName     = "init"
+	runCmdName      = "run"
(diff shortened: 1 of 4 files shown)


cmd/sql/flow.go
```diff
@@ -24,19 +24,21 @@ import (
...
+	flowCmdName            = "flow"
+	aboutCmdName           = "about"
+	configCmdName          = "config"
+	generateCmdName        = "generate"
+	initCmdName            = "init"
+	runCmdName             = "run"
+	validateCmdName        = "validate"
+	versionCmdName         = "version"
+	deployCmdName          = "deploy"
+	versionCmdOutputPrefix = "Astro SQL CLI"
```
feluelle, line 41: Can we move that to its own `const` section?

feluelle (approved):
LGTM

cmd/sql/flow.go
pankajkoti, line 41:
```suggestion
```

pankajkoti: This is no longer needed. We're now reading the json response.

kushalmalani (approved):

PR #1052, opened 2023-01-30
add new feature flag to disable creation of duplicate volumes

rob-1126:
Older versions of podman (< 4.x) fail to handle volumes defined as VOLUMEs in our Dockerfile and then re-defined again in our compose file appropriately causing astro dev start to fail.

Both Docker and newer versions of podman tolerate this syntax well but this feature-flag offers a method for Podman 3.x to use astro dev start by setting duplicate_volumes: false in their ~/astro/config.yaml.

Notably, Ubuntu LTS currently ships Podman 3.x.

## Description

Enable WSL + Ubuntu 2 22.04 LTS users to use astro dev start an appropriately configured astro-cli intances.

## 🧪 Functional Testing

Tested under wsl2+podman 3.x using the Ubuntu 22.04 LTS stock Podman packages and setting duplicate_values: false in ~/astro/config.yaml. Ran astro dev init. Ran astro dev start. Logged in through the web-browser.

## 📸 Screenshots

image

## 📋 Checklist

- [ X] Rebased from the main (or release if patching) branch (before testing)
- [ X] Ran `make test` (FAILS WITH codecov)
- [ X ] Ran `make lint` before taking out of draft


sunkickr (approved):
LGTM

airflow/include/composeyml.yml
```diff
@@ -33,7 +33,11 @@ services:
     ports:
       - 127.0.0.1:{{ .PostgresPort }}:5432
     volumes:
+      {{if .DuplicateImageVolumes}}
```
neel-astro, line 36: trying to understand this change, because the actual volume definition is slightly above: https://github.com/astronomer/astro-cli/pull/1052/files#diff-4e4b970bc22596b97b95d621fee0b2293a0df07fa231494d2f54da904dd8be05R18-R22, and this is just us referring that volume, so trying to figure out where is the duplicate definition?

rob-1126:
The first one definition is implicit - created from the VOLUME lines
present in the Dockerfile and the second is explicit - defined in the
compose file. My guess is that since its not referenced by any containers
that the volume settings associated with the volume that remain in the file
also effectively doesn't exist. The bug this feature-flag works around is
no longer present in current versions of podman so I didn't dig too much
deeper.

neel-astro (approved):
seemed like an odd thing to fail on Podman side, but functionally LGTM 👍. Thanks for all the effort in adding this compatibility to CLI

Issue #1072, opened 2023-02-08
Can't update a deployment or deploy with API keys without astro login


magdagultekin:
**Describe the bug**

I can't update a deployment or deploy using Deployment API keys when I'm not logged in.

```
(venv) (base) magdagultekin@Magdalenas-MacBook-Pro astro-project % astro deployment update magda-tmp --deployment-file deployment-config.yml

Using an Astro API key
Get "/organizations": unsupported protocol scheme ""

There was an error using API keys, using regular auth instead
Error: no context set, have you authenticated to Astro or Astronomer Software? Run astro login and try again```
```

**What CLI Version did you experience this bug?**

1.10.0

**This CLI bug is related to which Astronomer Platform?**

- [x] Astro
- [ ] Software
- [ ] None/Unknown

**What Operating System is the above CLI installed on?** macOs

**🪜 Steps To Reproduce**

1. Go to cloud.astronomer.io and create a Deployment
2. Create API keys
3. Set temporary OS-level env vars with:
```
export ASTRONOMER_KEY_ID=<your-key-id>
export ASTRONOMER_KEY_SECRET=<your-key-secret>
```
4. Run `astro logout`
5. Try to deploy with `astro deploy <deployment-id>`
6. See the error

**📸 Error**
```
(venv) (base) magdagultekin@Magdalenas-MacBook-Pro astro-project % astro deploy magda-tmp
Using an Astro API key
Get "/organizations": unsupported protocol scheme ""

There was an error using API keys, using regular auth instead
Error: no context set, have you authenticated to Astro or Astronomer Software? Run astro login and try again
Usage:
  astro deploy DEPLOYMENT-ID [flags]

Examples:

Specify the ID of the Deployment on Astronomer you would like to deploy this project to:

  $ astro deploy <deployment ID>

Menu will be presented if you do not specify a deployment ID:

  $ astro deploy


Flags:
  -d, --dags                     Push only DAGs to your Astro Deployment
  -n, --deployment-name string   Name of the deployment to deploy to
  -e, --env string               Location of file containing environment variables for Pytests (default ".env")
  -f, --force                    Force deploy even if project contains errors or uncommitted changes
  -h, --help                     help for deploy
  -i, --image-name string        Name of a custom image to deploy
      --parse                    Succeed only if all DAGs in your Astro project parse without errors
  -p, --prompt                   Force prompt to choose target deployment
      --pytest                   Deploy code to Astro only if the specified Pytests are passed
[...]

PR #1073, opened 2023-02-08
Fix API key context issue

sunkickr:
## Description

Fix API key issue. if user is not in any context make sure that they switch to astronomer.io by default when using API keys

## 🎟 Issue(s)

- https://github.com/astronomer/astro-cli/issues/1072

## 🧪 Functional Testing

manual testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/cloud/setup.go
+++ b/cmd/cloud/setup.go
@@ -213,13 +213,14 @@ func checkAPIKeys(astroClient astro.Client, coreClient astrocore.CoreClient, arg
 			if err != nil {
 				return false, err
 			}
-
-			// Switch context
-			err = context.Switch(domain)
-			if err != nil {
-				return false, err
-			}
 		}
+
+		// Switch context
+		err = context.Switch(domain)
+		if err != nil {
+			return false, err
+		}
+
 		c, err = context.GetContext(domain) // get current context
 		if err != nil {
 			return false, err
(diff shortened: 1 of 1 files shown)


cmd/cloud/setup.go
```diff
@@ -213,13 +213,14 @@ func checkAPIKeys(astroClient astro.Client, coreClient astrocore.CoreClient, arg
...
 			}
+		}
 
-			// Switch context
-			err = context.Switch(domain)
-			if err != nil {
-				return false, err
-			}
+		// Switch context
+		err = context.Switch(domain)
```
neel-astro, line 219: nit: I think the `Switch`function in itself also does the exact same thing as 
```
if !context.Exists(domain) {
			err := context.SetContext(domain)
			if err != nil {
				return false, err
			}
		}
```

so in theory, having the logic simply like this should give the same result:
```
        domain := "astronomer.io"
        err = context.Switch(domain)
        if err != nil {
                return false, nil
        }
```

neel-astro (approved):
looks fine, have left a nit to look at

cmd/cloud/setup.go
sunkickr: we only switch to astronomer.io if context doesn't exist because the user might be using astronomer-dev.io. We only want to switch to astronomer.io if the user has no context

PR #1092, opened 2023-02-23
Add error check for if there is a version mismatch

dimberman:
## Description

Currently, if a user does not have a matching version of the SQL CLI to our version map, they get the following error:

astro flow deploy
Error: Malformed constraint:

This does not help much in determining why the deployment is failing.

This PR adds better context so that future CREs can better aide customers.
## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/sql/flow_config.go
+++ b/sql/flow_config.go
@@ -119,5 +119,8 @@ func GetPythonSDKComptability(configURL, sqlCliVersion string) (astroRuntimeVers
 	}
 
 	SQLCLICompatibilityVersions := resp.Compatibility[sqlCliVersion]
+	if SQLCLICompatibilityVersions.AstroRuntime == "" {
+		return "", "", fmt.Errorf("could not find a matching SQL CLI compatibility version: current version %s version map: %v  %w", sqlCliVersion, resp.Compatibility, err)
+	}
 	return SQLCLICompatibilityVersions.AstroRuntime, SQLCLICompatibilityVersions.AstroSdkPython, nil
 }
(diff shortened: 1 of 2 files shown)


sql/flow_config.go
```diff
@@ -119,5 +119,8 @@ func GetPythonSDKComptability(configURL, sqlCliVersion string) (astroRuntimeVers
 	}
 
 	SQLCLICompatibilityVersions := resp.Compatibility[sqlCliVersion]
+	if SQLCLICompatibilityVersions.AstroRuntime == "" {
+		return "", "", fmt.Errorf("could not find a matching SQL CLI compatibility version: current version %s version map: %v  %w", sqlCliVersion, resp.Compatibility, err)
+	}
 	return SQLCLICompatibilityVersions.AstroRuntime, SQLCLICompatibilityVersions.AstroSdkPython, nil
```
tatiana, line 125: What do you think if we followed the same strategy we have at the moment for the `astroCliCompatiblity` section?

It feels like both are very similar, but we are dealing with them in different ways. Do we have exact matches for  `compatibility`? It would be great if we could normalise the behaviour of both of them, probably having a `default` section in `compatibility` as well

pankajkoti: I am happy with both approaches. Since we control the releases for SQL CLI, we could make sure that we add an entry for the release in the compatibility matrix. If not, do we know what default value to specify for the runtime image and python-sdk version irrespective of the version for SQL-CLI?

sunkickr (approved):

PR #1093, opened 2023-02-24
Update dagintegritytestdefault.py

dylanbstorey:
## Description

>Updates the os.getenv monkey patch to manage people who are setting defaults with args instead of kwargs

## 🎟 Issue(s)

Its come up with CRE/Field where customers are using the valid but less explicit `os.getenv(key,default)` in their code and our monkey patch wasn't honoring their default. This SHOULD address both `os.getenv(key,default)` and `os.getenv(key,default=value` patterns that are valid. 

## 🧪 Functional Testing

put a few `os.getenv()` statements in the top of a dag and did some echo debugging

Screenshot 2023-02-24 094746

## 📸 Screenshots

Screenshot 2023-02-24 091721

## 📋 Checklist

### DIFF
--- a/airflow/include/dagintegritytestdefault.py
+++ b/airflow/include/dagintegritytestdefault.py
@@ -25,17 +25,25 @@ BaseHook.get_connection = basehook_get_connection_monkeypatch
 
 
 # =========== MONKEYPATCH OS.GETENV() ===========
-def os_getenv_monkeypatch(key: str, *args, default=None, **kwargs):
-    print(
-        f"Attempted to fetch os environment variable during parse, returning a mocked value for {key}"
-    )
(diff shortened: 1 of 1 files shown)


collinmcnulty:
Why is there a None between x and that in the echo?

dylanbstorey:
the first prints occur when the module is imported , but before the mocks are applied to the DagBag import

dylanbstorey:
2 reasons - 

screen shot of the code wasn't correct. and the initial prints occur when the module is loaded by pytest. the next 2 are the 2 dagbag imports that occur and mocks are applied

ReadytoRocc (approved):
LGTM

sunkickr (approved):
I tested this out manually and it works fine

Issue #720, opened 2022-09-09
Import astronomer registry dags through astro cli


odaneau-astro:
- [x] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**
No a problem per se, more of an idea/suggestion.

**🧩 Describe the solution you'd like**

When creating a new project using the `astro cli init`, two simple examples are included by default. Those examples are ok to understand core concepts of Airflow, but do not help users deliver business value right away.

The Astronomer registry is a fantastic tool to help deliver business value quickly. It offers a wide variety of Dags to jumpstart common use cases such as running astro sdk ETLs, triggering DBT models, loading csv files to Snowflake, executing pipelines in Databricks, etc.

Having the registry Dags accessible through the CLI would allow beginners to start authoring with examples they can relate to and "fill the blanks" to deliver business value faster.

The value-add behind this is to provide a more "batteries included" experience and to accelerate "time-to-business-value" for users.

There are few different ways to go about how this could work:
1. Add an option in the `astro dev init` command to ask users if they want to add examples from the registry.
2. Add a command line argument to the `astro dev init --registry-url=https://registry.astronomer.io/dags/databricks-ml-example`
3. Add a new import command such as `astro import dag --registry-url=https://registry.astronomer.io/dags/databricks-ml-example`
   -> This way we could also import providers, modules, etc.

**🤔 Describe alternatives you've considered**

The current alternative is to browse through the Astronomer registry, search the sample that fits your needs, clone the repo locally, then copy-paste the dag code into the `dags` folder of the astro cli project.

Not only this takes more time and effort, newcomers don't always know the registry even exists.

I'm really open to feedback on this and curious to see if others see any potential value here.

cmarteepants:
@paolaperaza @sunkickr @virajmparekh -- this is a really cool idea. Do we have this in productboard?

sunkickr:
I think this is a really cool idea. We could potentially have the CLI query the registry DB directly. With something like `astro registry-dags list <search-term>` and then select a dag to import

odaneau-astro:
Would be nice to include Airflow's example DAGs as well. 
For example. [Google's dags here](https://github.com/apache/airflow/tree/main/airflow/providers/google/cloud/example_dags)

josh-fell:
[...]

PR #1148, opened 2023-04-11
New Registry Commands

fritz-astronomer:
## Description

- `astro registry dag add dbt-basic --version 1.0.4`
    - add a DAG file (just the `.py`) directly from the registry, at the filepath defined by the registry, optionally add any providers.
- `astro registry provider add snowflake`
    - add the provider to `requirements.txt` 
- ~`astro registry dev init https://github.com/astronomer/my-project/`~
    - ~init a project from the registry~

## 🧪 Functional Testing  + 📸 Screenshots

https://user-images.githubusercontent.com/80706212/231062658-9a2142c5-829e-4690-8ad7-80245b8510bd.mov

issue:

- https://github.com/astronomer/astro-cli/issues/720

## 📋 Checklist


fritz-astronomer:
First few commands complete. Want to add a few more quality-of-life things, and `init`

cmd/registry/root.go
```diff
@@ -0,0 +1,261 @@
...
+		providerString := fmt.Sprintf("%s==%s", name, version)
+		log.Infof("Writing %s to %s", providerString, filename)
+		_, err = f.WriteString(providerString + "\n")
+		logFatal(err)
+		defer f.Close()
+	}
+}
+
+func downloadDag(dagId string, dagVersion string) {
+	// TODO - DAG Version Optional
```
iancmoritz, line 101: for DAG version you're on the right track - you can just default to the keyword "latest" if a version is not provided

cmd/registry/root.go
```diff
@@ -0,0 +1,261 @@
...
+	cmd.Flags().StringVar(&dagVersion, "version", "", "Optional DAG Version to Download.")
+	return cmd
+}
+
+func newRegistryAddProviderCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "provider [PROVIDER]",
+		//Aliases: []string{"p"},
+		Short: "Utilize the Astronomer Registry to add a Provider",
+		Long:  "Utilize the Astronomer Registry to download a Provider like Snowflake to your project's dependencies.",
```
iancmoritz, line 221: Perhaps reword the "like Snowflake" into (like Snowflake or Great Expectations)... super open to other suggestions!

cmd/registry/root.go
```diff
@@ -0,0 +1,261 @@
...
+	getProviderUrl := url.URL{
+		Scheme:   "https",
+		Host:     registryHost,
+		Path:     fmt.Sprintf("%s/%s", registryApi, providerSearchRoute),
+		RawQuery: fmt.Sprintf(providerSearchQuery, url.QueryEscape(providerName)),
+	}
+	return getProviderUrl.String()
+}
+
+func getProviderRoute(providerId string, providerVersion string) string {
```
iancmoritz, line 31:
```suggestion
func getProviderRoute(providerName string, providerVersion string) string {
```

iancmoritz (changes requested):
Fritz - this is fantastic. Super small changes overall but thank you!

cmd/registry/root.go
fritz-astronomer: Nice! That is way easier than pinging the API and getting the first one, which I was about to implement 😅

cmd/registry/root.go
odaneau-astro: The airflow schema refers to _apache-airflow-provider-snowflake_ as a "provider package name" and "amazon" would be the "provider name". see here: https://github.com/apache/airflow/blob/main/airflow/provider_info.schema.json

fritz-astronomer:
Feedback notes:
- Linting, add test coverage
- command defs then functions
- `cmd.cloud.deployment_test`, `cloud.deployment_test.go`
- `pkg.util.go`
- `astro registry <dag|provider> <add|list>`

cmd/registry/root.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	cmd.Flags().StringVar(&dagVersion, "version", "latest", "Optional DAG Version to Download.")
+	return cmd
+}
+
+func newRegistryAddProviderCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "provider [PROVIDER]",
+		//Aliases: []string{"p"},
+		Short: "Utilize the Astronomer Registry to add a Provider",
```
jwitz, line 218:
```suggestion
		Short: "Download a provider package from the Astronomer Registry",
```
For my understanding: how do we anticipate someone to use this command? How is this easier than just pasting into `requirements.txt`?

cmd/registry/root.go
```diff
@@ -0,0 +1,259 @@
...
+		Long:  "Utilize the Astronomer Registry to download a Provider like Snowflake or Great Expectations to your project's dependencies.",
```
jwitz, line 219:
```suggestion
		Long:  "Download a provider package as an Astro project dependency.",
```
I think the example here is confusing, as a user I might think I just need to enter "Snowflake" or "Great Expectations" as an input

cmd/registry/root.go
```diff
@@ -0,0 +1,259 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			if len(args) > 0 {
+				providerName := args[0]
+				addProviderByName(providerName)
+			} else {
+				providerName := input.Text("Enter Provider ID to download: ")
```
jwitz, line 225:
```suggestion
				providerName := input.Text("Enter the name of the provider package to download: ")
```

cmd/registry/root.go
fritz-astronomer: @jwitz - Plenty of new to airflow folks. I think it's *absolutely* easier to type `astro registry provider add snowflake` than go look up what the correct import and version is, know what `requirements.txt` is, add it, version it - etc. ALSO this lets you `astro registry dag add ...` and have it auto-add the correct provider 😎

fritz-astronomer:
Future work:
- it'd be awesome if the registry had example dags for each provider
- it'd be awesome if the registry had package.txt additions

pkg/util/util.go
```diff
@@ -14,6 +14,7 @@ import (
 	goVersion "github.com/hashicorp/go-version"
 )
 
+// UTILS CAN GO HERE
```
neel-astro, line 17: nit: code cleanup
```suggestion
```

pkg/httputil/http.go
```diff
@@ -104,3 +113,48 @@ func newError(resp *http.Response) *Error {
...
+		_ = file.Close()
+	}(file)
+	logrus.Infof("Downloaded %s from %s", path, sourceURL)
+}
+
+func RequestAndGetJSONBody(route string) map[string]interface{} {
+	res, err := http.Get(route) //nolint
+	defer func(Body io.ReadCloser) {
+		_ = Body.Close()
+	}(res.Body)
```
neel-astro, line 146:
```suggestion
	defer res.Body.Close()
```
nit: simplifying the code

cmd/registry/dag.go
```diff
@@ -0,0 +1,100 @@
...
+	return cmd
+}
+
+func downloadDag(dagID, dagVersion string, addProviders bool, out io.Writer) {
+	// https://api.astronomer.io/registryV2/v1alpha1/organizations/public/dags?sorts=displayName%3Aasc&limit=1&query=foo
+	filledDagRoute := getDagRoute(dagID, dagVersion)
+	dagJSON := httputil.RequestAndGetJSONBody(filledDagRoute)
+
+	githubRawSourceURL, exists := dagJSON["githubRawSourceUrl"].(string)
+	printutil.LogKeyNotExists(exists, "githubRawSourceUrl", dagJSON)
```
neel-astro, line 71: nit: I think it make sense just to log things directly here that way it makes things more readable and straightforward, as of now while reading the code I have to switch to another package/file to figure out that it's just a log
```suggestion
	if !exists {
	        log.Fatalf("Couldn't find key githubRawSourceUrl in Response! %v", dagJSON)
	}
```

cmd/registry/provider.go
```diff
@@ -0,0 +1,125 @@
...
+	addProviderToRequirementsTxt(name, version, out)
+}
+
+func addProviderToRequirementsTxt(name, version string, out io.Writer) {
+	const filename = "requirements.txt"
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY|os.O_CREATE, writeAndReadPermissions)
+	defer func(f *os.File) {
+		_ = f.Close()
+	}(f)
+	printutil.LogFatal(err)
```
neel-astro, line 110: same here, it make sense to log the error here instead of having a separate function in another package/file
```suggestion
        if err != nil {
		log.Fatal(err)
	}
```

cmd/registry/provider.go
```diff
@@ -0,0 +1,125 @@
...
+
+	addProviderToRequirementsTxt(name, version, out)
+}
+
+func addProviderToRequirementsTxt(name, version string, out io.Writer) {
+	const filename = "requirements.txt"
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY|os.O_CREATE, writeAndReadPermissions)
+	defer func(f *os.File) {
+		_ = f.Close()
+	}(f)
```
neel-astro, line 109:
```suggestion
	defer f.Close()
```
nit: simplify code

cmd/registry/provider_test.go
```diff
@@ -0,0 +1,48 @@
...
+	cmd.SetArgs(args)
+	testUtil.SetupOSArgsForGinkgo()
+	_, err := cmd.ExecuteC()
+	return buf.String(), err
+}
+
+func TestProviderAdd(t *testing.T) {
+	testUtil.InitTestConfig(testUtil.CloudPlatform)
+
+	_ = os.Remove("requirements.txt")
```
neel-astro, line 31:
```suggestion
	defer os.Remove("requirements.txt")
```
nit: simplify code so that we don't need multiple `os.Remove`

pkg/httputil/http.go
```diff
@@ -104,3 +113,48 @@ func newError(resp *http.Response) *Error {
...
+		_ = Body.Close()
+	}(resp.Body)
+	printutil.LogFatal(err)
+
+	err = fileutil.WriteToFile(path, resp.Body)
+	printutil.LogFatal(err)
+
+	defer func(file *os.File) {
+		_ = file.Close()
+	}(file)
```
neel-astro, line 138:
```suggestion
	defer file.Close()
```
nit: simplify code

pkg/httputil/http.go
```diff
@@ -104,3 +113,48 @@ func newError(resp *http.Response) *Error {
...
+	client := http.Client{
+		CheckRedirect: func(r *http.Request, via []*http.Request) error {
+			r.URL.Opaque = r.URL.Path
+			return nil
+		},
+	}
+	resp, err := client.Get(sourceURL) //nolint
+	defer func(Body io.ReadCloser) {
+		_ = Body.Close()
+	}(resp.Body)
```
neel-astro, line 130:
```suggestion
	defer resp.Body.Close()
```
nit: same here, simplify the code

neel-astro (commented):
left some minor comments around simplifying the logic to improve readability, LGTM otherwise

cmd/registry/provider.go
fritz-astronomer: danke. I think that was a "pycharm doesn't like this" Special ™️

pkg/util/util.go
fritz-astronomer: my bad - reminder to myself during a convo with @sunkickr that snuck in :/ didn't intend to commit

cmd/registry/provider.go
fritz-astronomer: Yea - here's what it didn't like 
image

cmd/registry/dag_test.go
```diff
@@ -0,0 +1,43 @@
...
+	cmdArgs := []string{"add", "sagemaker-batch-inference"}
+	resp, err := execDagCmd(cmdArgs...)
+	assert.NoError(t, err)
+	assert.Contains(t, resp, "dags/sagemaker-batch-inference.py", "we mention where we downloaded it in stdout")
+
+	fileContents, _ := fileutil.ReadFileToString("dags/sagemaker-batch-inference.py")
+	assert.Contains(t, fileContents, "airflow", "The DAG has words in it")
+	assert.Contains(t, fileContents, "DAG", "The DAG has words in it")
+
+	_ = os.Remove("dags/sagemaker-batch-inference.py")
```
neel-astro, line 42: nit: this is not required now with the defer
```suggestion
```

cmd/registry/provider.go
```diff
@@ -0,0 +1,140 @@
...
+	addProviderToRequirementsTxt(name, version, out)
+}
+
+func addProviderToRequirementsTxt(name, version string, out io.Writer) {
+	const filename = "requirements.txt"
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY|os.O_CREATE, writeAndReadPermissions)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
neel-astro, line 121:
```suggestion
	defer f.Close()
```
nit: file close should be part of defer so that it is called from any return point in the function. Sorry I might have missed it in my earlier review

cmd/registry/provider.go
```diff
@@ -0,0 +1,140 @@
...
+	} else {
+		log.Debugf("Couldn't find %s already in %s", name, string(b))
+		providerString := fmt.Sprintf("%s==%s", name, version)
+		_, err = f.WriteString(providerString + "\n")
+		if err != nil {
+			log.Fatal(err)
+		}
+		_, _ = fmt.Fprintf(out, "\nWrote %s to %s", providerString, filename)
+	}
+	_ = f.Close()
```
neel-astro, line 139:
```suggestion
```
nit: not needed once we have it as defer

cmd/registry/provider_test.go
```diff
@@ -0,0 +1,48 @@
...
+	fileContents, _ := fileutil.ReadFileToString("requirements.txt")
+	assert.Regexp(t, `apache-airflow-providers-snowflake==\d+\.\d+\.\d+\n$`, fileContents, "We added the provider to the file")
+
+	_, err = execProviderCmd(cmdArgs...)
+	assert.NoError(t, err)
+	// TODO - assert against stdout "apache-airflow-providers-snowflake already exists in requirements.txt"
+	fileContents, _ = fileutil.ReadFileToString("requirements.txt")
+	assert.Regexp(t, `apache-airflow-providers-snowflake==\d+\.\d+\.\d+\n$`, fileContents, "We didn't write it again")
+
+	_ = os.Remove("requirements.txt")
```
neel-astro, line 47:
```suggestion
```
nit: not needed since we have it as defer already

neel-astro (approved):
Great work and thanks for the contribution, this would be quite useful for users. Code LGTM, has left some final nits to squash.

cmd/registry/dag_test.go
fritz-astronomer: -_- good catch, thanks.

cmd/registry/dag.go
```diff
@@ -0,0 +1,111 @@
...
+			addProviderByName(providerID, out)
+		}
+	}
+}
+
+func getDagRoute(dagID, dagVersion string) string {
+	filledDagRoute := fmt.Sprintf(dagRoute, dagID, dagVersion)
+	getDagURL := url.URL{
+		Scheme: "https",
+		Host:   registryHost,
```
kushalmalani, line 107: Would this work only on prod env?

fritz-astronomer:
@kushalmalani - they don't actually need to. They just add an entry to a `requirements.txt` or download a file to a `dags` folder. I think in both instances the folder or file is created if it doesn't exist.

cmd/registry/provider.go
neel-astro: ideally, we shouldn't have log.Fatal(), but rather have it as log.Error() and then return the error from the function, and then the caller should handle the error and act accordingly.

The linter is complaining because as of now f.Close() even if in defer won't be called if there is any log.Fatal() after that in the code execution path. It's not a big issue if we are just missing out on f.Close(), but ideally in Golang we don't exit abruptly and handle errors everywhere, which makes it simpler to maintain in the future.

kushalmalani (approved):

cmd/registry/dag.go
fritz-astronomer: didn't take a dev registry into account yet

cmd/registry/provider.go
fritz-astronomer: that makes sense. I'll see if I can address that

sunkickr (approved):

PR #1185, opened 2023-05-08
Bulk convert to using testify/suite for tests

ashb:
This is in some regards a "change for no purpose" but it lets us more
easily add setup or teardown to tests.

The `go test` compatible test was named after the Go packge
`./pkg/workspace` becomes `TestPkgWorkspaceSuite`.

This change was almost entirely automatically applied using
`github.com/uber-go/gopatch`

I have not converted 100% of tests to use suites, some of them proved to have ordering bugs (tests relying on state left over or setup from a previous test) that I wasn't able to easily/quickly solve, and I haven't applied Setup/Teardown functions universally, only where I specifically noticed them.

The patch file used was this:

```diff
@@
var foo identifier
@@
 import "testing"

-	t.Errorf(...)
+ s.Fail(...)

@@
@@
 import "github.com/stretchr/testify/assert"

-if err != nil {
-	t.Error(err)
-}
+assert.NoError(t, err)

@@
@@
 import "github.com/stretchr/testify/assert"

-assert.Nil(t, err)
+assert.NoError(t, err)

@@
var foo identifier
@@
-func foo(t *testing.T) {
+func (s *Suite) foo(t *testing.T) {
  ...
 }

@@
var name expression
@@
-t.Run(name, func(t *testing.T) {
+s.Run(name, func() {
  ...
 })

@@
var bar identifier
@@
 import "github.com/stretchr/testify/assert"

- assert.bar(t, ...)
+ s.bar(...)

@@
@@

-if err != nil {
-	t.Fatal(err)
-}
+s.Require().NoError(err)

@@
var foo identifier
@@
-func (s *Suite) foo(t *testing.T) {
+func (s *Suite) foo() {
  ...
 }

@@
var foo identifier
@@
-foo.AssertExpectations(t)
+foo.AssertExpectations(s.T())

# Need to change this one back
@@
@@
-    t.Cleanup(func() { mock.AssertExpectations(s.T()) })
+    t.Cleanup(func() { mock.AssertExpectations(t) })

@@
@@
import testUtil "github.com/astronomer/astro-cli/pkg/testing"

-    testUtil.MockUserInput(t, ...)
+    testUtil.MockUserInput(s.T(), ...)


# Reset the suite test back to taking testing.T
@@
var foo identifier
@@
import (
  "testing"
	"github.com/stretchr/testify/suite"
)

-func (s *Suite) foo() {
+func foo(t *testing.T) {
	  suite.Run(t, new(Suite))
 }


@@
var tt identifier
var foo identifier
[...]

PR #1192, opened 2023-05-09
Swap `strings.Contains` + `t.Errorf` to `assert.Contains` in tests

ashb:
## Description

Remove some more uses of `t.Errorf` in place of testify/assert methods

## 🎟 Issue(s)

Related #1185


ashb:
Oh yeah -- my IDE is doing that, but `gopatch` doesn't :D

PR #1202, opened 2023-05-11
set user agent header in CLI http requests

melugoyal:
## Description

set the User-Agent header in requests originating from the CLI, which is the standard for client identification

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/astro-client-core/client.go
+++ b/astro-client-core/client.go
@@ -41,6 +41,7 @@ func requestEditor(ctx httpContext.Context, req *http.Request) error {
 	req.Header.Add("x-astro-client-identifier", "cli")
 	req.Header.Add("x-astro-client-version", version.CurrVersion)
 	req.Header.Add("x-client-os-identifier", os+"-"+arch)
+	req.Header.Add("User-Agent", fmt.Sprintf("astro-cli/%s", version.CurrVersion))
 	return nil
 }
 


astro-client-core/client.go
```diff
@@ -41,6 +41,7 @@ func requestEditor(ctx httpContext.Context, req *http.Request) error {
 	req.Header.Add("x-astro-client-identifier", "cli")
 	req.Header.Add("x-astro-client-version", version.CurrVersion)
 	req.Header.Add("x-client-os-identifier", os+"-"+arch)
+	req.Header.Add("User-Agent", fmt.Sprintf("astro-cli/%s", version.CurrVersion))
```
kushalmalani, line 44: isn't this the same as 

req.Header.Add("x-astro-client-identifier", "cli")
req.Header.Add("x-astro-client-version", version.CurrVersion)

We use this for core-api to populate audit logs for cli requests

melugoyal: yeah it's the same use case. just felt like we should use User-Agent since it's more standard. Core API will start to get requests from
more places when it becomes public so we should start tracking/logging User-Agent there. this will help with that

melugoyal: we already log the user-agent in our logging library btw

PR #1355, opened 2023-08-15
Add more paths to dockerignore & gitignore

kaxil:
These dirs aren't needed and will allow users to run astro-cli along with airflow from the same dir

### DIFF
--- a/airflow/include/dockerignore
+++ b/airflow/include/dockerignore
@@ -3,3 +3,7 @@ astro
 .env
 airflow_settings.yaml
 logs/
+.venv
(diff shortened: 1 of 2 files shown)


airflow/include/dockerignore
```diff
@@ -3,3 +3,7 @@ astro
 .env
 airflow_settings.yaml
 logs/
+.venv
+airflow.db
+airflow.cfg
+
```
kaxil, line 9:
```suggestion
```

kushalmalani (approved):

PR #1365, opened 2023-08-23
update get deployment query

Simpcyclassy:
## Description

The GetDeployment query has been updated with some breaking changes on houston.

This PR changes

- The where argument from `id` to `deploymentId`
- Gets the first item in the deployment query return

## 🎟 Issue(s)

Related #astronomer/issues/issues/5763

## 🧪 Functional Testing

unit testing

## 📸 Screenshots

Only supported on
0.30 : 0.30.8 and upwards
0.32: 0.32.3 and upwards

And 0.33.1 and upwards

### DIFF
--- a/houston/deployment.go
+++ b/houston/deployment.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-var errDeploymentNotFound = "deployment with id %s not found"
+var errDeploymentNotFound = fmt.Errorf("deployment not found")
 
 // ListDeploymentsRequest - filters to list deployments according to set values
 type ListDeploymentsRequest struct {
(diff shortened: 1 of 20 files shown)


houston/deployment.go
```diff
@@ -551,7 +573,7 @@ func (h ClientImplementation) GetDeployment(deploymentID string) (*Deployment, e
 		return nil, handleAPIErr(err)
 	}
 
-	return &res.Data.GetDeployment, nil
+	return res.Data.GetDeployment, nil
```
neel-astro, line 576: nit: Do we expect more than one deployment to be returned by Houston for a query by id? If not, then I would say it might be simpler to check the length of the response and return just the first element, that way function signature would remain the same and we won't require any further changes in the caller functions.

Or if we go by what's currently implemented then in all caller functions instead of directly fetching the first element, we would still need to add check on the length of the array first.

```Golang
if len(res.Data.GetDeployment) > 0 {
        return &res.Data.GetDeployment[0], nil
}
return nil, fmt.Errorf("deployment with id %s not found", deploymentID)
```

neel-astro (commented):
Left a comment as of now to simplify the logic. Also would love to see some screenshots of the new query working fine with both active Software versions.

neel-astro (approved):
LGTM

PR #1441, opened 2023-11-08
Set default version back to v1.20.0

ashb:
Now that the issue has been fixed (https://status.astronomer.io/incidents/4z4r646s6mkx) we can make this version the default again


neel-astro (approved):
LGTM

kushalmalani (approved):

Issue #1387, opened 2023-09-13
astro dev pytest generated artifacts not persistant


SebOpo:
**Describe the bug**

Acorrindg to docs https://docs.astronomer.io/astro/cli/astro-dev-pytest  all files saved in `include` folder should be accessible outside contianer. In v1.19.0 this feature was dissabled: https://github.com/astronomer/astro-cli/commit/046705c8b098aab1ac1ca7a01af7e874643ed61a
Without volume mounts all generated files i.e code coverage or test reports are gone along with the pytest container .

```
astro dev pytest --args "--cov --cov-report xml:include/coverage.xml"
```

**What CLI Version did you experience this bug?**

1.19.0+

**This CLI bug is related to which Astronomer Platform?**

- [x] Astro
- [x] Software
- [ ] None/Unknown

**What Operating System is the above CLI installed on?**

**🪜 Steps To Reproduce**

Run
``` 
astro dev pytest --args "--cov --cov-report xml:include/coverage.xml"
```
Check if "coverage.xml" file exists in `include` folder outside docker container after test has finished.

sunkickr:
We removed the volumes because they were causing tests and dags to be missing in certain CI/CD tools we will have to think of a way to fix this issue with this in mind

sunkickr:
@neel-astro idea is to simulate the old behavior by coping the include folder from the docker container to the local include folder after the pytest runs

zach-nicoll-wcq:
Would love to see this fixed.

In fact, would love to fix it and open a PR.

I've found [this line](https://github.com/astronomer/astro-cli/blob/18d0c7515e5a614beb6860edb3d3179e6c4ecf79/airflow/docker_image.go#L169) that implies the test report is being copied out of the container using `docker cp`.

However, it doesn't look like `htmlReport` is being set to true at any point, so this isn't running.

Additionally, this doesn't take into account `.xml` coverage outputs.

Could we change the command to copy all `*.html` and `*.xml` files every time, regardless of `htmlReport` flag state?

Not sure who to @ here, so @sunkickr @neel-astro thoughts?

PR #1459, opened 2023-11-27
fix: #1387 persist include dir after running pytest

zach-nicoll-wcq:
## Description

After running `astro dev pytest`, the coverage reports created with the `include` dir are not being persisted outside of the test container.

This is a problem, because the coverage reports cannot then be used to update PR comments etc. with the report results. Essentially, coverage reports are useless.

In the current state, [this statement is false](https://docs.astronomer.io/astro/cli/astro-dev-pytest):

> If your test generates artifacts, such as code coverage reports, you can output the artifacts to the include folder of your Astro project so they can be accessed after the test has finished.

This PR makes the above statement true, copying the include folder on the host machine after the Pytest run completes.

## 🎟 Issue(s)

https://github.com/astronomer/astro-cli/issues/1387

## 🧪 Functional Testing

1. `astro dev init`
2. Add `pytest-cov==4.1.0` to `requirements.txt` (or any other pytest-cov version)
3. `astro dev pytest --args "--cov --cov-report=xml:include/coverage.xml --cov-report=html:include/html"`
4. Observe that the `include` dir is populated once the tests succeed

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -120,11 +120,13 @@ func (d *DockerImage) Pytest(pytestFile, airflowHome, envFile, testHomeDirectory
 		stdout = nil
 		stderr = nil
 	}
+
 	// create pytest container
 	docErr := cmdExec(dockerCommand, stdout, stderr, args...)
 	if docErr != nil {
 		return "", docErr
 	}
+
 	// cp DAGs folder
 	args = []string{
 		"cp",
@@ -135,6 +137,7 @@ func (d *DockerImage) Pytest(pytestFile, airflowHome, envFile, testHomeDirectory
 	if docErr != nil {
 		return "", docErr
 	}
+
 	// cp .astro folder
 	// on some machine .astro is being docker ignored, but not
 	// on every machine, hence to keep behavior consistent
@@ -148,11 +151,13 @@ func (d *DockerImage) Pytest(pytestFile, airflowHome, envFile, testHomeDirectory
 	if docErr != nil {
 		return "", docErr
 	}
+
 	// start pytest container
 	docErr = cmdExec(dockerCommand, stdout, stderr, []string{"start", "astro-pytest", "-a"}...)
 	if docErr != nil {
 		log.Debugf("Error starting pytest container: %s", docErr.Error())
 	}
+
 	// get exit code
 	args = []string{
 		"inspect",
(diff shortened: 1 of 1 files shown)


sunkickr:
did you try `astro dev pytest --args "--cov --cov-report xml:include/coverage.xml"`?

sunkickr:
The way volumes is handled by the pytest command changed which must have broke the functionality described in docs. we can add the cp command if needed. The code within `if htmlReport` is used by the `astro dev upgrade-test` command to create an html report using `pytest-html`. That code would need to stay

zach-nicoll-wcq:
> did you try `astro dev pytest --args "--cov --cov-report xml:include/coverage.xml"`?

Yes, if you see my functional testing steps, this is one of them to verify the functionality. Alas, it does not work in the current state.

> The way volumes is handled by the pytest command changed which must have broke the functionality described in docs. we can add the cp command if needed. The code within `if htmlReport` is used by the `astro dev upgrade-test` command to create an html report using `pytest-html`. That code would need to stay

I see - to simplify the process, would it be easier to just put the `dag-test-report.html` in the `include` dir, and then perform a copy? That way everything can be done at once.

zach-nicoll-wcq:
> @zach-nicoll-wcq we copy the test report into a specific testHomeDirectory which is separate from include. it makes sense to me to continue to only copy the html-report into testHomeDirectory for the upgrade-test command and add this more general copy of the include folder every time the pytest function is called

@sunkickr roger that - I have amended PR to perform the copy after the html report logic is run.

I also noticed that the container shutdown logic was repeated, so I refactored the error handling at the end of the Pytest function to log the docker errors, and only remove the docker container in a single place.

zach-nicoll-wcq:
@sunkickr @neel-astro @kushalmalani any movement on this? Would love to start using this functionality in my own projects.

airflow/docker_image.go
```diff
@@ -148,38 +151,43 @@ func (d *DockerImage) Pytest(pytestFile, airflowHome, envFile, testHomeDirectory
...
+	// Persist the include folder from the Docker container to local include folder
+	docErr = cmdExec(dockerCommand, nil, stderr, "cp", "astro-pytest:/usr/local/airflow/include/", ".")
+	if docErr != nil {
+		log.Debugf("Error copying include folder from the pytest container: %s", docErr.Error())
+	}
+
 	// delete container
-	err = cmdExec(dockerCommand, nil, stderr, "rm", "astro-pytest")
+	docErr = cmdExec(dockerCommand, nil, stderr, "rm", "astro-pytest")
 	if err != nil {
```
SebOpo, line 189: `if docErr != nil {`

zach-nicoll-wcq: @SebOpo have amended :)

zach-nicoll-wcq:
@sunkickr @neel-astro @kushalmalani @SebOpo 

Any movement? This feels like a very easy win for the project, and will align the functionality to what people expect from the docs...

sunkickr (approved):

sunkickr:
Worried about is verifying that other command that use the function still work:
 - `astro dev pytest` without using the --args flag
 - `astro dev upgrade-test` command

I'll just verify before release

PR #1479, opened 2024-01-02
Add --dag-only flag to the astro deploy CLI command and also validate the dag_deploy config

rujhan-arora-astronomer:
## Description

Add --dag-only flag to the astro deploy CLI command and also validate the dag_deploy config.
Will write the code to upload dags in the next PR

## 🎟 Issue(s)

Related #6055

## 🧪 Functional Testing

Tested it locally.
Added unit tests.

## 📋 Checklist

### DIFF
--- a/cmd/software/deploy.go
+++ b/cmd/software/deploy.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/astronomer/astro-cli/cmd/utils"
 	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/context"
+	"github.com/astronomer/astro-cli/houston"
 	"github.com/astronomer/astro-cli/pkg/git"
 	"github.com/astronomer/astro-cli/software/deploy"
 
@@ -52,7 +54,9 @@ func NewDeployCmd() *cobra.Command {
 	cmd.Flags().BoolVarP(&saveDeployConfig, "save", "s", false, "Save deployment in config for future deploys")
 	cmd.Flags().BoolVarP(&ignoreCacheDeploy, "no-cache", "", false, "Do not use cache when building container image")
 	cmd.Flags().StringVar(&workspaceID, "workspace-id", "", "workspace assigned to deployment")
-	cmd.Flags().BoolVarP(&isDagOnlyDeploy, "dag-only", "", false, "Deploy only those dags which are present in the /dags folder")
+	if !context.IsCloudContext() && houston.VerifyVersionMatch(houstonVersion, houston.VersionRestrictions{GTE: "0.34.0"}) {
+		cmd.Flags().BoolVarP(&isDagOnlyDeploy, "dag-only", "", false, "Deploy only those dags which are present in the /dags folder")
+	}
 	return cmd
 }
 
--- a/cmd/software/deploy.go
+++ b/cmd/software/deploy.go
@@ -20,16 +20,17 @@ var (
 
 	EnsureProjectDir   = utils.EnsureProjectDir
 	DeployAirflowImage = deploy.Airflow
+	isDagOnlyDeploy    bool
 )
 
 var deployExample = `
 Deployment you would like to deploy to Airflow cluster:
 
-  $ astro deploy <deployment-id>
+$ astro deploy <deployment-id>
 
 Menu will be presented if you do not specify a deployment name:
 
-  $ astro deploy
+$ astro deploy
 `
 
 const (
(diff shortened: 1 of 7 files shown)


houston/deployment.go
```diff
@@ -347,6 +347,9 @@ var (
 						type
 						url
 					}
+					dagDeployment {
```
neel-astro, line 350: just checking if this new field is present in all Houston versions from 0.29 onwards or added recently.

software/deploy/deploy.go
```diff
@@ -289,3 +291,29 @@ func getAirflowUILink(deploymentID string, deploymentURLs []houston.DeploymentUR
...
+	// Throw error if the feature is disabled at Deployment level
+	deploymentInfo, err := houston.Call(houstonClient.GetDeployment)(deploymentID)
+	if err != nil {
+		return fmt.Errorf("failed to get deployment info: %w", err)
+	}
+	if !isDagOnlyDeploymentEnabledForDeployment(deploymentInfo) {
+		return errDagOnlyDeployNotEnabledForDeployment
+	}
+
+	return nil
```
neel-astro, line 318: I don't see any logic that would just deploy DAGs from the local `/dags` folder. Am I missing something? Could you add more description in the PR on what the expectations and behavior observed during testing?

cmd/software/deploy.go
```diff
@@ -51,6 +52,7 @@ func NewDeployCmd() *cobra.Command {
 	cmd.Flags().BoolVarP(&saveDeployConfig, "save", "s", false, "Save deployment in config for future deploys")
 	cmd.Flags().BoolVarP(&ignoreCacheDeploy, "no-cache", "", false, "Do not use cache when building container image")
 	cmd.Flags().StringVar(&workspaceID, "workspace-id", "", "workspace assigned to deployment")
+	cmd.Flags().BoolVarP(&isDagOnlyDeploy, "dag-only", "", false, "Deploy only those dags which are present in the /dags folder")
```
neel-astro, line 55: Do we want to provide this feature flag for all software platform versions, if not we would need to shield it. Attaching examples for reference:
https://github.com/astronomer/astro-cli/blob/main/cmd/airflow.go#L154-L155
https://github.com/astronomer/astro-cli/blob/main/cmd/software/deployment.go#L223-L225

neel-astro, line 55: The flag comment does not seem inline with what's the actual behaviour of the command, i.e. we don't seem to be deploying any DAGs

neel-astro (commented):
Left few comments to be addressed

software/deploy/deploy.go
rujhan-arora-astronomer: @neel-astro I will work on it now and will create a new PR for it.  Should I add a todo comment?
The idea is to create small PRs rather than a big one

houston/deployment.go
rujhan-arora-astronomer: Yes it is

cmd/software/deploy.go
rujhan-arora-astronomer: Sure, let me add it 👍

rujhan-arora-astronomer: I will work on it now and will create a new PR for it. Should I add a todo comment?
The idea is to create small PRs rather than a big one @neel-astro

software/deploy/deploy.go
neel-astro: ok totally fine with that, could we then hide the flag from users, in case someone wants to cut out the release, they shouldn't be blocked on this 🙂

rujhan-arora-astronomer: cc: @kushalmalani  @sunkickr  Please review it, but don't merge it until the next PR which I will raise also gets approved. I will let you know.

cmd/software/deploy.go
neel-astro: Well, what's the expectation when a user uses the new feature flag with an older Software version?
I mean it feels misleading to say to the user ``` to perform this operation, set both deployments.dagOnlyDeployment and deployments.configureDagDeployment to true in the Astronomer Platform``` when that configuration is not available for older Houston versions

rujhan-arora-astronomer: Updated the msg

rujhan-arora-astronomer: Added the software version check

rujhan-arora-astronomer: Added the software version check

neel-astro (approved):
LGTM

sunkickr:
Why don't we use `astro deploy --dags` instead of `astro deploy --dag-only` to be consistent with the astro command? I know we also have a marketing campaign around `astro deploy --dags` so it would be consistent with our marketing materials as well

rishkarajgi:
> Why don't we use `astro deploy --dags` instead of `astro deploy --dag-only` to be consistent with the astro command? I know we also have a marketing campaign around `astro deploy --dags` so it would be consistent with our marketing materials as well

Agree. We should keep consistency with the astro command here and keep it as astro deploy --dags

PR #1516, opened 2024-02-01
Add the ability to change the deployment type to dag_deploy from the CLI (both for create and update)

rujhan-arora-astronomer:
## Description

Add the ability to change the deployment type to dag_deploy from the CLI (both for create and update)

## 🎟 Issue(s)

Related issue
https://github.com/astronomer/issues/issues/6119

## 🧪 Functional Testing

Locally

## 📸 Screenshots

## 📋 Checklist


rishkarajgi (approved):
lgtm

neel-astro (approved):

cmd/software/deployment.go
```diff
@@ -120,17 +131,18 @@ func newDeploymentCreateCmd(out io.Writer) *cobra.Command {
...
 		triggererEnabled = appConfig.Flags.TriggererEnabled
 		gitSyncDAGDeploymentEnabled = appConfig.Flags.GitSyncEnabled
 		runtimeEnabled = appConfig.Flags.AstroRuntimeEnabled
+		dagOnlyDeployEnabled = appConfig.Flags.DagOnlyDeployment
 	}
 
 	// let's hide under feature flag
-	if nfsMountDAGDeploymentEnabled || gitSyncDAGDeploymentEnabled {
-		cmd.Flags().StringVarP(&dagDeploymentType, "dag-deployment-type", "t", "", "DAG Deployment mechanism: image, volume, git_sync")
+	if nfsMountDAGDeploymentEnabled || gitSyncDAGDeploymentEnabled || dagOnlyDeployEnabled {
```
kushalmalani, line 144: 1. What happens when both `gitSyncDAGDeploymentEnabled` and `dagOnlyDeployEnabled` are set to true?
2. How does nfsMountDAGDeploymentEnabled being turned on or off impact this? I dont see we are using that value, so why are we checking it here?

rujhan-arora-astronomer: @kushalmalani 
1. If both are enabled, that's a valid case.
2. I am not sure tbh. Since this is an old existing code, I didn't want to touch it in this particular PR.

kushalmalani: For the second part, i see how there could be a bug. If we nfsMountDAGDeploymentEnabled as false and gitSyncDAGDeploymentEnabled and dagOnlyDeployEnabled are set to true, then the command description would display like

`deploymentTypeMsg := "DAG Deployment mechanism: image, volume, git_sync, dag_only"` which would be incorrect

rujhan-arora-astronomer: What should be the correct deploymentTypeMsg in this case?

kushalmalani: if my understanding is correct volume indicates nfsMountDAGDeploymentEnabled is true. so in the above case, the description should be 

`DAG Deployment mechanism: image, git_sync, dag_only`.

on a side note, i would suggest to not change the description based on what is enabled. You can keep the description the same

`cmd.Flags().StringVarP(&dagDeploymentType, "dag-deployment-type", "t", "", "DAG Deployment mechanism: image, volumne, git_sync, dag_only"`

Upon passing a particular type, if the flag is not enabled, we should throw an error message.

rujhan-arora-astronomer: Yeah, I've streamlined the message to show `DAG Deployment mechanism: image, volumne, git_sync, dag_only`. We have backend checks for these already

kushalmalani: Can you just define that a string constant instead of a function call.

kushalmalani (approved):
Thanks for addressing the comments

PR #1521, opened 2024-02-01
Update to latest Go versions

ashb:
Also switch from using deprecated CircleCI images in favour of the new ones


neel-astro (approved):

neel-astro:
I guess we would also need to upgrade compatible golangci-lint version, and fix the errors after that 😅

ashb:
Huh, the problem seems to be https://github.com/go-critic/go-critic/issues/1359

sunkickr (approved):

PR #1523, opened 2024-02-01
fix dag deploy updating on accident

sunkickr:
## Description

When running the command `astro deployment update` dag deploy is always turned off. This is because the case for dag deploy not changing didn't exist.

Added an assert to the update test to test for this bug.

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots
screenshot of updating the description of both a deployment with dag deploy enabled and disabled. Showing that in both cases dag deploy stays the same:
Screenshot 2024-02-02 at 11 46 31 AM

## 📋 Checklist

### DIFF
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -69,6 +69,7 @@ var (
 	tickNum                    = 10
 	timeoutNum                 = 180
 	dedicatedDeploymentRequest = astroplatformcore.UpdateDedicatedDeploymentRequest{}
+	dagDeployEnabled           bool
 )
 
 func newTableOut() *printutil.Table {
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -745,7 +745,8 @@ func Update(deploymentID, name, ws, description, deploymentName, dagDeploy, exec
 		}
 	}
 	// determine isDagDeployEnabled
-	if dagDeploy == enable {
+	switch dagDeploy {
+	case enable:
 		if currentDeployment.IsDagDeployEnabled {
 			fmt.Println("\nDAG deploys are already enabled for this Deployment. Your DAGs will continue to run as scheduled.")
 			return nil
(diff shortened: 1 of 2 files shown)


cloud/deployment/deployment.go
```diff
@@ -69,6 +69,7 @@ var (
 	tickNum                    = 10
 	timeoutNum                 = 180
 	dedicatedDeploymentRequest = astroplatformcore.UpdateDedicatedDeploymentRequest{}
+	dagDeployEnabled           bool
```
vandyliu, line 72: what was the reasoning behind putting it up here, it seems to fit the style of code the way we had it before

cloud/deployment/deployment.go
```diff
@@ -768,6 +767,10 @@ func Update(deploymentID, name, ws, description, deploymentName, dagDeploy, exec
 		dagDeployEnabled = false
 	}
 
+	if dagDeploy == "" {
```
vandyliu, line 770:
```suggestion
	else if dagDeploy == "" {
```

kushalmalani:
Can you add some screenshots for this. This bug fix is critical

cloud/deployment/deployment_test.go
```diff
@@ -1288,6 +1288,7 @@ func TestUpdate(t *testing.T) { //nolint
 		// success with hybrid type
 		err := Update("", "", ws, "", "", "", CeleryExecutor, "", "", "", "", "", "", "", 0, 0, workerQueueRequest, hybridQueueList, newEnvironmentVariables, true, mockCoreClient, mockPlatformCoreClient)
```
kushalmalani, line 1289: Can we add a comment up here w.r.t to what are we updating? 

Eventually would want to get rid of that huge list of arguments and have defined input type, but we can do that later.

cloud/deployment/deployment.go
sunkickr: We need it exist outside of the function so we can assert that it is the right value in the test

cloud/deployment/deployment.go
sunkickr: fixed

cloud/deployment/deployment_test.go
sunkickr: added comments

kushalmalani (approved):

PR #1525, opened 2024-02-02
Don't hardcode the registry URL to authenticate to, use the API response

ashb:
## Description

There were a few paths where the registry to log in to was hardcoded
based on the environment (but not all of them -- for instance the Push
path using the go client library was already using this field to auth against!)

This change makes the Push and Pull flow use the pre-existing
`ImageRepository` flow in the Deploy(ment) response from Astro.

This also removes a bit of duplicated code -- various places were
stripping of the `Bearer ` prefix from the password before calling the
Push function, but that function itself did that.

## 🧪 Functional Testing

I have used this to deploy to Astro cloud without trouble

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -36,6 +36,12 @@ const (
 
 var errGetImageLabel = errors.New("error getting image label")
 
+func realGetDockerClient() (client.APIClient, error) {
+	return client.NewClientWithOpts(client.FromEnv)
+}
+
+var getDockerClient func() (client.APIClient, error) = realGetDockerClient
+
 type DockerImage struct {
 	imageName string
 }
@@ -299,11 +305,16 @@ func (d *DockerImage) CreatePipFreeze(altImageName, pipFreezeFile string) error 
 	return nil
 }
 
-func (d *DockerImage) Push(registry, username, token, remoteImage string) error {
+func (d *DockerImage) Push(remoteImage, username, token string) error {
 	dockerCommand := config.CFG.DockerCommand.GetString()
 	err := cmdExec(dockerCommand, nil, nil, "tag", d.imageName, remoteImage)
 	if err != nil {
 		return fmt.Errorf("command '%s tag %s %s' failed: %w", dockerCommand, d.imageName, remoteImage, err)
+	}
+
+	registry, err := d.getRegistryToAuth(remoteImage)
+	if err != nil {
+		return err
 	}
 
 	// Push image to registry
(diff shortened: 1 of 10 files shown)


ashb:
I haven't yet tested this against the Houston API. That needs to happen.

kushalmalani:
> I haven't yet tested this against the Houston API. That needs to happen.

Adding @rujhan-arora-astronomer @rishkarajgi to confirm the software/deploy changes

kushalmalani:
Ran e2e tests from a binary of your branch, and tests around deploy look good - https://app.circleci.com/pipelines/github/astronomer/astro/115485/workflows/8cfe0817-29a4-4e92-b81b-c05534b7c3bc/jobs/5304060

airflow/docker_image.go
```diff
@@ -368,12 +379,34 @@ func (d *DockerImage) Push(registry, username, token, remoteImage string) error
...
+func (d *DockerImage) getRegistryToAuth(imageName string) (string, error) {
+	domain, err := config.GetCurrentDomain()
+	if err != nil {
+		return "", err
+	}
+
+	if domain == "localhost" {
+		return config.CFG.LocalRegistry.GetString(), nil
+	}
+	parts := strings.SplitN(imageName, "/", 2)
```
vandyliu, line 392: nit can we add a comment of what an example would be

vandyliu: nvm see it in the tests

vandyliu (approved):
code changes look good
tests have good coverage
i'm not too familiar with the docker pull/push part of this code so hopefully someone with more context can also approve

kushalmalani (approved):
LGTM. If possible, can you add some more tests on uncovered lines to bump the coverage

ashb:
I've increased coverage a little bit and refactored to get it a bit higher, but hitting the patch target isn't practical I don't think

rujhan-arora-astronomer (approved):
LGTM, except uncovered lines

airflow/docker_image.go
```diff
@@ -36,6 +36,12 @@ const (
 
 var errGetImageLabel = errors.New("error getting image label")
 
+func realGetDockerClient() (client.APIClient, error) {
+	return client.NewClientWithOpts(client.FromEnv)
+}
+
+var getDockerClient func() (client.APIClient, error) = realGetDockerClient
```
neel-astro, line 43:
```suggestion
var getDockerClient = func() (client.APIClient, error) {
	return client.NewClientWithOpts(client.FromEnv)
}
```
nit: just reducing an extra function declaration and figuring out a name for that 🙂

airflow/docker_image.go
```diff
@@ -353,27 +382,40 @@ func (d *DockerImage) Push(registry, username, token, remoteImage string) error
...
-	// Delete the image tags we just generated
-	err = cmdExec(dockerCommand, nil, nil, "rmi", remoteImage)
-	if err != nil {
-		return fmt.Errorf("command '%s rmi %s' failed: %w", dockerCommand, remoteImage, err)
+
+	if domain == "localhost" {
+		return config.CFG.LocalRegistry.GetString(), nil
 	}
-	return nil
+	parts := strings.SplitN(imageName, "/", 2)
```
neel-astro, line 401: could we try this change out with a custom registry on Software, not sure if it will break or not @rujhan-arora-astronomer @rishkarajgi 
with a custom registry, our image name would be of the form `fmt.Sprintf("%s:%s", registry, fmt.Sprintf("%s-%s", name, tag))`

neel-astro (approved):
left a minor nit and a scenario to validate, but otherwise LGTM, a good cleanup 🧹

airflow/docker_image.go
ashb: Oh yeah, there's a good chance that this won't work in that case.

ashb: We already have the domain here, so we could look if we are in custom registry mode and return.... `registry`? in that case?

neel-astro: If I recollect correctly, the reason we had such a structure was that some of the custom registry endpoints were of the form: `custom-registry.com/dev-airflow`, and it won't allow us to append `/<image_name>` after that.

The only part I am uncertain about is whether trying to log in to just `custom-registry.com` would work or not.

neel-astro: Wait a sec, for Software we don't pass username, so we won't try to login during image push, so I think we are good 🙂.
For software either we log in to the registry during `astro login` or we expect the customer to log in manually for the custom registry.

ashb: Oh yeah!

The login has to be to `custom-registry.com` (i.e. without the repo, but it's not _always_ a bare hostname, it could be `internal.company.com/registry` for instance)


So we're good to merge this?

neel-astro: yup

PR #1425, opened 2023-10-27
Migrate all Deployment Commands to Core

sunkickr:
## Description

migrate the following commands to Astro CORE API:
- astro deploy(just deployment selection)
- astro deployment list
- astro deployment create
- astro deployment create --deployment-file
- astro deployment update
- astro deployment update --deployment-file
- astro deployment logs
- astro deployment variable create
- astro deployment variable update
- astro deployment variable list
- astro deployment worker-queue create
- astro deployment worker-queue update
- astro deployment worker-queue delete
- astro dev upgrade-test(pull deployment image from astro)

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

manually tested all commands with hybrid/hosted/kubernetes/celery Deployments

## 📸 Screenshots

> Add screenshots to illustrate the validity of these changes.

## 📋 Checklist

- [ ] Rebased from the main (or release if patching) branch (before testing)
- [ ] Ran `make test` before taking out of draft
- [ ] Ran `make lint` before taking out of draft
- [ ] Added/updated applicable tests
- [ ] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary).
- [ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary).
- [ ] Communicated to/tagged owners of respective clients potentially impacted by these changes.
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/)

PR #1558, opened 2024-02-20
Fix `deployment variable list --save` with secrets

ashb:
## Description

After the migration to the REST API in #1425 it wasn't dealing with secret variables (which don't have a value anymore).

And while I'm there do a bit of tidying up:
- `_, x := range` instead of slice index access
- Remove the use of `environmentVariablesObjects` global variable!
- Refactoring of selecting the variable key to one place only (rather
  than in multiple)

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

I ran the command in question on a deployment with secret variables

## 📋 Checklist

### DIFF
--- a/cloud/deployment/deployment_variable.go
+++ b/cloud/deployment/deployment_variable.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -15,18 +16,16 @@ import (
 )
 
 var (
-	errVarBool                  = false
-	errVarCreateUpdate          = errors.New("there was an error while creating or updating one or more of the environment variables. Check the command output above for more information")
-	environmentVariablesObjects = []astroplatformcore.DeploymentEnvironmentVariable{}
-	printValue                  = "****"
+	errVarBool         = false
+	errVarCreateUpdate = errors.New(
+		"there was an error while creating or updating one or more of the environment variables. Check the command output above for more information",
+	)
 )
 
+const maskedSecret = "****"
+
 func VariableList(deploymentID, variableKey, ws, envFile, deploymentName string, useEnvFile bool, platformCoreClient astroplatformcore.CoreClient, out io.Writer) error {
-	varTab := printutil.Table{
-		Padding:        []int{5, 30, 30, 50},
-		DynamicPadding: true,
-		Header:         []string{"#", "KEY", "VALUE", "SECRET"},
-	}
+	environmentVariablesObjects := []astroplatformcore.DeploymentEnvironmentVariable{}
 
 	// get deployment
 	currentDeployment, err := GetDeployment(ws, deploymentID, deploymentName, false, platformCoreClient, nil)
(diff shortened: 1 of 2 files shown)


ashb:
This will need backporting to 1.23 too (it applies cleanly)

cloud/deployment/deployment_variable.go
```diff
@@ -37,42 +41,47 @@ func VariableList(deploymentID, variableKey, ws, envFile, deploymentName string,
...
-		if environmentVariablesObjects[i].Key == variableKey {
-			nbEnvVarFound++
-			varTab.AddRow([]string{strconv.Itoa(nbEnvVarFound), environmentVariablesObjects[i].Key, printValue, strconv.FormatBool(environmentVariablesObjects[i].IsSecret)}, false)
-			break
-		} else if variableKey == "" {
-			nbEnvVarFound++
-			varTab.AddRow([]string{strconv.Itoa(nbEnvVarFound), environmentVariablesObjects[i].Key, printValue, strconv.FormatBool(environmentVariablesObjects[i].IsSecret)}, false)
+	for i, variable := range environmentVariablesObjects {
+		if variable.Value != nil && !variable.IsSecret {
+			printValue = *variable.Value
```
vandyliu, line 64: do we not have to reset this to `****`
eg. the first variable is not secret then the next one is, won't do the next `printValue` be the same as the first one?

vandyliu: if that is indeed the case, we should add a test case to cover that
if it's not, maybe you can help me fill in the missing gaps im missing

cloud/deployment/deployment_variable.go
```diff
@@ -188,30 +197,27 @@ func readLines(path string) ([]string, error) {
...
-			_, err := f.WriteString("\n" + environmentVariablesObjects[i].Key + "=" + *environmentVariablesObjects[i].Value)
-			if err != nil {
-				fmt.Println("unable to write variable " + environmentVariablesObjects[i].Key + " to file:")
-				fmt.Println(err)
-			}
+	for _, variable := range environmentVariablesObjects {
+		var value string
+		if variable.IsSecret {
+			value = " # secret"
+		} else {
```
vandyliu, line 212:
```suggestion
		} else if variables.Value != nil {
```

cloud/deployment/deployment_variable.go
sunkickr: Screenshot 2024-02-20 at 6 36 17 PM
 seems to be working as intended

vandyliu: what if theres another secret var after the third one

kushalmalani: Agree with Vandy's comment here, can we just declare a temporary variable inside the loop? A unit test case is required here.

ashb: I didn't actually change this behaviour -- before it was

```go

printValue = *environmentVariablesObjects[i].Value
```

I agree though, while we're here lets fix it

ashb: Global variables bad!

kushalmalani (approved):

vandyliu (approved):
thanks for the fix and additionally the improvements!

PR #1564, opened 2024-02-22
Add dep api token crud

aliotta:
## Description

Adds the ability to manage deployment api tokens including Creation, Removal, Updating the apiToken role, and Listing.

## 🎟 Issue(s)

Related #18928

## 🧪 Functional Testing

Use the cli to create, update, remove and delete deployment api tokens passing in the flags that are required and testing both passing in and not passing in the optional flags

## 📸 Screenshots

Screenshot 2024-02-26 at 3 57 02 PM

## 📋 Checklist


cloud/apitoken/apitoken.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	fmt.Println("deployment: " + deployment)
+	resp, err := client.UpdateDeploymentApiTokenWithResponse(httpContext.Background(), ctx.Organization, deployment, apiToken.Id, mutateAPITokenInput)
+	if err != nil {
+		fmt.Println("error in MutateDeploymentApiTokenRoleWithResponse")
+		return err
+	}
+	err = astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body)
+	if err != nil {
+		fmt.Println("error in NormalizeAPIError")
```
vandyliu, line 102: @sunkickr @kushalmalani 
do we normally have this here and if so do we want to keep doing it like this?
I'm wondering since it would be meaningless for the user to see, but I guess it would be useful for us if we need to debug

cloud/apitoken/apitoken.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func RemoveDeploymentAPIToken(apiTokenID, deployment string, out io.Writer, client astrocore.CoreClient) error {
+	ctx, err := context.GetCurrentContext()
+	if err != nil {
+		return err
+	}
+
+	if apiTokenID == "" {
+		// Get all org apiTokens. Setting limit to 1000 for now
```
vandyliu, line 169:
```suggestion
		// Get all deployment apiTokens. Setting limit to 1000 for now
```

vandyliu:
@shri-astro @sreenusuuda 
Just a heads up for more QA that will need to be done for adding deployment tokens to CLI

cloud/apitoken/apitoken.go
sunkickr: I think most API calls have this but I have never seen it actually come up

sunkickr:
Why do we call it remove instead of delete? The api token is being deleted correct? Remove makes it sound like the token is being removed from the deployment but still exists

cloud/apitoken/apitoken.go
kushalmalani: Yeah would be useful for debugging. Can we improve that error message though? it looks like a debugging log

aliotta:
> Why do we call it remove instead of delete? The api token is being deleted correct? Remove makes it sound like the token is being removed from the deployment but still exists

Good call. That verbiage was a holdover from remove team from deployment but does not fit here. I will update to delete.

aliotta:
> Why do we call it remove instead of delete? The api token is being deleted correct? Remove makes it sound like the token is being removed from the deployment but still exists

done

cloud/apitoken/apitoken.go
aliotta: tbh it seems unecessary to me as the method that calls it prints out a prettier error message

e.g.

`shrividyahegde@Hegde ~ % astro deployment user update --deployment-id clt6yb6vm001r01nmwzffhbs9
Enter a user Deployment role or custom role name to update user: sample

Please select the user:
 #     FULLNAME            EMAIL                             ID                            DEPLOYMENT ROLE      CREATE DATE              
 1     Shrividya Hegde     shrividya.hegde@astronomer.io     cl27lfgxr8518182mwleoycmz     DEPLOYMENT_ADMIN     2022-04-20T13:11:07Z     

> 1
error in NormalizeAPIError
Error: Invalid request: Role sample does not exist
shrividyahegde@Hegde ~ % `

aliotta: I removed them

vandyliu (approved):

cmd/cloud/deployment.go
```diff
@@ -920,6 +925,148 @@ func removeDeploymentUser(cmd *cobra.Command, args []string, out io.Writer) erro
...
+func newDeploymentAPITokenCreateCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [apiTokenID]",
+		Short: "Create a apiToken in an Astro Deployment with a specific role",
+		Long:  "Create a apiToken in an Astro Deployment with a specific role\n$astro deployment apiToken add [apiTokenID] --role [ DEPLOYMENT_ADMIN or the custom role name ].",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return createDeploymentAPIToken(cmd, nil, out)
+		},
+	}
+	cmd.Flags().StringVarP(&apiTokenRole, "role", "r", "", "The role for the "+
```
kushalmalani, line 956: Should DEPLOYMENT_ADMIN be default role?

cloud/apitoken/apitoken.go
```diff
@@ -0,0 +1,257 @@
...
+// Prints a list of all of an deployments apiTokens
+//
+//nolint:dupl
+func ListDeploymentAPITokens(out io.Writer, client astrocore.CoreClient, deployment string) error {
+	table := printutil.Table{
+		Padding:        []int{30, 50, 10, 50, 10, 10, 10},
+		DynamicPadding: true,
+		Header:         []string{"NAME", "DESCRIPTION", "ID", "DEPLOYMENT ROLE", "CREATE DATE", "UPDATE DATE"},
+	}
+	apiTokens, err := GetDeploymentAPITokens(client, deployment, apiTokenPagnationLimit)
```
kushalmalani, line 233: if there are no api tokens, we should display an error message, rather than an empty table with only headers

cloud/apitoken/apitoken.go
```diff
@@ -0,0 +1,257 @@
...
+			return ErrNoAPITokenNameProvided
+		}
+	}
+
+	mutateAPITokenInput := astrocore.CreateDeploymentApiTokenRequest{
+		Role:        role,
+		Name:        name,
+		Description: &description,
+	}
+	resp, err := client.CreateDeploymentApiTokenWithResponse(httpContext.Background(), ctx.Organization, deployment, mutateAPITokenInput)
```
kushalmalani, line 45: we should save the api token from response to ASTRO_API_TOKEN variable. Right now, the token gets created, but user cannot use the token value

aliotta: will get a pr up for your comments

aliotta: I think instead of setting it we really want to print it to screen so they can use it where they want to.

kushalmalani: worried about the token leaking in ci cd logs...

PR #1613, opened 2024-04-01
Allow users to add dag_integrity_exceptions.txt to ignore errors

collinmcnulty:
## Description

Sometime a DAG doesn't parse in a CI pipeline for a reason that is ok. This provides a way to make that exception just for one dag file instead of having to use `-f`. What I'm not sure of right now is how to put this code such that astro dev parse is what's affected, as we were using an existing astro project.

## 🧪 Functional Testing

I wrote a dag that doesn't parse, made sure it failed astro dev parse. Then added it to tests/dag_integrity_exceptions.txt and it didn't fail to astro dev parse anymore.

## 📋 Checklist

### DIFF
--- a/airflow/airflow.go
+++ b/airflow/airflow.go
@@ -115,6 +115,7 @@ func Init(path, airflowImageName, airflowImageTag string) error {
 		"README.md":                            Readme,
 		"tests/dags/test_dag_example.py":       DagExampleTest,
 		".astro/test_dag_integrity_default.py": DagIntegrityTestDefault,
+		".astro/dag_integrity_exceptions.txt":  "# Add dag files to exempt from parse test below. ex: dags/<test-file>",
 	}
 
 	// Initailize directories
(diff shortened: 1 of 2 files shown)


sunkickr:
@collinmcnulty we should also make `astro dev init` create the file `.astro/dag_integrity_exceptions.txt` for users. If it was just an empty text file would that be okay?

airflow/airflow.go
```diff
@@ -115,6 +115,7 @@ func Init(path, airflowImageName, airflowImageTag string) error {
 		"README.md":                            Readme,
 		"tests/dags/test_dag_example.py":       DagExampleTest,
 		".astro/test_dag_integrity_default.py": DagIntegrityTestDefault,
+		".astro/dag_integrity_exceptions.txt":  "# Add dag files to exempt from parse test below. ex: dags/<test-file>",
```
kushalmalani, line 118: How would users with existing projects fix this? Do we need something in our release notes for this ?

sunkickr: yeah If they want this file they would need to delete `.astro/test_dag_integrity_default.py` and run `astro dev init`

PR #1670, opened 2024-06-24
Add `astro dbt deploy` command

jeremybeard:
## Description

This change adds a new command `astro dbt deploy` for deploying a dbt project to a deployment independently of the Astro project of the deployment. This uses the new bundle deploy type, where arbitrary files can be mounted on the Airflow containers at a selected mount path.

The selection of the deployment is in line with `astro deploy`, where a deployment id argument can be provided, or one can be selected from a list.

By default the dbt project will be mounted on the Airflow containers at `/usr/local/airflow/dbt/{dbt project name}`, where the project name is extracted from the dbt project's `dbt_project.yml` file. This mount path can be overridden with `--mount-path`.

The command is currently hidden behind the CLI config `dbt_deploys_enabled` until the feature is GA.

The `goerr113` lint rule is also removed because it was repeatedly being overridden, and doesn't seem to add a lot of value.

## 🧪 Functional Testing

- Unit tests added and updated
- Ran `astro dbt deploy` and manually downloaded the bundle to confirm its contents
- Ran `astro deploy --dags` to confirm DAG deploys are not impacted

## 📸 Screenshots

Screenshot 2024-06-24 at 3 35 28 PM

## 📋 Checklist

### DIFF
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -2,9 +2,11 @@ package deploy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	astrocore "github.com/astronomer/astro-cli/astro-client-core"
 	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
@@ -12,7 +14,6 @@ import (
 	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/pkg/fileutil"
 	"github.com/astronomer/astro-cli/pkg/git"
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -32,19 +33,19 @@ func DeployBundle(input *DeployBundleInput, platformCoreClient astroplatformcore
 	}
 
 	// get the current deployment so we can check the deploy is valid
-	currentDeployment, err := getDeployment(c.Organization, input.DeploymentID, platformCoreClient)
+	currentDeployment, err := deployment.CoreGetDeployment(c.Organization, input.DeploymentID, platformCoreClient)
 	if err != nil {
 		return err
 	}
 
 	// if CI/CD is enforced, check the subject can deploy
 	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(c.Token) {
-		return fmt.Errorf(errCiCdEnforcementUpdate, currentDeployment.Name) //nolint
+		return fmt.Errorf(errCiCdEnforcementUpdate, currentDeployment.Name)
 	}
 
 	// check the deployment is enabled for DAG deploys
 	if !currentDeployment.IsDagDeployEnabled {
-		return fmt.Errorf(enableDagDeployMsg, input.DeploymentID) //nolint
+		return fmt.Errorf(enableDagDeployMsg, input.DeploymentID)
 	}
 
 	// retrieve metadata about the local Git checkout. returns nil if not available
@@ -58,7 +59,7 @@ func DeployBundle(input *DeployBundleInput, platformCoreClient astroplatformcore
 
 	// check we received an upload URL
 	if deploy.BundleUploadUrl == nil {
-		return errors.New("No bundle upload URL received from Astro")
+		return errors.New("no bundle upload URL received from Astro")
 	}
 
 	// upload the bundle
@@ -115,19 +116,6 @@ func uploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool)
 	return versionID, nil
 }
 
-func getDeployment(organizationID, deploymentID string, platformCoreClient astroplatformcore.CoreClient) (*astroplatformcore.Deployment, error) {
-	resp, err := platformCoreClient.GetDeploymentWithResponse(context.Background(), organizationID, deploymentID)
-	if err != nil {
-		return nil, err
-	}
-	err = astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body)
-	if err != nil {
-		return nil, err
-	}
-
-	return resp.JSON200, nil
-}
-
 func createBundleDeploy(organizationID string, input *DeployBundleInput, deployGit *astrocore.DeployGit, coreClient astrocore.CoreClient) (*astrocore.Deploy, error) {
 	request := astrocore.CreateDeployRequest{
 		Description:     &input.Description,
@@ -174,7 +162,7 @@ func finalizeBundleDeploy(organizationID, deployID, tarballVersion string, input
 }
 
 func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
-	if git.HasUncommittedChanges() {
+	if git.HasUncommittedChanges(bundlePath) {
 		logrus.Warn("Local repository has uncommitted changes, skipping Git metadata retrieval")
 		return nil
 	}
@@ -182,23 +170,29 @@ func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
 	gitMetadata := &astrocore.DeployGit{}
 
 	// get the remote repository details, assume the remote is named "origin"
-	host, account, repo, err := git.GetRemoteRepository("origin")
+	repoURL, err := git.GetRemoteRepository(bundlePath, "origin")
 	if err != nil {
 		logrus.Debugf("Failed to retrieve remote repository details, skipping Git metadata retrieval: %s", err)
 		return nil
 	}
-	switch host {
+	switch repoURL.Host {
 	case "github.com":
 		gitMetadata.Provider = astrocore.DeployGitProviderGITHUB
 	default:
-		logrus.Debugf("Unsupported Git provider, skipping Git metadata retrieval: %s", host)
+		logrus.Debugf("Unsupported Git provider, skipping Git metadata retrieval: %s", repoURL.Host)
 		return nil
 	}
-	gitMetadata.Account = account
-	gitMetadata.Repo = repo
+	urlPath := strings.TrimPrefix(repoURL.Path, "/")
+	firstSlashIndex := strings.Index(urlPath, "/")
+	if firstSlashIndex == -1 {
+		logrus.Debugf("Failed to parse remote repository path, skipping Git metadata retrieval: %s", repoURL.Path)
+		return nil
+	}
+	gitMetadata.Account = urlPath[:firstSlashIndex]
+	gitMetadata.Repo = urlPath[firstSlashIndex+1:]
 
 	// get the path of the bundle within the repository
-	path, err := git.GetLocalRepositoryPathPrefix(bundlePath)
+	path, err := git.GetLocalRepositoryPathPrefix(bundlePath, bundlePath)
 	if err != nil {
 		logrus.Debugf("Failed to retrieve local repository path prefix, skipping Git metadata retrieval: %s", err)
 		return nil
@@ -208,7 +202,7 @@ func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
 	}
 
 	// get the branch of the local commit
-	branch, err := git.GetBranch()
+	branch, err := git.GetBranch(bundlePath)
 	if err != nil {
 		logrus.Debugf("Failed to retrieve branch name, skipping Git metadata retrieval: %s", err)
 		return nil
@@ -216,7 +210,7 @@ func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
 	gitMetadata.Branch = branch
 
 	// get the SHA of the local commit
-	sha, err := git.GetHeadCommitSHA()
+	sha, err := git.GetHeadCommitSHA(bundlePath)
 	if err != nil {
 		logrus.Debugf("Failed to retrieve commit SHA, skipping Git metadata retrieval: %s", err)
 		return nil
@@ -224,15 +218,16 @@ func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
 	gitMetadata.CommitSha = sha
 
 	// derive the remote URL of the local commit
-	commitURL, err := git.GetCommitURL(host, account, repo, sha)
-	if err != nil {
-		logrus.Debugf("Failed to derive commit URL, skipping Git metadata retrieval: %s", err)
+	switch repoURL.Host {
+	case "github.com":
+		gitMetadata.CommitUrl = fmt.Sprintf("https://%s/%s/%s/commit/%s", repoURL.Host, gitMetadata.Account, gitMetadata.Repo, sha)
+	default:
+		logrus.Debugf("Unsupported Git provider, skipping Git metadata retrieval: %s", repoURL.Host)
 		return nil
 	}
-	gitMetadata.CommitUrl = commitURL
 
 	// get the author name of the local commit
-	authorName, _, err := git.GetHeadCommitAuthor()
+	authorName, _, err := git.GetHeadCommitAuthor(bundlePath)
 	if err != nil {
 		logrus.Debugf("Failed to retrieve commit author, skipping Git metadata retrieval: %s", err)
 		return nil
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -1,0 +1,247 @@
+package deploy
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	astrocore "github.com/astronomer/astro-cli/astro-client-core"
+	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/pkg/fileutil"
+	"github.com/astronomer/astro-cli/pkg/git"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type DeployBundleInput struct {
+	BundlePath   string
+	MountPath    string
+	DeploymentID string
+	BundleType   string
+	Description  string
+	Wait         bool
+}
+
+func DeployBundle(input *DeployBundleInput, platformCoreClient astroplatformcore.CoreClient, coreClient astrocore.CoreClient) error {
+	c, err := config.GetCurrentContext()
+	if err != nil {
+		return err
+	}
+
+	// get the current deployment so we can check the deploy is valid
+	currentDeployment, err := getDeployment(c.Organization, input.DeploymentID, platformCoreClient)
+	if err != nil {
+		return err
+	}
+
+	// if CI/CD is enforced, check the subject can deploy
+	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(c.Token) {
+		return fmt.Errorf(errCiCdEnforcementUpdate, currentDeployment.Name) //nolint
+	}
+
+	// check the deployment is enabled for DAG deploys
+	if !currentDeployment.IsDagDeployEnabled {
+		return fmt.Errorf(enableDagDeployMsg, input.DeploymentID) //nolint
+	}
+
+	// retrieve metadata about the local Git checkout. returns nil if not available
+	gitMetadata := retrieveLocalGitMetadata(input.BundlePath)
+
+	// initialize the deploy
+	deploy, err := createBundleDeploy(c.Organization, input, gitMetadata, coreClient)
+	if err != nil {
+		return err
+	}
+
+	// check we received an upload URL
+	if deploy.BundleUploadUrl == nil {
+		return errors.New("No bundle upload URL received from Astro")
+	}
+
+	// upload the bundle
+	tarballVersion, err := uploadBundle(config.WorkingPath, input.BundlePath, *deploy.BundleUploadUrl, false)
+	if err != nil {
+		return err
+	}
+
+	// finalize the deploy
+	_, err = finalizeBundleDeploy(c.Organization, deploy.Id, tarballVersion, input, coreClient)
+	if err != nil {
+		return err
+	}
+	fmt.Println("Successfully uploaded bundle with version " + tarballVersion + " to Astro.")
+
+	// if requested, wait for the deploy to finish by polling the deployment until it is healthy
+	if input.Wait {
+		err = deployment.HealthPoll(currentDeployment.Id, currentDeployment.WorkspaceId, dagOnlyDeploySleepTime, tickNum, timeoutNum, platformCoreClient)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func uploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool) (string, error) {
+	tarFilePath := filepath.Join(tarDirPath, "bundle.tar")
+	defer func() {
+		err := os.Remove(tarFilePath)
+		if err != nil {
+			fmt.Println("\nFailed to delete tar file: ", err.Error())
+			fmt.Println("\nPlease delete the tar file manually from path: " + tarFilePath)
+		}
+	}()
+
+	// Generate the bundle tar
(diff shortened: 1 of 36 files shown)


cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.JSON200, nil
+}
```
ashb, line 129: This exists almost identically as `CoreGetDeployment()` from ./cloud/deployment  package -- could we use that instead please?

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+	resp, err := coreClient.CreateDeployWithResponse(context.Background(), organizationID, input.DeploymentID, request)
+	if err != nil {
+		return nil, err
+	}
+	err = astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	return resp.JSON200, nil
+}
```
ashb, line 159: Similar here, rather than a whole new function, could this be merged in to `createDeploy` in this package (inside `./deploy.go`)?

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+
+func finalizeBundleDeploy(organizationID, deployID, tarballVersion string, input *DeployBundleInput, coreClient astrocore.CoreClient) (*astrocore.UpdateDeployResponse, error) {
+	request := astrocore.UpdateDeployRequest{
+		BundleTarballVersion: &tarballVersion,
+	}
+	resp, err := coreClient.UpdateDeployWithResponse(context.Background(), organizationID, input.DeploymentID, deployID, request)
```
ashb, line 165: `finalizeDeploy` calls `platformCoreClient.FinalizeDeployWithResponse` -- what's the difference bewtween finalizeDeploy and updateDeploy? (and why do we have both)

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, err
+	}
+	return resp, nil
+}
+
+func retrieveLocalGitMetadata(bundlePath string) *astrocore.DeployGit {
+	if git.HasUncommittedChanges() {
+		logrus.Warn("Local repository has uncommitted changes, skipping Git metadata retrieval")
+		return nil
+	}
```
ashb, line 180: I don't know where this ultimately ends up, but instead of having no git metadata could we have the `-dirty` flag appended like `git describe does`:

```bash
$ git describe --tags --dirty
v1.27.0-12-gd79dc210-dirty
```

pkg/git/git.go
```diff
@@ -1,25 +1,104 @@
...
-// HasUncommittedChanges checks repository for uncommitted changes
-func HasUncommittedChanges() bool {
-	if IsGitRepository() {
-		// There is no system independent way to get exit code
-		// so we treat all exit codes the same, this works as long as we check IsGitRepository first
-		_, err := exec.Command("git", "diff", "--quiet", "HEAD").Output()
-		if err != nil {
-			return true
-		}
+func GetRemoteRepository(remote string) (host, account, repo string, err error) {
```
ashb, line 29: I feel we should probably have some unit tests of this parsing function please.

I think it will fail on ssh URLs like this one `﻿git@github.com:astronomer/astro-cli`

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,154 @@
...
+		// if provided, use the deployment ID from the command argument
+		deploymentID = args[0]
+	} else {
+		// otherwise, prompt the user to select a deployment
+		selectedDeployment, err := deployment.GetDeployment(workspaceID, "", deploymentName, false, nil, platformCoreClient, astroCoreClient)
+		if err != nil {
+			return err
+		}
+		deploymentID = selectedDeployment.Id
+	}
```
ashb, line 109: Is it worth extracting this into a common/shared function so that the two deploy commands always work the same?

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,154 @@
...
+	fmt.Println("Initiating dbt deploy for deployment ID: " + deploymentID)
+
+	// if the mount path is not provided, derive it from the dbt project name
+	if mountPath == "" {
```
ashb, line 113: Do we need to validate that there aren't any "funny" things in the mount path such as `../../` etc? (Does CoreAPI also validate this?)

ashb (commented):
A few Qs, and one issue in the git URL parsing (some of my Qs might be about the CoreAPI endpoints though, rather than the CLI impl)

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,154 @@
...
+	dbtProjectPath string
+
+	DeployBundle = cloud.DeployBundle
+)
+
+const (
+	dbtDeployExample = `
+Specify the ID of the Deployment on Astronomer you would like to deploy this dbt project to:
+
+  $ astro dbt deploy <deployment ID>
```
neel-astro, line 27: This concerns me a little because we are using `astro dbt deploy` which implicitly states that dbt is a resource in this case that we would be deploying to, but then we end up passing `deployment id` as the argument to the command. 
Also, this goes against the existing convention of passing the deploy type as a flag to `astro deploy`, like `astro deploy --dags`

I think we could go with one of the following:
`astro dbt deploy --deployment <deployment ID>`
`astro deploy --dbt <deployment ID>` (Because we already assume deploy is an action on the deployment implicitly, a probable course correction for 2.x 🙂 )

Might prefer 1, as I can see ourselves adding new commands in the dbt space from the notion doc

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,154 @@
...
+		Use:     "deploy DEPLOYMENT-ID",
+		Short:   "Deploy your dbt project to a Deployment on Astro",
+		Long:    "Deploy your dbt project to a Deployment on Astro. This command bundles your dbt project files and uploads it to your Deployment.",
+		Args:    cobra.MaximumNArgs(1),
+		RunE:    deployDbt,
+		Example: dbtDeployExample,
+	}
+
+	cmd.Flags().StringVarP(&mountPath, "mount-path", "m", "", "Path to mount dbt project in Airflow, for reference by DAGs. Default /usr/local/dbt/{dbt project name}")
+	cmd.Flags().StringVarP(&dbtProjectPath, "project-path", "b", "", "Path to the dbt project to deploy. Default current directory")
```
neel-astro, line 61:
```suggestion
	cmd.Flags().StringVarP(&dbtProjectPath, "project-path", "p", "", "Path to the dbt project to deploy. Default current directory")
```
nit: maybe a better short flag, or is it already taken up by something?

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	cloud "github.com/astronomer/astro-cli/cloud/deploy"
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/astronomer/astro-cli/config"
+	"github.com/pkg/errors"
```
neel-astro, line 11:
```suggestion
	"errors"
```
same argument as before regarding pkg/errors being not actively developed

cmd/cloud/dbt_test.go
```diff
@@ -0,0 +1,154 @@
...
+	astrocore "github.com/astronomer/astro-cli/astro-client-core"
+	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
+	astroplatformcore_mocks "github.com/astronomer/astro-cli/astro-client-platform-core/mocks"
+	cloud "github.com/astronomer/astro-cli/cloud/deploy"
+	testUtil "github.com/astronomer/astro-cli/pkg/testing"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestDbtDeploy_PickDeployment(t *testing.T) {
```
neel-astro, line 18: same as before, might want to restructure these tests to use the new `testify/suite`

config/cloud.go
```diff
@@ -65,7 +65,7 @@ func (c *Context) GetPublicGraphQLAPIURL() string {
 // GetPublicRESTAPIURL returns full core API Url for the provided Context
 func (c *Context) GetPublicRESTAPIURL(version string) string {
 	if c.Domain == localhostDomain || c.Domain == astrohubDomain {
-		return CFG.LocalCore.GetString()
+		return CFG.LocalCore.GetString() + "/" + version
```
neel-astro, line 68: good catch

pkg/git/git.go
```diff
@@ -1,25 +1,104 @@
...
 
 import (
+	"encoding/json"
+	"fmt"
+	"net/url"
 	"os/exec"
+	"strings"
 )
 
+var Path = ""
```
neel-astro, line 11: nit: could this be passed on as function argument instead of a global variable, as it would be hard to track the value of a global function if down the line multiple callers are interacting with this package

neel-astro (commented):
Left some comments to be squashed, but looks fine overall 🚀

cloud/deploy/bundle.go
jeremybeard: In v1alpha1 updateDeploy doesn't allow the user to just update the deploy entry, it is only used to tell Astro to finalize the deploy (i.e. roll out the image/DAGs to the deployment). In v1beta1 we decided to follow CRUD semantics more closely and have updateDeploy be for updating the deploy entry and finalizeDeploy be a verb route (/finalize) that instructs Astro to roll out the deploy.

cloud/deploy/bundle.go
jeremybeard: The intention of the Git metadata is to correlate a deploy with a commit that users can look up to see what the contents are/were (like we have done for GitHub integration), so we don't want to stamp the deploy with metadata that only relates to a dirty state on a local machine.

cmd/cloud/dbt.go
jeremybeard: The mount path is absolute. I expect before GA we will at least restrict users from specifying `/usr/local/airflow/dags` to not interfere with DAG deploys. I'd prefer to maintain that list only on the API side so we don't have them go out of sync with old CLI versions etc.

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// get the current deployment so we can check the deploy is valid
+	currentDeployment, err := getDeployment(c.Organization, input.DeploymentID, platformCoreClient)
+	if err != nil {
+		return err
+	}
+
+	// if CI/CD is enforced, check the subject can deploy
+	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(c.Token) {
+		return fmt.Errorf(errCiCdEnforcementUpdate, currentDeployment.Name) //nolint
```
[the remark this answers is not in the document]
jeremybeard: Ha, I actually did drop the linter, but missed removing that comment.

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,247 @@
...
+	"os"
+	"path/filepath"
+
+	astrocore "github.com/astronomer/astro-cli/astro-client-core"
+	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
+	"github.com/astronomer/astro-cli/cloud/deployment"
+	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/pkg/fileutil"
+	"github.com/astronomer/astro-cli/pkg/git"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
jeremybeard: Thanks, this is all across the CLI, but I'm happy to clean it up.

cmd/cloud/dbt.go
jeremybeard: Agree that it is inconsistent with the existing deploy command. It's more of a product thing. We want a dedicated top-level dbt command that we can show off to analytic engineers who build dbt projects and don't want to be concerned with the Airflow things like images or DAGs that their data engineers have set up for them. And yeah, we can use this structure for `astro dbt delete` and maybe others too.

I can switch the deployment id from an argument to a flag though. It would probably want to be `--deployment-id` to clearly distinguish from `--deployment-name`.

cmd/cloud/dbt.go
jeremybeard: Thanks, that was an oversight.

cmd/cloud/dbt.go
jeremybeard: It's a fair question. In this change I really grappled with how much I wanted to unravel the existing deploy logic to fit in bundles. Frankly what is already there has over the years become a mess of poorly scoped function calls. This spot is a good example -- the existing deploy command has deployment resolution (including interactive prompting) not in the command code (where I would say it belongs) but buried deep in the service code in `deployment.GetDeployment`. Here this also calls that function so at least we're reusing that part.

cloud/deploy/bundle.go
jeremybeard: True, will use that.

cloud/deploy/bundle.go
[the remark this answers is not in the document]
jeremybeard: Hmm turns out the CLI heavily uses things in pkg/errors that is not in errors (e.g. `errors.Wrap`) so I can fix it for my change but I'll leave out fixing the rest of the CLI here.

cloud/deploy/deploy.go
```diff
@@ -160,47 +160,23 @@ func shouldIncludeMonitoringDag(deploymentType astroplatformcore.DeploymentType)
...
 			return "", err
 		}
-	}
 
-	// Generate the dags tar
-	err := fileutil.Tar(dagsPath, path)
-	if err != nil {
-		return "", err
+		// Remove the monitoring dag file after the upload
+		defer os.Remove(monitoringDagPath)
```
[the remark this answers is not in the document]
jeremybeard: I believe I did it that way, can you check again?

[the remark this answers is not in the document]
neel-astro: ohh lol, I misread the color code 🙈

cloud/deploy/bundle.go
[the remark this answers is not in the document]
neel-astro: That makes sense, I will keep an eye out to clean it up as well
I cleaned it up a couple of years ago (https://github.com/astronomer/astro-cli/pull/484), sad to see that it has crept it again.

pkg/git/git.go
jeremybeard: TIL about Git having these "SCP-like" URLs.

neel-astro (approved):
LGTM

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,150 @@
...
+	dbtProjectPath string
+
+	DeployBundle = cloud.DeployBundle
+)
+
+const (
+	dbtDeployExample = `
+Specify the ID of the Deployment on Astronomer you would like to deploy this dbt project to:
+
+  $ astro dbt deploy -d <deployment ID>
```
kushalmalani, line 27: just to stay consistent with deploy flow where in we pass the deploymentId as part of the command and not via a flag.

`astro dbt deploy <deployment ID>`

Scanning through the code base, i see few instances where do use the flag. I think moving forward, we should pick one approach and stay consistent with it. IMO i like this flow better, since its something users are currently used to plus it leaves us the opportunity of using short hand flags for other purposes

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,150 @@
...
+
+	dbtDefaultMountPathPrefix = "/usr/local/airflow/dbt/"
+	dbtProjectYmlFilename     = "dbt_project.yml"
+	dbtBundleType             = "dbt"
+)
+
+func newDbtCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "dbt",
+		Short: "Manage your dbt projects deployed to Deployments running on Astronomer",
```
kushalmalani, line 42:
```suggestion
		Short: "Deploy your dbt project to a Deployment on Astro",
```

Stays consistent with other deploy command and plus since Astronomer has multiple products managed by cli, gives clarity which product is this command for

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,150 @@
...
+	cmd.SilenceUsage = true
+
+	// if the dbt project path is not provided, use the current directory
+	if dbtProjectPath == "" {
+		dbtProjectPath = config.WorkingPath
+	}
+
+	// check that there is a valid dbt project at the dbt project path
+	dbtProjectYamlPath := filepath.Join(dbtProjectPath, dbtProjectYmlFilename)
+	if _, err := os.Stat(dbtProjectYamlPath); os.IsNotExist(err) {
```
kushalmalani, line 81: Can we do this as part of PreRunE and move this check to utils

cmd/cloud/dbt.go
```diff
@@ -0,0 +1,150 @@
...
+		var err error
+		workspaceID, err = coalesceWorkspace()
+		if err != nil {
+			return fmt.Errorf("failed to find a valid workspace: %w", err)
+		}
+	}
+
+	// if the deployment ID is not provided, get it from the deployment name or prompt the user
+	if deploymentID == "" {
+		selectedDeployment, err := deployment.GetDeployment(workspaceID, "", deploymentName, false, nil, platformCoreClient, astroCoreClient)
```
kushalmalani, line 100: We should do this under cloud/deploy/bundle.go. My understanding of cmd/cloud is for command definitions. Any work specific to the application should be under cloud/ and cmd/cloud should deal with constructing the command based on the flags passed and any pre-checks needed

cmd/cloud/root.go
```diff
@@ -23,10 +24,14 @@ func AddCmds(astroPlatformCoreClient astroplatformcore.CoreClient, coreClient as
...
 	astroCoreIamClient = iamCoreClient
 	airflowAPIClient = airflowClient
-	return []*cobra.Command{
+	cmds := []*cobra.Command{
 		NewDeployCmd(),
 		newDeploymentRootCmd(out),
 		newWorkspaceCmd(out),
 		newOrganizationCmd(out),
 	}
+	if config.CFG.DbtDeploysEnabled.GetBool() {
```
kushalmalani, line 33: I dont think we need this, since this is behind a feature flag anyway, and folks trying to use this without the feature flag would be thrown the appropriate error.

config/config.go
```diff
@@ -57,7 +57,7 @@ var (
 		Contexts:              newCfg("contexts", ""),
 		DockerCommand:         newCfg("container.binary", "docker"),
 		LocalAstro:            newCfg("local.astrohub", "http://localhost:8871/v1"),
-		LocalCore:             newCfg("local.core", "http://localhost:8888/v1alpha1"),
+		LocalCore:             newCfg("local.core", "http://localhost:8888"),
```
kushalmalani, line 60: 👍

cloud/deploy/bundle.go
kushalmalani: +1 to what jeremy said. We will eventually get the v1alpha1 and v1beta1 terminology the same, once we know all users have moved to a cli version that uses v1beta1 endpoints and we have a couple of customers using v1alpha1 endpoints, those will be migrated to using v1beta1 endpoints

cmd/cloud/dbt.go
jeremybeard: @neel-astro I think @kushalmalani is right, it is more consistent with the existing commands to have it as an argument. I'll change it back to that, but let me know if you strongly disagree.

cmd/cloud/dbt.go
jeremybeard: What is the advantage of calling it from `PreRunE`? I like having these checks in the `Run` function so it's in one place. `astro deploy` has a few checks in the `Run` function and that doesn't seem to be an issue. I will move the logic to a util though.

cmd/cloud/dbt.go
jeremybeard: I'd say this part is about the command, because we have multiple ways that the command can specify the deployment:
- The id as an argument
- The name as a flag
- Otherwise from an interactive prompt, which itself can be filtered by another (workspace id) flag

I don't think it is right to pass down a bunch of argument/flag values to the service function and expect deployment resolution there -- that is coupling the application logic of deploying a bundle to the concern that the command that called it has multiple ways of specifying the deployment, and in this case it's even worse than that because it can sometimes involve an interactive prompt.

It should help that line 100 here is passing all the work of talking to the API to an existing service function.

cmd/cloud/root.go
jeremybeard: I think it's a good idea to keep the command hidden by default until the feature is at least in public preview. We did it like this for environment objects and it worked well.

cmd/cloud/dbt.go
jeremybeard: That `astro deploy` as a command doesn't refer to a resource is already confusing yeah. Thankfully there is no "dbt" resource in Astro that we would be getting mixed up with here.

cmd/cloud/root.go
kushalmalani: We shouldnt be hiding the command...If folks run it they would see the error message and if they are curious and want to opt in, thats a good sign for us. Plus for anyone who wants to opt in, i dont see a knob for 2 configurations to be enabled for them to use this...

cmd/cloud/dbt.go
jeremybeard: I know, but we don't have to use it because it exists. In my view it isn't helpful to have an artificial split like that. We don't use it for most checks in `astro deploy` -- it only checks that it is running in a project directory, not whether the workspace can be determined, or that the `dags` and `image` flags aren't used together, or that the config can be saved, or that there aren't uncommitted changes.

cmd/cloud/dbt.go
kushalmalani: You make a good point and i do see an advantage in this, its just my understanding on how i have understood this cobra package. We might need to refactor this in a bunch of different places in the future

cmd/cloud/dbt.go
kushalmalani: I just checked that the path is an user input and not derived from where this command is executed. This flow is different from how dag/image deploys work. We might want to confirm this with product once just so we are on the same page. This was not how i was imagining this check would work

My comment was running the preRunE check to confirm the current directory path is a dbt project or not.

jeremybeard: > the path is an user input and not derived from where this command is executed

Not quite, the code directly above this uses the current working directory by default, but the user can override it with a flag. This seems pretty useful as an option for CI where you've cloned a repo and want to deploy a dbt project in some subdirectory of it.

jeremybeard: Afaik we also do this in `astro deploy` with the hidden (?) `dags-path` flag.

kushalmalani: gotcha! moving this to utils should be good then

kushalmalani:
@rujhan-arora-astronomer Looping you in since there are some changes to the software code because of change in a common util function.

cmd/cloud/dbt.go
jeremybeard: I've made it a separate function in the file so we can reuse it for `astro dbt delete`.

cmd/cloud/root.go
jeremybeard: I've removed the config and the command is now always added.

cloud/deploy/bundle.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func DeployBundle(input *DeployBundleInput, platformCoreClient astroplatformcore.CoreClient, coreClient astrocore.CoreClient) error {
+	c, err := config.GetCurrentContext()
+	if err != nil {
+		return err
+	}
+
+	// get the current deployment so we can check the deploy is valid
+	currentDeployment, err := deployment.CoreGetDeployment(c.Organization, input.DeploymentID, platformCoreClient)
```
kushalmalani, line 36: Seems like we are doing get deployment twice in this flow...Once in cmd/cloud/dbt.go and again in this function call.

PR #1672, opened 2024-07-01
Add `astro dbt delete` command

jeremybeard:
## Description

This change adds a new command `astro dbt delete`, to delete a dbt project bundle from a deployment. This is effectively the inverse of `astro dbt deploy`. The command effects a delete by creating a new deploy where the reported bundle tarball version is left empty, which Astro understands to mean that the bundle should be deleted from the deployment.

The dbt project to be deleted is automatically inferred from the current working directory, or the project path can be specified, or if the dbt project is not locally accessible the mount path can be provided instead.

## 🧪 Functional Testing

- Unit tests added
- Manually tested

## 📸 Screenshots

Screenshot 2024-07-01 at 1 41 54 PM

## 📋 Checklist

### DIFF
--- a/cmd/cloud/dbt.go
+++ b/cmd/cloud/dbt.go
@@ -18,19 +18,10 @@ var (
 	dbtProjectPath string
 
 	DeployBundle = cloud.DeployBundle
+	DeleteBundle = cloud.DeleteBundle
 )
 
 const (
-	dbtDeployExample = `
-Specify the ID of the Deployment on Astronomer you would like to deploy this dbt project to:
-
-  $ astro dbt deploy <deployment ID>
-
-Menu will be presented if you do not specify a deployment ID:
-
-  $ astro dbt deploy
-`
-
 	dbtDefaultMountPathPrefix = "/usr/local/airflow/dbt/"
 	dbtProjectYmlFilename     = "dbt_project.yml"
 	dbtBundleType             = "dbt"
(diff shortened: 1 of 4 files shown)


cmd/cloud/dbt.go
```diff
@@ -119,19 +111,91 @@ func deployDbt(cmd *cobra.Command, args []string) error {
...
+
+  $ astro dbt delete <deployment ID>
+
+Menu will be presented if you do not specify a deployment ID:
+
+  $ astro dbt delete
+`,
+	}
+
+	cmd.Flags().StringVarP(&mountPath, "mount-path", "m", "", fmt.Sprintf("Path to mount dbt project in Airflow, for reference by DAGs. Default %s{dbt project name}", dbtDefaultMountPathPrefix))
```
neel-astro, line 143:
```suggestion
	cmd.Flags().StringVarP(&mountPath, "mount-path", "m", "", fmt.Sprintf("Mount path of the dbt project to be deleted from the Deployment. Default %s{dbt project name}", dbtDefaultMountPathPrefix))
```

cmd/cloud/dbt.go
```diff
@@ -119,19 +111,91 @@ func deployDbt(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return err
 	}
+	fmt.Println("Initiating dbt delete deploy for deployment ID: " + deploymentID)
 
-	return nil
+	// if the mount path is not provided, derive it from the dbt project name
+	if mountPath == "" {
+		// extract the dbt project's name
+		dbtProjectName, err := extractDbtProjectName(dbtProjectPath)
```
neel-astro, line 175: general questions: 
Do we feel not having any form of identifier for the deploy bundle makes it harder to maintain dbt projects for the users? In this particular case, the user needs to have the project at `dbtProjectPath` to delete the remote bundle (which feels a bit weird).

Also, otherwise, would users be able to figure out mountPath for each deploy bundle in an Airflow deployment, so that they can provide it in this command?

neel-astro (commented):
Left some general newbie questions, but the PR looks fine otherwise

cmd/cloud/dbt.go
```diff
@@ -119,19 +111,91 @@ func deployDbt(cmd *cobra.Command, args []string) error {
...
+
+  $ astro dbt delete
+`,
+	}
+
+	cmd.Flags().StringVarP(&mountPath, "mount-path", "m", "", fmt.Sprintf("Path to mount dbt project in Airflow, for reference by DAGs. Default %s{dbt project name}", dbtDefaultMountPathPrefix))
+	cmd.Flags().StringVarP(&dbtProjectPath, "project-path", "p", "", "Path to the dbt project to delete from the Deployment. Default current directory")
+	cmd.Flags().StringVar(&workspaceID, "workspace-id", "", "Workspace for your Deployment")
+	cmd.Flags().StringVarP(&deploymentName, "deployment-name", "n", "", "Name of the Deployment to deploy to")
+	cmd.Flags().StringVarP(&deployDescription, "description", "", "", "Description to store on the deploy")
```
kushalmalani, line 147: We should update descriptions to reflect this is a delete

cmd/cloud/dbt.go
```diff
@@ -119,19 +111,91 @@ func deployDbt(cmd *cobra.Command, args []string) error {
...
+	// if the workspace ID is not provided, try to find a valid workspace
+	if workspaceID == "" {
+		var err error
+		workspaceID, err = coalesceWorkspace()
+		if err != nil {
+			return fmt.Errorf("failed to find a valid workspace: %w", err)
+		}
+	}
+
+	// get the deployment id to deploy the dbt project to
```
kushalmalani, line 165:
```suggestion
	// get the deployment id to delete the dbt project
```

kushalmalani (approved):
Left a few nits on updating description. Looks good otherwise

cmd/cloud/dbt.go
jeremybeard: We're no longer adding unstructured information to the description that is already on the deploy object. The UI should understand that the deploy is a delete and show that to the user.

cmd/cloud/dbt.go
jeremybeard: If we generated an identifier I think that would have the same discoverability problem as the mount path. It might be a good idea for us to also have a CLI command that lists the dbt projects on the deployment, so that the user can see what is present and what mount path would be used to remove one (in case the dbt project itself is not locally accessible).

jeremybeard: I think in general having a low level thing like a mount path be the key would be a bit odd, but that is the actual value that users will have to plug into their DAGs to use the files. e.g. in the Cosmos `ProjectConfig` https://astronomer.github.io/astronomer-cosmos/getting_started/astro.html#create-a-dagfile

cmd/cloud/dbt.go
kushalmalani: oh sorry, i meant the description of the flag.

cmd/cloud/dbt.go
neel-astro: > It might be a good idea for us to also have a CLI command that lists the dbt projects on the deployment, so that the user can see what is present and what mount path would be used to remove one

True, that makes sense

> but that is the actual value that users will have to plug into their DAGs to use the files. e.g. in the Cosmos ProjectConfig https://astronomer.github.io/astronomer-cosmos/getting_started/astro.html#create-a-dagfile

ahh, got it, that's solves my concern to some extent at least. 🙂

neel-astro (approved):

PR #1694, opened 2024-07-31
Add skip logic to registry label to fix double entry error

Simpcyclassy:
## Description

This PR fixes double entries for deploy history by introducing a new registry label `io.astronomer.skip.revision`. When this label is present, true, and a dag only deploy, this would skip creating a revision from registry.

This flag would always be true for software. The decision to skip or not to skip is handled here [astronomer/houston-api/blob/main/src/lib/utilities/index.js#L686](https://github.com/astronomer/houston-api/blob/main/src/lib/utilities/index.js#L686)

## 🎟 Issue(s)

Related astronomer/issues#6480  astronomer/issues#6543

## Houston

Related astronomer/houston-api/pull/1783

## 🧪 Functional Testing

Tested on dev and confirmed duplicate history is not created for dag only deploys
image

## 📸 Screenshots

image

## 📋 Checklist


jeremybeard:
@Simpcyclassy can you describe how unconditionally hard-coding this label on every image solves the issue? Are there other scenarios where it could be false or not defined?

airflow/docker_image.go
```diff
@@ -64,6 +64,8 @@ func (d *DockerImage) Build(dockerfile, buildSecretString string, buildConfig ai
 		"-f",
 		dockerfile,
 		".",
+		"--label",
```
neel-astro, line 67: I would very much prefer this label to be only added if the user is connected to Software and deploying to Software if this change is only intended for Software.  Maybe we could allow passing label array to the function, and only the Software side of deploy logic sends `io.astronomer.skip.revision` as part of that, wdyt?

Simpcyclassy: I will make this improvement.

Simpcyclassy:
> @Simpcyclassy can you describe how unconditionally hard-coding this label on every image solves the issue? Are there other scenarios where it could be false or not defined?

@jeremybeard  this is needed for software going forward, and the value will always be true; it is just our way of identifying registry requests. I will go ahead and make it a software-only feature

airflow/docker_image.go
pgvishnuram: @neel-astro we should generalize this option similar to --build-args so that we can also expand this feature for astro cli to allow passing custom labels as well, and for software we either hardcode or expand this option with default skip label passed

neel-astro: I don't mind that either, just that if we are going that route, let's ensure existing functionality is intact and we refactor other existing args used in the method

neel-astro (commented):
minor nits, looks ok otherwise

airflow/docker_image.go
```diff
@@ -69,6 +69,10 @@ func (d *DockerImage) Build(dockerfile, buildSecretString string, buildConfig ai
 		args = append(args, "--no-cache")
 	}
 
+	if len(buildConfig.Labels) > 0 {
+		args = append(args, buildConfig.Labels...)
```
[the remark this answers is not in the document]
rujhan-arora-astronomer: Agreed with @neel-astro. The lib function should set --label, so that we follow DRY (for caller functions.)

rujhan-arora-astronomer (changes requested):

rujhan-arora-astronomer (approved):

Simpcyclassy:
Can I get help with merging this? It says to me, "You’re not authorized to merge this pull request."

rujhan-arora-astronomer:
> Can I get help with merging this? It says to me, "You’re not authorized to merge this pull request."

@neel-astro or @kushalmalani can help here.

airflow/docker_image.go
```diff
@@ -69,6 +69,12 @@ func (d *DockerImage) Build(dockerfile, buildSecretString string, buildConfig ai
 		args = append(args, "--no-cache")
 	}
 
+	if len(buildConfig.Labels) > 0 {
+		for _, label := range buildConfig.Labels {
+			args = append(args, "--label", label)
+		}
+	}
```
neel-astro, line 76:
```suggestion
	for _, label := range buildConfig.Labels {
		args = append(args, "--label", label)
	}
```
left over nit: we don't require the if clause now with the for loop

neel-astro (approved):
LGTM, apart from a minor nit

PR #1702, opened 2024-08-12
Fix error output

stuart23:
Fixes the following output where the `a` in the output is accidentally escaped:

image


neel-astro (approved):
🚀

kushalmalani (approved):

PR #1728, opened 2024-10-08
Add `--template` flag on CLI Project Initialization to setup template based project

pritt20:
## Description

This PR includes changes to add `--templates` flag to `astro dev init` command. This option will allow users to specify runtime templates name like `etl`, `dbt-on-astro`, generative-ai` etc and CLI will setup an astro project based on provided template by user. 

These changes are extension to the enhancements being made in onboarding-workflow where once user has completed the process of creating the deployment and now wants to make changes to deployed dags or deployment, we want them to explore astro-cli option. 

These changes will make it easy for users to modify/update dags as the template based project would setup all the required files which were auto-deployed during deployment create process as part of onboarding-workflow.

## 🎟 Issue(s)

Related https://github.com/astronomer/astro/issues/24343

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


airflow/airflow.go
```diff
@@ -118,14 +126,21 @@ func Init(path, airflowImageName, airflowImageTag string) error {
...
 	}
 
-	// Initailize directories
-	if err := initDirs(path, dirs); err != nil {
-		return errors.Wrap(err, "failed to create project directories")
-	}
+	if template != "" {
+		err := templateClient.DownloadAndExtractTemplate(astroTemplateRepoURL, template, path)
+		if err != nil {
+			return errors.Wrap(err, "failed to setup template based astro project")
```
lzdanski, line 132:
```suggestion
			return errors.Wrap(err, "failed to set up template-based astro project")
```

lzdanski: "setup" = a noun
"set up" = verb

English grammar quirk :/

cmd/airflow.go
```diff
@@ -150,6 +151,7 @@ func newAirflowInitCmd() *cobra.Command {
 	}
 	cmd.Flags().StringVarP(&projectName, "name", "n", "", "Name of Astro project")
 	cmd.Flags().StringVarP(&airflowVersion, "airflow-version", "a", "", "Version of Airflow you want to create an Astro project with. If not specified, latest is assumed. You can change this version in your Dockerfile at any time.")
+	cmd.Flags().StringVarP(&template, "template", "t", "", "Tempate name that you want to use to create the local astro project. Posisble values can be etl, dbt-on-astro, generative-ai and learning-airflow  ")
```
lzdanski, line 154:
```suggestion
	cmd.Flags().StringVarP(&template, "template", "t", "", "Tempate name that you want to use to create the local astro project. Possible values can be etl, dbt-on-astro, generative-ai, and learning-airflow  ")
```
Do you use any particular formatting to indicate something can be used as a possible value? we use code formatting or `" " ` to indicate strings in docs and code examples.

cmd/errors.go
```diff
@@ -14,4 +14,5 @@ var (
 
 	errInvalidSetArgs    = errors.New("must specify exactly two arguments (key value) when setting a config")
 	errInvalidConfigPath = errors.New("config does not exist, check your config key")
+	errInvalidTemplate   = errors.New("not a valid template name. It can only be etl, dbt-on-astro, generative-ai or learning-airflow")
```
lzdanski, line 17:
```suggestion
	errInvalidTemplate   = errors.New("not a valid template name. Template names can only be etl, dbt-on-astro, generative-ai or learning-airflow")
```

lzdanski (changes requested):
Found some grammar/spelling nits.

airflow/airflow.go
```diff
@@ -118,14 +126,21 @@ func Init(path, airflowImageName, airflowImageTag string) error {
 		".astro/dag_integrity_exceptions.txt":  "# Add dag files to exempt from parse test below. ex: dags/<test-file>",
 	}
 
-	// Initailize directories
-	if err := initDirs(path, dirs); err != nil {
-		return errors.Wrap(err, "failed to create project directories")
-	}
+	if template != "" {
+		err := templateClient.DownloadAndExtractTemplate(astroTemplateRepoURL, template, path)
```
neel-astro, line 130: nit: keeping the `templateClient` initialization only when `template != ""` might be a bit better

airflow/airflow_test.go
```diff
@@ -77,6 +77,33 @@ func (s *Suite) TestInit() {
 	}
 }
 
+func (s *Suite) TestTemplateInit() {
```
neel-astro, line 80: could we cover negative test cases as well

cmd/airflow.go
```diff
@@ -944,3 +952,7 @@ func prepareDefaultAirflowImageTag(airflowVersion string, httpClient *airflowver
 	}
 	return defaultImageTag
 }
+
+func isValidTemplate(template string) bool {
+	return template == "etl" || template == "dbt-on-astro" || template == "learning-airflow" || template == "generative-ai"
+}
```
neel-astro, line 958: nit: a possibly bit cleaner way might be: to declare the template variations as a list of strings variable globally (so that it could be reused in other places if required down the line), and check if the template is part of template variations.

i.e.:
```golang
var templateVariations = []string{"etl", "dbt-on-astro", "learning-airflow", "generative-ai"}

return slices.Contains(templateVariations, template)
```

airflow/airflow.go
```diff
@@ -7,9 +7,15 @@ import (
...
 
 	"github.com/astronomer/astro-cli/pkg/fileutil"
+	"github.com/astronomer/astro-cli/pkg/httputil"
+	runtimeTemplateClient "github.com/astronomer/astro-cli/runtime-template-client"
 	"github.com/pkg/errors"
 )
 
+const (
+	astroTemplateRepoURL = "https://github.com/astronomer/templates"
+)
```
neel-astro, line 17: this should be part of the `runtime-template-client` package

runtime-template-client/runtime-template-client.go
```diff
@@ -0,0 +1,197 @@
...
+	*httputil.HTTPClient
+}
+
+func NewHTTPAstroTemplateClient(client *httputil.HTTPClient) *HTTPAstroTemplateClient {
+	return &HTTPAstroTemplateClient{
+		client,
+	}
+}
+
+func (c *HTTPAstroTemplateClient) DownloadAndExtractTemplate(repoURL, templateDir, destDir string) error {
```
neel-astro, line 34: let's keep the `repoURL` as constant to this package, as the package is intended to be used only for this purpose

runtime-template-client/runtime-template-client.go
```diff
@@ -0,0 +1,197 @@
+package runtimetemplateclient
```
neel-astro, line 1: update the file name to `runtime_template_client.go`, I realised that we might not have followed it diligently, but generally underscore are preferred over hyphen (no a rule set in stone, but almost all the files in the project follows that as now)

runtime-template-client/runtime-template-client.go
```diff
@@ -0,0 +1,197 @@
...
+
+type Client interface {
+	DownloadAndExtractTemplate(repoURL, branch, templateDir, destDir string) error
+}
+
+type HTTPAstroTemplateClient struct {
+	*httputil.HTTPClient
+}
+
+func NewHTTPAstroTemplateClient(client *httputil.HTTPClient) *HTTPAstroTemplateClient {
```
neel-astro, line 28: Let's initialize the client here instead of letting the caller function create it so that the client is consistent across the repo for this package.

runtime-template-client/runtime-template-client.go
```diff
@@ -0,0 +1,197 @@
...
+		return fmt.Errorf("template directory %s not found", templateDir)
+	}
+	if err := copyDir(srcDir, destDir); err != nil {
+		return fmt.Errorf("failed to copy template directory: %w", err)
+	}
+
+	return nil
+}
+
+func extractTarGz(r io.Reader, dest, templateDir string) error {
```
neel-astro, line 77: could we add unit test for this file logic 🙂

airflow/airflow.go
```diff
@@ -118,14 +126,21 @@ func Init(path, airflowImageName, airflowImageTag string) error {
 		".astro/dag_integrity_exceptions.txt":  "# Add dag files to exempt from parse test below. ex: dags/<test-file>",
 	}
 
-	// Initailize directories
-	if err := initDirs(path, dirs); err != nil {
-		return errors.Wrap(err, "failed to create project directories")
-	}
+	if template != "" {
```
neel-astro, line 129: how would the custom runtime version work with this new template logic?

neel-astro (commented):
Left some comments, but great work for a first time 🚀

airflow/airflow.go
pritt20: Thanks @lzdanski ! I have made the required changes.

airflow/airflow_test.go
pritt20: I've included the negative test case now.

cmd/airflow.go
pritt20: Yeah, this make sense. I've updated the logic for this validation.

runtime-template-client/runtime-template-client.go
pritt20: Yes, I have included the tests for runtime-template-client

airflow/airflow.go
pritt20: Template based astro projects will always have latest runtime version as we are fetching these files from `astronomer/templates` repo. I have included a note in description that `airflow-version` flag will not be respected when using `--template` flag.

cmd/airflow_test.go
```diff
@@ -191,21 +198,16 @@ func (s *AirflowSuite) cleanUpInitFiles() {
...
+		".astro",
 		"./astro",
-		"tests/dags/test_dag_example.py",
-		"tests/dags",
 		"tests",
 		"dags",
 	}
 	for _, f := range files {
-		e := os.Remove(f)
+		e := os.RemoveAll(f)
```
jeremybeard, line 210: Is there some way we can clean this up in one call? Seems very brittle to list these out individually.

airflow/airflow.go
```diff
@@ -118,14 +119,21 @@ func Init(path, airflowImageName, airflowImageTag string) error {
...
```
jeremybeard, line 122: This placement is mixed in with the existing non-template init flow, which is a bit confusing. Can we add the `template != ""` check at the top of the function and have it return early so we don't do unnecessary things like construct the `files` map?

cmd/airflow.go
```diff
@@ -107,17 +110,18 @@ astro dev init --airflow-version 2.2.3
 	errPytestArgs          = errors.New("you can only pass one pytest file or directory")
 	buildSecrets           = []string{}
 	errNoCompose           = errors.New("cannot use '--compose-file' without '--compose' flag")
+	templateVariations     = []string{"etl", "dbt-on-astro", "learning-airflow", "generative-ai"}
```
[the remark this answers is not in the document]
schnie: I think this could makes sense - the less steps involved with adding a new template, the better. We could use that same list for validation right?

cmd/airflow.go
```diff
@@ -146,10 +150,13 @@ func newAirflowInitCmd() *cobra.Command {
...
 			return nil
 		},
-		RunE: airflowInit,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return airflowInit(cmd, args, runtimeTemplateClient)
+		},
 	}
 	cmd.Flags().StringVarP(&projectName, "name", "n", "", "Name of Astro project")
 	cmd.Flags().StringVarP(&airflowVersion, "airflow-version", "a", "", "Version of Airflow you want to create an Astro project with. If not specified, latest is assumed. You can change this version in your Dockerfile at any time.")
+	cmd.Flags().StringVarP(&template, "template", "t", "", "Tempate name that you want to use to create the local astro project. Possible values can be etl, dbt-on-astro, generative-ai and learning-airflow. Please note template based astro project use latest runtime version and airflow-version flag will not be respected when creating a project with template flag")
```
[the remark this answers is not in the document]
schnie: I think that makes sense, a little more clear.

runtime-template-client/runtime_template_client.go
```diff
@@ -0,0 +1,200 @@
+package runtimetemplateclient
```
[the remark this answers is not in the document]
schnie: This is something @pritt20 and I discussed. As far as I know, we don't currently list `git` as a prerequisite. In this particular instance, we're trying to get the user up and running with the least amount of dependencies and prerequisites as possible. Even though most of our user base will probably have `git` installed, it just felt too risky to rely on for such a critical onboarding step where we're doing everything we can to optimize user experience and remove friction. 

Are there other places in the CLI code where `git` is currently a hard dependency?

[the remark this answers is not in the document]
jeremybeard: Yeah that is true, we do currently use `git` in a few places but not as a hard dependency. If we go with the tar/gzip approach can we follow a similar structure to what I listed above, so we do not need a new top-level client? Where a top-level variable can be replaced by a unit test to simulate the file download. We do this in a few other places, e.g. for [deploying an image](https://github.com/astronomer/astro-cli/blob/40db395aa829e50a4dc0781d626365fd80c23da6/cmd/cloud/deploy.go#L41) which a unit test cannot really do.

I'm also hoping we can extract the contents straight into the target dir instead of into a temp dir and then copying the dirs and files across?

cmd/airflow.go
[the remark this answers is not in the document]
pritt20: Thanks @jeremybeard! that makes sense. 

Would it be okay if we make template options interactively available to users that they can choose from? I tried to implement this and named this flag as `--select-template` ( we can change this if we find better name). `--select-template` is boolean flag and if passed, user will be provided with all available template options that user can then select from and CLI will create selected template based project. Please find below example:

```
astro dev init --select-template
 #     TEMPLATE             
 1     dbt-on-astro         
 2     learning-airflow     
 3     generative-ai        
 4     etl                  

> 2

Initializing Astro project
Pulling Airflow development files from Astro Runtime 12.2.0
/Users/pritt/Astronomer/plg/astro-cli/airflow-test-2 
You are not in an empty directory. Are you sure you want to initialize a project? (y/n) y
Reinitialized existing Astro project in /Users/pritt/Astronomer/plg/astro-cli/airflow-test-2
```

cmd/airflow.go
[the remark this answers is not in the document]
pritt20: Renamed the flag to `--from-template`

runtime-template-client/runtime_template_client.go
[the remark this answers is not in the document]
pritt20: Thanks for the suggestion @jeremybeard and @schnie.

I have updated the code to reflect above changes.

I tried to dump the extracted files to target directory but the tarball that is extracted contains all templates directories and use path like `target-directory/astronomer-templates-7a8aa17/<templates-directories>`. We want the selected template directory files to be in `target-directory` without any sub directories. Therefore I have kept the temp directory logic intact to move files around and copy only required files to target directory 😅

airflow/runtime_templates.go
```diff
@@ -0,0 +1,273 @@
+package airflow
```
neel-astro, line 1: I feel like we are missing unit tests for all this new logic, could we add that please, it's more so important because the logic around extract and copy is not so straight forward.

cmd/airflow.go
```diff
@@ -150,6 +157,8 @@ func newAirflowInitCmd() *cobra.Command {
 	}
 	cmd.Flags().StringVarP(&projectName, "name", "n", "", "Name of Astro project")
 	cmd.Flags().StringVarP(&airflowVersion, "airflow-version", "a", "", "Version of Airflow you want to create an Astro project with. If not specified, latest is assumed. You can change this version in your Dockerfile at any time.")
+	cmd.Flags().StringVarP(&fromTemplate, "from-template", "t", "", "Template name that you want to use to create the local astro project. Possible values can be etl, dbt-on-astro, generative-ai and learning-airflow. Please note template based astro project use latest runtime version and airflow-version flag will be ignored when creating a project with template flag")
```
neel-astro, line 160: maybe either we don't specify the option in the description if that is subject to change, or we call the `airflow.FetchTemplateList` as part of the `newAirflowInitCmd` itself so that the description could be dynamic based on response, wdyt?

cmd/airflow_test.go
neel-astro: maybe create and run the tests in a temp dir, and delete the temp dir itself at the end of each test as part of `TearDownSubTest`

cmd/airflow.go
```diff
@@ -150,6 +157,8 @@ func newAirflowInitCmd() *cobra.Command {
...
+	cmd.Flags().BoolVarP(&selectTemplate, "select-template", "s", false, "Provides a list of templates to select from and create the local astro project based on selected template")
```
jeremybeard, line 161: I quite like this UX, the user is presented with the valid choices and picks the one they want. Can we just make this the only way to create a project from a template? It could even be what `--from-template` does and then we don't need `--select-template`. We shouldn't need a non-interactive variant because this isn't going to be called by scripts/CI/etc.

.mockery.yaml
```diff
@@ -46,6 +46,11 @@ packages:
       ClientWithResponsesInterface:
         config:
           filename: client.go
+  github.com/astronomer/astro-cli/runtime-template-client:
```
jeremybeard, line 49: Not used anymore?

airflow/runtime_templates.go
```diff
@@ -0,0 +1,273 @@
...
+	gr, err := gzip.NewReader(r)
+	if err != nil {
+		return fmt.Errorf("failed to create gzip reader: %w", err)
+	}
+	defer gr.Close()
+
+	tarReader := tar.NewReader(gr)
+	var baseDir string
+
+	limitTarReader := io.LimitReader(tarReader, maxExtractSize)
```
jeremybeard, line 163: Do we need this?

airflow/runtime_templates.go
```diff
@@ -0,0 +1,273 @@
...
+
+		// Extract the base directory from the first valid header
+		if baseDir == "" {
+			parts := strings.Split(header.Name, "/")
+			if len(parts) > 1 {
+				baseDir = parts[0]
+			}
+		}
+
+		// Skip files that are not part of the desired template directory
```
jeremybeard, line 187: If we're already only extracting the files of the desired template, why do we need to extract it to a temp dir and then copy it all over to the final dir, instead of just extracting to the final dir?

airflow/runtime_templates.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	defer in.Close()
+
+	out, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	if _, err = io.Copy(out, in); err != nil {
```
jeremybeard, line 269: I've hoping we don't need to do this copy, but if we do I'd expect this should just be a rename not a literal copy.

.mockery.yaml
pritt20: I have cleaned this unused code as we moved away from initial approach of creating top level client.

cmd/airflow.go
pritt20: Hey @jeremybeard ,

The initial Idea was to provide users with simple set of commands that they can execute to set up template based astro project. Eg: `astro dev init --from-template etl` 

If we keep only `--select-template` flag functionality then we will have to guide users in our onboarding workflow to select the template and set up astro project accordingly. 

CC: @schnie

cmd/airflow.go
pritt20: Thanks @neel-astro ! I've updated the description of flag to remove the template names

airflow/runtime_templates.go
pritt20: Updated the required tests now.

cmd/airflow.go
schnie: @pritt20 we could use the same flag _if_ we used the `=` format right? 

So `astro dev init --from-template` would be interactive, and
`astro dev init --from-template=etl` would work for the one-shot version?

pritt20: yeah, this format can be implemented to use same flag for interactive and for one-shot version. Its just that we don't really use this format anywhere in CLI for existing commands😅

schnie: Yea that does make it a little awkward I suppose. The main use-case for the one-shot style will be to copy and paste from the UI, so maybe that won't matter too much. 🤔

schnie: I kind of lean towards using a single flag and just accept that we need to show the `=` in the commands we display to copy/paste.

airflow/runtime_templates.go
pritt20: Updated the code to use tarReader and have flagged it with `nolint` comment.

cmd/airflow_test.go
pritt20: Updated all tests to make use of temp directory (which will then be removed after tests are executed) instead of `cmd` directory

cmd/airflow.go
pritt20: Hey @jeremybeard @neel-astro, 

Could you please let us know your thoughts on above discussion. If we want to use single flag and use it both ways, shall we go ahead with the format `astro dev init --from-template=etl` ?

neel-astro: From a CLI UX perspective, I might be a bit concerned about changing the behavior of a single flag in the CLI, where a user could only pass a flag value with `=`.
From what I understand based on comments if passing a template value in the command isn't going to be a major form of usage (because the user has to copy/paste the value from somewhere), then should we just skip that and always present an interactive mode? . `astro dev init` does not have a CI-CD use case, so we don't necessarily need to support a non-interactive mode.

cmd/airflow_test.go
```diff
@@ -224,7 +222,77 @@ func (s *AirflowSuite) mockUserInput(i string) (r, stdin *os.File) {
...
+		err := airflowInit(cmd, args)
+		s.NoError(err)
+
+		b, _ := os.ReadFile(filepath.Join(s.tempDir, "requirements.txt"))
+		fileContents := string(b)
+		s.True(strings.Contains(fileContents, "test requirements file"))
+	})
+
 	s.Run("success", func() {
+		config.WorkingPath = s.tempDir
```
neel-astro, line 295: this could be part of setup, wdyt?

cmd/airflow_test.go
```diff
@@ -155,59 +168,44 @@ func (s *AirflowSuite) TestNewAirflowUpgradeCheckCmd() {
...
+		// assert contents of Dockerfile
+		b, _ := os.ReadFile(filepath.Join(s.tempDir, "Dockerfile"))
+		dockerfileContents := string(b)
+		s.True(strings.Contains(dockerfileContents, "FROM quay.io/astronomer/astro-runtime:"))
+	})
 }
 
 func (s *AirflowSuite) cleanUpInitFiles() {
 	s.T().Helper()
-	files := []string{
```
neel-astro, line 186: thanks for fixing this 🙇‍♂️

cmd/airflow.go
schnie: Our next steps for the onboarding flow are to show the user a small block of commands to copy and execute in their terminal. I think we want a non-interactive way to do this to reduce the keystrokes (minor, I know) and steps involved to get setup. It'd be great to just paste in the terminal and go with no more interactions. 

Given that, we have a choice to either piggyback on the same flag (`--from-template`), or add a separate, but similar command. The latter feels a little awkward and a little confusing. For that reason, I lean towards just using the `=` format for this command alone right now, but keep our existing pattern for other flags. We can treat this as a special case for onboarding purposes. 

I imagine most users exploring the CLI on their own would run the command interactively. For our onboarding use-case we want to specify the exact template to use, so we'd provide them a command to run that matches the template they selected when going through the onboarding flow, like `astro dev init --from-template=generative-ai`.

Maybe we just document the bare, interactive `--from-template` mode and reserve the `=` format (`--from-template=generative-ai` just for the case where we give them the command to copy/paste.

jeremybeard: > Maybe we just document the bare, interactive --from-template mode and reserve the = format (--from-template=generative-ai just for the case where we give them the command to copy/paste.

That sounds good to me.

neel-astro: > Our next steps for the onboarding flow are to show the user a small block of commands to copy and execute in their terminal.

Cool, I was missing this context, I am on board with that approach 👍

pritt20: Thanks everyone! I've updated the PR to reflect above discussed changes.

cmd/airflow_test.go
pritt20: Thanks @neel-astro ! I've updated the tests now.

airflow/runtime_templates_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/astronomer/astro-cli/pkg/fileutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFetchTemplateList_Success(t *testing.T) {
```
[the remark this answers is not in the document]
pritt20: Yeah, that make sense! I've updated the tests to use test suit to keep these in line with existing tests.

cmd/airflow.go
```diff
@@ -509,6 +517,24 @@ func airflowInit(cmd *cobra.Command, args []string) error {
...
+
+	if fromTemplate != "" {
+		templateList, err := TemplateList()
+		if err != nil {
+			return fmt.Errorf("unable to fetch template list: %w", err)
+		}
+		if !isValidTemplate(templateList, fromTemplate) {
+			return fmt.Errorf("%s is not a valid template name. Available templates are: %s", fromTemplate, templateList)
+		}
+	}
```
neel-astro, line 536:
```suggestion
	if fromTemplate == "select-template" {
		selectedTemplate, err := selectedTemplate()
		if err != nil {
			return fmt.Errorf("unable to select template from list: %w", err)
		}
		fromTemplate = selectedTemplate
	} else if fromTemplate != "" {
		templateList, err := TemplateList()
		if err != nil {
			return fmt.Errorf("unable to fetch template list: %w", err)
		}
		if !isValidTemplate(templateList, fromTemplate) {
			return fmt.Errorf("%s is not a valid template name. Available templates are: %s", fromTemplate, templateList)
		}
	}
```
minor nit: we don't need to re-fetch the list and validate the selection from the user

cmd/airflow_test.go
```diff
@@ -224,6 +217,72 @@ func (s *AirflowSuite) mockUserInput(i string) (r, stdin *os.File) {
 }
 
 func (s *AirflowSuite) TestAirflowInit() {
+	TemplateList = func() ([]string, error) { return []string{"A", "B", "C", "D"}, nil }
+	s.Run("initialize template based project via select-template flag", func() {
+		cmd := newAirflowInitCmd()
+		cmd.Flag("name").Value.Set("test-project-name")
+		cmd.Flag("from-template").Value.Set("")
```
neel-astro, line 224: nit: I don't feel like this test case is actually triggering the template selection code path, could we add a bit more robust check to validate that some/any files are correctly created based on the template selection value. Maybe just set it to "select-template". 

Saying it based on the codecov report 😅 (https://app.codecov.io/gh/astronomer/astro-cli/pull/1728)

neel-astro (approved):
Left some nitpicks to squash, but otherwise LGTM

cmd/airflow.go
pritt20: Yes, that make sense. Updated the code block.

jeremybeard (approved):
Looks good! Thanks for working through all the feedback. I left a nit comment but approving ahead of that

schnie:
Blocking feedback request was handled and approved async from Github. Administrative merge for that reason.

lzdanski:
Docs update required: https://www.astronomer.io/docs/astro/cli/astro-dev-init

Issue #1407, opened 2023-10-02
Always `--pull` when building image


jedcunningham:
- [x] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

We'd like to start offering floating tags for Runtime (e.g. a `astro-runtime:9`) which will always point to the latest release in a given major. One challenge, however, is to make sure the latest image is used when building the image. 

**🧩 Describe the solution you'd like**

Add `--pull` to the build command.

**🤔 Describe alternatives you've considered**

We could alternatively document customers need to do an explicit pull before an `astro deploy` etc, but that seems easy to forget and not very polished. One extra request to make sure the image is up to date is pretty cheap (locally, typically .1 second or so).

**Is your feature request specific to a particular Astronomer Platform?**

- [ ] Astro
- [ ] Software
- [x] None/Unknown

cmarteepants:
@vishwas-astro @nicholasfilippi - I see this is in icebox. Is this something we can prioritize please? This is a blocker for us publishing runtime images with floating tags for major runtime versions. 

If someone uses this batch of images, patch and minor runtime updates will happen automatically. If we don't change the pull method when building, docker will not check whether or not we have published a more recent version and will default to whatever is in the cache - meaning no automatic update.

sunkickr:
does this fix the issue? https://github.com/astronomer/astro-cli/pull/1415

jedcunningham:
Yes it does.

PR #1731, opened 2024-10-10
pull runtime images on build

melugoyal:
## Description

when building images with the CLI, if it's an astro-runtime image, instruct `docker build` to pull the image first. this will allow is to start using floating tags for astro-runtime going forward.

## 🎟 Issue(s)

Closes https://github.com/astronomer/astro-cli/issues/1407

this was implemented once and reverted, due to incompatibility with customers' pipelines using their own images that are not public. liming this to runtime images (which are public) solves that.

## 🧪 Functional Testing

unit testing covers the cases where `--pull` should and should not be added.

## 📸 Screenshots

unit tests passing is sufficient

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -1,6 +1,7 @@
 package airflow
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


airflow/docker_image.go
```diff
@@ -48,23 +49,52 @@ func DockerImageInit(image string) *DockerImage {
...
+	if addPullFlag {
+		args = append(args, "--pull")
+	}
+	err = os.Chdir(buildConfig.Path)
 	if err != nil {
 		return err
 	}
-	args := []string{
+	args = append(args, []string{
 		"build",
```
jeremybeard, line 91: `build` is repeated which fails the command.

airflow/docker_image.go
```diff
@@ -48,23 +49,52 @@ func DockerImageInit(image string) *DockerImage {
...
+	if err != nil {
+		return false, fmt.Errorf("opening file: %w", err)
+	}
+	defer file.Close()
+
+	// scan the Dockerfile to check if any FROM instructions reference an image other than astro runtime
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "FROM ") && !strings.HasPrefix(line, fmt.Sprintf("FROM %s:", FullAstroRuntimeImageName)) {
```
jeremybeard, line 63: Could we just rely on `astro-runtime` so we are not tied to a specific image registry?

airflow/docker_image_test.go
```diff
@@ -63,6 +63,52 @@ func (s *Suite) TestDockerImageBuild() {
 		s.NoError(err)
 	})
 
+	s.Run("build with no --pull", func() {
+		dockerfileContent := "FROM nginx"
+		dockerfile, _ := os.CreateTemp(os.TempDir(), "temp-Dockerfile")
+		defer func() {
```
jeremybeard, line 69: nit: can just use `defer os.Remove(dockerfile.Name())`

airflow/docker_image.go
melugoyal: resolved offline

jeremybeard (approved):

PR #1740, opened 2024-11-01
Adds Runtime Agnostic Healthcheck

schnie:
## Description

This PR aims to provide a consistent local dev startup experience across container runtimes. Today, `astro dev start` will print the status update and attempt to open the browser to the airflow dashboard once its up and running. This is nice and it works for Docker Desktop. It does not work as expected for `orbstack` or `podman`. For these, the command line hangs and we end up opening the browser to a URL that isn't ready yet. This is a bad experience. 

We're exploring the idea of using `podman` as our default runtime going forward and this is another small fix towards preparing for something like that. This fix will improve todays UX either way.

## 🎟 Issue(s)

Related to https://github.com/astronomer/astro/issues/24344

## 🧪 Functional Testing

Have tested this using `astro dev start` on:
- Mac OS, Docker Desktop
- Mac OS, Podman
- Mac OS, Orbstack
- Windows OS, Podman
- Windows OS, Docker Desktop

## 📸 Screenshots

## 📋 Checklist

- [x] ~Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary).~
- [x] ~Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary).~


airflow/docker.go
```diff
@@ -301,10 +299,23 @@ func (d *DockerCompose) Start(imageName, settingsFile, composeFile, buildSecretS
...
+
+	// Check the health of the webserver, up to the timeout.
+	// If we fail to get a 200 status code, we'll return an error message.
+	err = checkWebserverHealth(healthUrl, startupTimeout)
+	if err != nil {
+		return err
+	}
+
+	// If we've successfully gotten a healthcheck response, print the status.
+	err = printStatus(settingsFile, envConns, project, d.composeService, airflowDockerVersion, noBrowser)
```
schnie, line 314: `printStatus()` was previously called by the original `checkWebserverHealth()` function. That function has been removed and replaced and doesn't concern itself with printing the message. That now just happens at the "top-level" here.

airflow/docker.go
```diff
@@ -1416,11 +1384,7 @@ func printStatus(settingsFile string, envConns map[string]astrocore.EnvironmentO
 			}
 		}
 	}
-	if config.CFG.DockerCommand.GetString() == podman {
-		fmt.Println("\nComponents will be available soon. If they are not running in the next few minutes, run 'astro dev logs --webserver | --scheduler' for details.")
-	} else {
-		fmt.Println("\nProject is running! All components are now available.")
-	}
+	fmt.Println("\nProject is running! All components are now available.")
```
schnie, line 1387: The healthcheck now works for all runtimes so there's no need to differentiate here.

airflow/suite_test.go
```diff
@@ -19,7 +17,7 @@ type Suite struct {
 	origCmdExec              func(cmd string, stdout, stderr io.Writer, args ...string) error
 	origGetDockerClient      func() (client.APIClient, error)
 	origInitSettings         func(id, settingsFile string, envConns map[string]astrocore.EnvironmentObjectConnection, version uint64, connections, variables, pools bool) error
-	origCheckWebserverHealth func(settingsFile string, envConns map[string]astrocore.EnvironmentObjectConnection, project *types.Project, composeService api.Service, airflowDockerVersion uint64, noBrowser bool, timeout time.Duration) error
+	origCheckWebserverHealth func(url string, timeout time.Duration) error
```
schnie, line 20: Just updating the signature.

airflow/docker.go
```diff
@@ -82,7 +81,6 @@ const (
 var (
 	errNoFile                  = errors.New("file specified does not exist")
 	errSettingsPath            = "error looking for settings.yaml"
-	errComposeProjectRunning   = errors.New("project is up and running")
```
schnie, line 85: This was an error that we used to indicate success. We don't do that anymore.

airflow/docker.go
```diff
@@ -1461,23 +1406,6 @@ var CheckTriggererEnabled = func(imageLabels map[string]string) (bool, error) {
...
-		return false
-	}
-	runtimeVersion, ok := imageLabels[runtimeVersionLabelName]
-	if !ok {
-		// cannot determine runtime version print message by default
-		return true
-	}
-
-	return versions.LessThanOrEqualTo(runtimeVersion, M1ImageRuntimeVersion)
-}
```
schnie, line 1479: This function is purely used to decide if the `astro dev start --wait` flag defaults to `1m` or `5m`. Thinking we just simplify and use `5m`. The CLI progresses as soon as it sees the `200` from the webserver `/health` endpoint.

airflow/docker.go
```diff
@@ -275,36 +270,31 @@ func (d *DockerCompose) Start(imageName, settingsFile, composeFile, buildSecretS
...
-	startupTimeout = waitTime
-	// check if user provided a waitTime
-	// default is 1 minute
-	if waitTime != 1*time.Minute {
-		startupTimeout = waitTime
-	} else if CheckM1Image(imageLabels) {
-		// user did not provide a waitTime
-		// if running darwin/M1 architecture
-		// we wait for a longer startup time
-		startupTimeout = 5 * time.Minute
```
schnie, line 301: I'm removing a chunk of logic around this M1 architecture check and runtime comparison. The only reason we have all this is to bump the startup timeout from `1m` to `5m`. I've just made `5m` the default across the board.

schnie: If we feel strongly about _not_ changing the default across the board, we could revert this piece. It just feels like unnecessary cruft at this point, and this feels like a fine time to remove it.

airflow/docker.go
```diff
@@ -1349,49 +1339,6 @@ var createDockerProject = func(projectName, airflowHome, envFile, buildImage, se
...
-				return fmt.Errorf("there might be a problem with your project starting up. The webserver health check timed out after %s but your project will continue trying to start. Run 'astro dev logs --webserver | --scheduler' for details.\n\nTry again or use the --wait flag to increase the time out", timeout)
-			}
-			if !errors.Is(err, errComposeProjectRunning) {
-				return err
-			}
-		}
-	}
-	return nil
-}
-
```
schnie, line 1394: This implementation is being removed in favor of the new one declared in its own file below - `health_check.go`.

airflow/docker.go
jeremybeard: Good call.

cmd/airflow_test.go
```diff
@@ -407,7 +407,7 @@ func (s *AirflowSuite) TestAirflowStart() {
 
 		mockContainerHandler := new(mocks.ContainerHandler)
 		containerHandlerInit = func(airflowHome, envFile, dockerfile, imageName string) (airflow.ContainerHandler, error) {
-			mockContainerHandler.On("Start", "", "airflow_settings.yaml", "", "", false, false, 1*time.Minute, map[string]astrocore.EnvironmentObjectConnection(nil)).Return(nil).Once()
+			mockContainerHandler.On("Start", "", "airflow_settings.yaml", "", "", false, false, 5*time.Minute, map[string]astrocore.EnvironmentObjectConnection(nil)).Return(nil).Once()
```
jeremybeard, line 410: Ideally unit tests should use the constants.

jeremybeard (approved):
This is excellent, it simplifies the code and gains more functionality as a result! Appreciate the proactive PR comments too, makes it all very clear.

Approving ahead of a couple of small comments to look into.

airflow/health_check_test.go
```diff
@@ -0,0 +1,108 @@
...
+	// Create a mock server that returns a 200 status
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer server.Close()
+
+	err := checkWebserverHealth(server.URL, 5*time.Second)
+	if err != nil {
+		t.Errorf("Expected no error, got: %v", err)
+	}
```
neel-astro, line 21: nit: you could ignore it if you follow the current style
```suggestion
	assert.NoError(t, err)
```

cmd/airflow.go
```diff
@@ -107,6 +107,7 @@ astro dev init --airflow-version 2.2.3
 	errPytestArgs          = errors.New("you can only pass one pytest file or directory")
 	buildSecrets           = []string{}
 	errNoCompose           = errors.New("cannot use '--compose-file' without '--compose' flag")
+	defaultWaitTime        = 5 * time.Minute
```
neel-astro, line 110: I think the default should still be a minute or at best 2 minutes, the reason behind not having 5 minutes timeout earlier (and only restricting that to M1 with runtime version less than 6.x) was to have a faster response cycle back to the customer if things go wrong. I don't see a reason for increasing it based on the PR description, so maybe I missing some context.

neel-astro (changes requested):
This is a step in the right direction, but I am slightly concerned about potentially breaking the behavior for folks using docker-compose override or customer docker-compose in their local setup (applies to both docker and podman users, and I have seen a decent amount of examples where we suggest customers use it to customize any change), because in those cases we can't for sure say that the port mapping would be 8080 or even the same as that of the webserver.port in config (because changing port number is quite a common occurrence).
We need to think through a way to keep supporting those. Maybe adjust our error messaging to ask the user to set the webserver port in the config even if they are using custom docker-compose or override.

airflow/docker.go
```diff
@@ -275,36 +270,31 @@ func (d *DockerCompose) Start(imageName, settingsFile, composeFile, buildSecretS
...
-	if waitTime != 1*time.Minute {
-		startupTimeout = waitTime
-	} else if CheckM1Image(imageLabels) {
-		// user did not provide a waitTime
-		// if running darwin/M1 architecture
-		// we wait for a longer startup time
-		startupTimeout = 5 * time.Minute
+	// Airflow webserver should be hosted at localhost
+	// from the perspective of the CLI running on the host machine.
+	healthURL := "http://localhost:8080/health"
```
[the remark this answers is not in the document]
schnie: Ah yes, I forgot to change this. I'll update it.

cmd/airflow.go
schnie: That would be fine too. I would be fine changing it back to 1, but figured we could take this opportunity to simplify things. Having a different timeout based on the host architecture and the runtime version feels a little awkward I think. I also noticed that the runtime version we're checking for with this M1 logic is no longer under LTS support, so it seemed like we might be able to clean this up a bit. What do you think @neel-astro @jeremybeard?

airflow/docker.go
schnie: Yea, I'm fine with making it 1m across the board now. @neel-astro that sound ok to you?

cmd/airflow.go
neel-astro: I agree on cleaning up the bits we added around M1, but we bumped the timings to 5 minutes for M1 as an exception until 6.x, when runtime added native arm support. So I might favor simplifying things but keeping the default timing to 1 minute for all.

airflow/docker.go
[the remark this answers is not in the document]
schnie: This is done.

cmd/airflow.go
schnie: This is done.

schnie:
@neel-astro I've updated the healthcheck URL to get the port from the config now. I also attempted to override compose with:
```
webserver:
  ports:
  - 9090:9090
```

This results in both our standard 8080 _and_ 9090 being bound, so the override file can't wipe out our existing port and it'll always be the first in the list of ports (or the overridden port from the config). 

Also, the healthcheck that was defined previously via the compose file had 8080 hard-coded, so this should actually result in an improvement here now that we grab it from the config.

airflow/health_check_test.go
schnie: Fixed the full test suite to use the existing patterns.

neel-astro (approved):
LGTM, this change would still be slightly disruptive for anyone using `astro dev start --compose-file <custom-compose-file>` with a different webserver port, but there is a workaround for them by setting `astro config set webserver.port <webserver-port>`. But since this solves the problem for a bunch of other broken use cases, I am onboard with the change.

schnie:
@neel-astro do you have an example of what that compose file would look like to override the websever port? As far as I know, any ports defined in a custom compose file are _added_ to the standard definition we provide as the base, so 8080 (or whatever the user configures) will always be bound. 

Or are you referring to overriding the airflow setting to change the port that `airflow webserver` listens on?

neel-astro:
> @neel-astro do you have an example of what that compose file would look like to override the webserver port in a problematic way? As far as I know, any ports defined in a custom compose file are added to the standard definition we provide as the base, so 8080 (or whatever the user configures) will always be bound.

@schnie there is a functionality within `astro dev` commands to export the current docker-compose file locally, edit it, and use that local file to spin components up. This is a bit advanced use case, and it is also a backdoor for us to unblock users running into various non-common issues due to how they are running local setups.
`astro dev object export --compose --compose-file <compose-file>` to export the compose file
`astro dev start --compose-file <compose-file>` to use the local compose file to spin up components.

So theoretically, users could export the compose file, edit it to replace the existing 8080 webserver port, and spin up the local setup.

schnie:
Hmm, yea I was able to recreate that here locally. Does seem like a bit of an edge case.

I did have some code I was playing with that actually grabbed the port from the webserver service in the compose project. `ports` is a list, so if someone is really messing with things at this level they could also have multiple ports listed, in which case the port to healthcheck on becomes ambiguous.. unless we scanned all of them, but that doesn't feel quite right either.

Do you know of any use-case where someone out there would be doing this kind of change? If we know of someone actually doing this on the webserver maybe we think about it some more. Otherwise, it does feel like a lot for someone to go through for something like this.

neel-astro:
Yeah, I think it's fine because in such cases the new way to handle the case would be for them to set the `webserver.port` in CLI config.
The other way around might be for CLI to fetch the port from the container named "webserver", i.e. parse and get it from the `docker port webserver` output.
I have seen use cases for custom compose file used by the users (mostly around volume mounting), but I am not sure if they would be using that with a custom webserver port or not.

PR #1748, opened 2024-11-22
Refactor Docker Init with `ContainerRuntime` Interface

schnie:
## Description

Functionally, this really doesn't change much of anything. It is really just refactoring the `startDocker()` command to a new place. We've defined a new `ContainerRuntime` interface that will grow a bit in the next PR to cover container runtime lifecycle management functions. This places the original, directly in-line code into this new lifecycle management structure. We then call these functions from cobra pre-run hooks on our `astro dev` commands.

My original podman PR grew too large, so this is part 1 to get some of the basic structure in place for expansion, while not changing any behavior quite yet. A follow up PR will expand on the `ContainerRuntime` interface and add podman support.

## 🎟 Issue(s)

Related to https://github.com/astronomer/astro/issues/24344

## 🧪 Functional Testing

- I've updated a few unit tests, refactored some, and added a few new for the new code paths.
- All `astro dev` subcommands have been tested manually and confirmed to be working as expected. (we should automate this as one of our next steps)

## 📸 Screenshots

## 📋 Checklist


airflow/runtimes/podman_runtime.go
```diff
@@ -0,0 +1,7 @@
+package runtimes
+
+type PodmanRuntime struct{}
+
+func (p PodmanRuntime) Initialize() error {
+	return nil
+}
```
schnie, line 7: Coming in the next PR.

airflow/runtimes/docker.go
```diff
@@ -0,0 +1,76 @@
...
+			return fmt.Errorf(timeoutErrMsg)
+		case <-ticker.C:
+			_, err := d.CheckDockerCmd()
+			if err != nil {
+				continue
+			}
+			return nil
+		}
+	}
+}
```
schnie, line 76: This function is a refactored, more testable version of the previous `startDocker()` and `waitForDocker()` functions that we've deleted above.

airflow/docker_test.go
```diff
@@ -1694,46 +1694,6 @@ func (s *Suite) TestDockerComposeRunDAG() {
...
-				return errExecMock
-			default:
-				return errExecMock
-			}
-		}
-		err := startDocker()
-		s.Contains(err.Error(), "timed out waiting for docker")
-	})
-}
-
```
schnie, line 1736: Refactored and replaced below.

airflow/docker.go
```diff
@@ -1416,57 +1404,3 @@ func checkServiceState(serviceState, expectedState string) bool {
...
-			buf.Reset()
-			err := cmdExec(containerRuntime, buf, buf, "ps")
-			if err != nil {
-				continue
-			} else {
-				return nil
-			}
-		}
-	}
-}
```
schnie, line 1472: This block of code (`startDocker()`) was previously called directly by the `astro dev start` handler. It's now been removed and ultimately called from our new hooks - `EnsureRuntime`. This hook then ultimately calls the docker initialization code which now does this setup work. The refactored version is below.

airflow/runtimes/container_runtime.go
```diff
@@ -1,17 +1,59 @@
...
+	containerRuntimeInitMessage = " Astro uses container technology to run your Airflow project. " +
+		"Please wait while we get things started…"
+	spinnerRefresh = 100 * time.Millisecond
+)
+
+// ContainerRuntime interface defines the methods that manage
+// the container runtime lifecycle.
+type ContainerRuntime interface {
+	Initialize() error
+}
```
schnie, line 35: This interface will grow to include more lifecycle functions when we follow up with podman support.

cmd/airflow.go
```diff
@@ -122,6 +124,11 @@ func newDevRootCmd(platformCoreClient astroplatformcore.CoreClient, astroCoreCli
 		Aliases: []string{"d"},
 		Short:   "Run your Astro project in a local Airflow environment",
 		Long:    "Run an Apache Airflow environment on your local machine to test your project, including DAGs, Python Packages, and plugins.",
+		// Most astro dev sub-commands require the container runtime,
+		// so we set that configuration in this persistent pre-run hook.
+		// A few sub-commands don't require this, so they explicitly
+		// clobber it with a no-op function.
+		PersistentPreRunE: ConfigureContainerRuntime,
```
schnie, line 131: Here we insert a new `PersistentPreRunE` function to setup the runtime for all the sub commands. This is a quick function and doesn't actually run any initialization steps that take time. When needed, the value set from this function is referenced and we run any lifecycle hooks that interact with the runtime.

With this PR, we're only running this on the `astro dev start` command, as the other commands explicitly clobber this hook. In the following PR for podman support, we'll utilize this much more. 

The methods that are not yet using this functionality explicitly define their own `PersistentPreRunE` function of `DoNothing`. This was already happening previously with an inline function definiton. This is really just cleaning that up and defining all our hooks in a single place.

cmd/airflow.go
```diff
@@ -144,16 +151,13 @@ func newDevRootCmd(platformCoreClient astroplatformcore.CoreClient, astroCoreCli
...
 	cmd := &cobra.Command{
-		Use:     "init",
-		Short:   "Create a new Astro project in your working directory",
-		Long:    "Create a new Astro project in your working directory. This generates the files you need to start an Airflow environment on your local machine and deploy your project to a Deployment on Astro or Astronomer Software.",
-		Example: initCloudExample,
-		Args:    cobra.MaximumNArgs(1),
-		// ignore PersistentPreRunE of root command
-		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			return nil
-		},
```
schnie, line 155: All of these blocks are have been replaced with the same function - `DoNothing`.

airflow/runtimes/command.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Execute runs the Podman command and returns the output.
+func (p *Command) Execute() (string, error) {
+	cmd := exec.Command(p.Command, p.Args...) //nolint:gosec
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	cmd.Stderr = &out
+	err := cmd.Run()
+	return out.String(), err
+}
```
schnie, line 22: Just a generic command execution structure we'll use for shelling out `docker`, `podman`, etc commands.

airflow/docker.go
```diff
@@ -200,18 +200,6 @@ func DockerComposeInit(airflowHome, envFile, dockerfile, imageName string) (*Doc
...
-	if err != nil {
-		return err
-	}
-	if runtime.GOOS == "darwin" && containerRuntime == dockerCmd {
-		err := startDocker()
-		if err != nil {
-			return err
-		}
-	}
-
```
schnie, line 214: This has been moved to a cobra `PersistentPreRunE` function named `ConfiguredContainerRuntime` and a `PreRunE` function named `EnsureRuntime`. These will be rolled out more broadly across the `astro dev` commands in the next PR.

airflow/docker_image.go
```diff
@@ -14,6 +14,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/astronomer/astro-cli/airflow/runtimes"
+
```
schnie, line 18: Just bumped all this new container runtime management stuff into a sub package `runtimes`, so that's just getting updated below.

airflow/container_runtime_test.go
schnie, line 1: This is just being moved and expanded within the sub-package `runtimes`.

cmd/airflow_test.go
```diff
@@ -128,13 +128,13 @@ func (s *AirflowSuite) TestDevInitCommandSoftware() {
...
-	cmd := newAirflowInitCmd()
+func (s *AirflowSuite) TestNewAirflowDevRootCmd() {
+	cmd := newDevRootCmd(nil, nil)
 	s.Nil(cmd.PersistentPreRunE(new(cobra.Command), []string{}))
 }
 
-func (s *AirflowSuite) TestNewAirflowStartCmd() {
-	cmd := newAirflowStartCmd(nil)
+func (s *AirflowSuite) TestNewAirflowInitCmd() {
+	cmd := newAirflowInitCmd()
```
schnie, line 137: Weird looking change here, but I've:
- removed the test for `TestNewAirflowStartRootCmd`, the `PersistentPreRunE` function has been moved to the parent command so this test fails and really isn't necessary any more.
- added the test for the `TestNewAirflowDevRootCmd` since it now holds the shared `PersistentPreRunE` function.
- left the test for `TestNewAirflowInitCmd` in place.

airflow/runtimes/container_runtime.go
```diff
@@ -1,17 +1,59 @@
...
+
+	"github.com/briandowns/spinner"
 
 	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/pkg/fileutil"
 	"github.com/astronomer/astro-cli/pkg/util"
 	"github.com/pkg/errors"
 )
 
+var spinnerCharSet = spinner.CharSets[14]
```
jeremybeard, line 18: Do we maybe want to match (or reuse? or change?) the charset in `pkg/ansi/spinner.go` for consistency?

airflow/runtimes/container_runtime.go
```diff
@@ -1,17 +1,59 @@
...
+
+const (
+	docker                         = "docker"
```
jeremybeard, line 21: Do we need these verbatim constants? I'd think if the value changed we'd probably just rename the constant anyway.

airflow/runtimes/docker.go
```diff
@@ -0,0 +1,83 @@
...
+	tickNum               = 500
+	open                  = "open"
+	timeoutErrMsg         = "timed out waiting for docker"
+	dockerOpenNotice      = "We couldn't start the docker engine automatically. Please start it manually and try again."
+)
+
+// DockerInitializer is a struct that contains the functions needed to initialize Docker.
+// The concrete implementation that we use is DefaultDockerInitializer below.
+// When running the tests, we substitute the default implementation with a mock implementation.
+type DockerInitializer struct {
```
jeremybeard, line 21: This looks like it should be an interface, what do you think?

jeremybeard (commented):
This is excellent, I really like the new structure as a way of treating the runtime options equally. Left a couple of small comments.

airflow/runtimes/docker.go
schnie: Ah yea good call, meant to do that. It's now an interface and I've updated the mock implementation and tests accordingly.

airflow/runtimes/container_runtime.go
schnie: Looks like we created a custom one for login, etc. I think standardizing would be good but I'm trying to keep my changes to a minimum and keep them highly correlated. I've started and stopped this PR a few times because I've gone too far down the refactor path. I do plan to keep going, and intend to standardize as much as possible as we go it'll just be in targeted chunks.

This particular spinner charset actually mimics the ones that docker-compose uses so it looks somewhat intentional at the moment since we're using it for the messages where we're dealing with containers and the runtimes, etc.

Screenshot 2024-11-25 at 1 14 50 PM

airflow/runtimes/container_runtime.go
schnie: What are you thinking? Change them back to `dockerCmd`, etc?

airflow/runtimes/container_runtime.go
jeremybeard: Ok fair enough. I appreciate the refactor considering it can feel like a bit of a never-ending rabbit hole.

airflow/runtimes/container_runtime.go
jeremybeard: Could just use the literal string.

jeremybeard (approved):

pritt20 (approved):
Tested the PR changes locally by setting up local astro-project using docker and changes work as expected. 

Thanks @schnie ! 🚀

PR #1752, opened 2024-12-10
Introduction of sha_as_tag=True config in astro cli which will be used during Astro deploy

rujhan-arora-astronomer:
## Description

Introduction of sha_as_tag=True config in astro cli which will be used during Astro deploy.
1. Option to set sha_as_tag true in config.
2. During `astro deploy`, using the sha digest when we make a call to the updateDeploymentImage endpoint in Houston, instead of the runtime version tag.

## 🎟 Issue(s)
Related #6836

## 🧪 Functional Testing

Testing locally and on staging cluster too.

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow/container.go
+++ b/airflow/container.go
@@ -46,7 +46,7 @@ type RegistryHandler interface {
 // ImageHandler defines methods require to handle all operations on/for container images
 type ImageHandler interface {
 	Build(dockerfile, buildSecretString string, config types.ImageBuildConfig) error
-	Push(remoteImage, username, token string) error
+	Push(remoteImage, username, token string, getRemoteShaTag bool) (string, error)
 	Pull(remoteImage, username, token string) error
 	GetLabel(altImageName, labelName string) (string, error)
 	DoesImageExist(image string) error
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -350,19 +350,19 @@ func (d *DockerImage) CreatePipFreeze(altImageName, pipFreezeFile string) error 
 	return nil
 }
 
-func (d *DockerImage) Push(remoteImage, username, token string) error {
+func (d *DockerImage) Push(remoteImage, username, token string, getRemoteShaTag bool) (string, error) {
 	containerRuntime, err := runtimes.GetContainerRuntimeBinary()
 	if err != nil {
-		return err
+		return "", err
 	}
 	err = cmdExec(containerRuntime, nil, nil, "tag", d.imageName, remoteImage)
 	if err != nil {
-		return fmt.Errorf("command '%s tag %s %s' failed: %w", containerRuntime, d.imageName, remoteImage, err)
+		return "", fmt.Errorf("command '%s tag %s %s' failed: %w", containerRuntime, d.imageName, remoteImage, err)
 	}
 
 	registry, err := d.getRegistryToAuth(remoteImage)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	// Push image to registry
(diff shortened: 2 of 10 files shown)


airflow/docker_image.go
```diff
@@ -398,16 +398,34 @@ func (d *DockerImage) Push(remoteImage, username, token string) error {
...
+		if err != nil {
+			return remoteDigest, fmt.Errorf("failed to get digest for image %s: %w", remoteImage, err)
+		}
+		// Parse and clean the output
+		digestOutput := strings.TrimSpace(out.String())
+		if digestOutput != "" {
+			parts := strings.Split(digestOutput, "@")
+			if len(parts) == 2 {
+				remoteDigest = parts[1] // Extract the digest part (after '@')
+			}
```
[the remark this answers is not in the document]
rujhan-arora-astronomer: @neel-astro Actually this is tightly coupled to push because we are getting the digest that we receive from remote registry response after pushing the image. It's not the local SHA of the newly built image.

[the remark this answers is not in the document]
neel-astro: It is not. Correct me if I am wrong, but here we are executing a separate Docker command to fetch the sha information (`docker inspect`). That operation needs to run on the same underlying image and not necessarily the remote image (the SHA tag would remain the same irrespective of the image tag being used to fetch it). But it's a different operation if you think from `ImageHandler` pov.

I am just trying to limit the `Push` method from getting into an operation it didn't sign up for. It already has a tag/untag that shouldn't be part of it if you go by the method definition.

airflow/container.go
```diff
@@ -56,6 +56,7 @@ type ImageHandler interface {
 	Pytest(pytestFile, airflowHome, envFile, testHomeDirectory string, pytestArgs []string, htmlReport bool, config types.ImageBuildConfig) (string, error)
 	ConflictTest(workingDirectory, testHomeDirectory string, buildConfig types.ImageBuildConfig) (string, error)
 	CreatePipFreeze(altImageName, pipFreezeFile string) error
+	GetBuiltImageSha() (string, error)
```
neel-astro, line 59:
```suggestion
	GetImageSha() (string, error)
```

software/deploy/deploy.go
```diff
@@ -181,11 +181,18 @@ func buildPushDockerImage(houstonClient houston.ClientInterface, c *config.Conte
...
 		return err
 	}
-
+	sha := ""
+	if useShaAsTag {
+		sha, err = imageHandler.GetBuiltImageSha()
+		if (sha == "") || (err != nil) {
+			return fmt.Errorf("failed to get image sha: %w", err)
+		}
+	}
```
neel-astro, line 195:
```suggestion
```
nit: keeping usage and value fetching close together to simplify the logic, hence moved it down where it is being used

software/deploy/deploy.go
```diff
@@ -195,13 +202,14 @@ func buildPushDockerImage(houstonClient houston.ClientInterface, c *config.Conte
...
 	}
-
 	err = imageHandler.Push(remoteImage, "", token)
 	if err != nil {
 		return err
 	}
-
 	if byoRegistryEnabled {
+		if useShaAsTag {
+			remoteImage = fmt.Sprintf("%s@%s", registry, sha)
```
neel-astro, line 211:
```suggestion
			sha, err := imageHandler.GetBuiltImageSha()
			if (sha == "") || (err != nil) {
				return fmt.Errorf("failed to get image sha: %w", err)
			}
			remoteImage = fmt.Sprintf("%s@%s", registry, sha)
```
nit: keeping usage and value fetching close together to simplify the logic

software/deploy/deploy.go
```diff
@@ -195,13 +202,14 @@ func buildPushDockerImage(houstonClient houston.ClientInterface, c *config.Conte
...
 		token = c.Token
 	}
-
 	err = imageHandler.Push(remoteImage, "", token)
 	if err != nil {
 		return err
 	}
-
 	if byoRegistryEnabled {
+		if useShaAsTag {
```
neel-astro, line 210: can we add unit tests for this new logic please

neel-astro (commented):
Left final set of comments, but LGTM otherwise

neel-astro (approved):

PR #1756, opened 2024-12-12
Upgrade golangci-lint to v1.62.2

jeremybeard:
## Description

This change upgrades the golangci-lint dependency to the latest version so that it can support Go 1.23, which the CLI is also upgraded to. The corresponding linting errors are then also fixed.

## 📸 Screenshots

Screenshot 2024-12-12 at 12 37 01 PM

## 📋 Checklist


airflow/docker_image.go
```diff
@@ -210,8 +210,8 @@ func (d *DockerImage) Pytest(pytestFile, airflowHome, envFile, testHomeDirectory
 
 	// start pytest container
 	err = cmdExec(containerRuntime, stdout, stderr, []string{"start", "astro-pytest", "-a"}...)
-	if docErr != nil {
-		log.Debugf("Error starting pytest container: %s", docErr.Error())
+	if err != nil {
```
neel-astro, line 213: good catch 👍

airflow/runtimes/docker_runtime.go
```diff
@@ -68,15 +68,15 @@ func (rt DockerRuntime) initializeDocker(timeoutSeconds int) error {
 	// If we got an error, Docker is not running, so we attempt to start it.
 	_, err = rt.Engine.Start()
 	if err != nil {
-		return fmt.Errorf(dockerOpenNotice) //nolint:stylecheck
+		return errors.New(dockerOpenNotice)
```
neel-astro, line 71: nit: we should stay with `fmt.Errorf` I think, since `pkg/errors` is deprecated now, wdyt?

neel-astro (approved):
LGTM, left a minor nit around `pkg/errors` vs `fmt.Errorf`, but approving ahead of time,

airflow/runtimes/docker_runtime.go
jeremybeard: I can change this instance from `github.com/pkg/errors` to `errors`?

We might want a separate cleanup PR at some point to remove `github.com/pkg/errors` completely from the CLI, I see many instances of it, and in ways that can't be transparently switched to `errors`.

PR #1757, opened 2024-12-12
Use new Logrus logger

jeremybeard:
## Description

This changes the CLI's use of the Logrus logging library to use a new logger instance instead of reusing the existing global logger instance of the library. This prevents debug logs being suppressed when third-party libraries that the CLI imports change the logging level of the global logger, as is at least the case for the use of the Docker Compose package mid-way through `astro dev start`.

Note that there are other cases where logs are being suppressed because the root "PersistentPreRunE" function that sets up the logging is overridden by some sub-commands. That issue will be fixed in another PR.

## 🧪 Functional Testing

- Debug logs are continuing to print

## 📋 Checklist


neel-astro (commented):
Thanks for fixing it. 
Nit: the `logger.Logger` wording feels too verbose to use everywhere in the project (not sure why Linter is happy with it 😬😅).

I think either we `dot` import (not a big fan either 😅 , but don't mind)
Or we could have the log level functions (as public functions) in the cli logger package, with the same function signature as the logrus lib, so that we could directly call `logger.Infof` (Same as what logrus global log does).
WDYT?

jeremybeard:
@neel-astro Yes, agreed. I like the logger functions idea. I've pushed a commit with that and it looks a lot cleaner.

neel-astro (approved):
Thanks for the cleanup of the logic, this looks much better ❤️

PR #1763, opened 2024-12-13
Introduction of astro deploy --image-name=image_name --remote --runtime-version=rt_version

rujhan-arora-astronomer:
## Description

Added the support for --image-name flag in the astro deploy command. Internally, it will call the updateDeploymentImage API. No image will be pushed to the registry. The command will be like this -
astro deploy --image-name IMAGE_URL --runtime-version IMAGE_RUNTIME_VERSION

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/6859

## 🧪 Functional Testing

locally
## 📸 Screenshots

Screenshot 2024-12-13 at 5 00 48 PM
Screenshot 2024-12-13 at 4 51 08 PM

## 📋 Checklist


software/deploy/deploy.go
```diff
@@ -164,6 +165,33 @@ func validateRuntimeVersion(houstonClient houston.ClientInterface, tag string, d
 	return nil
 }
 
+func updateDeploymentImageAPICall(houstonClient houston.ClientInterface, imageName, releaseName, airflowVersion, runtimeVersion string) (interface{}, error) {
```
neel-astro, line 168: what's the value addition by having this extra function, that internally just calls the Houston API?

software/deploy/deploy.go
```diff
@@ -164,6 +165,33 @@ func validateRuntimeVersion(houstonClient houston.ClientInterface, tag string, d
...
+
+func UpdateDeploymentImage(houstonClient houston.ClientInterface, deploymentID, wsID, runtimeVersion, imageName string) (string, error) {
+	if runtimeVersion == "" {
+		return "", ErrRuntimeVersionNotPassedForRemoteImage
+	}
+	deploymentIDForCurrentCmd, _, err := getDeploymentIDForCurrentCommandVar(houstonClient, wsID, deploymentID, deploymentID == "")
+	if err != nil {
+		return "", err
+	}
+	deploymentID = deploymentIDForCurrentCmd
```
neel-astro, line 181:
```suggestion
	deploymentID, _, err := getDeploymentIDForCurrentCommandVar(houstonClient, wsID, deploymentID, deploymentID == "")
	if err != nil {
		return "", err
	}
```

software/deploy/deploy_test.go
```diff
@@ -1017,3 +1017,81 @@ func (s *Suite) TestDeployDagsOnlyFailure() {
...
+func (s *Suite) TestUpdateDeploymentImage() {
+	testUtil.InitTestConfig(testUtil.SoftwarePlatform)
+	deploymentID := "test-deployment-id"
+	wsID := "test-workspace-id"
+	runtimeVersion := "12.1.1"
+	imageName := "imageName"
+	releaseName := "releaseName"
+
+	s.Run("When runtimeVersion is empty", func() {
+		houstonMock := new(houston_mocks.ClientInterface)
```
neel-astro, line 1030: Houston mock init should be part of suite's `SetupSubTest` if it is being initialized in every test

software/deploy/deploy_test.go
```diff
@@ -1017,3 +1017,81 @@ func (s *Suite) TestDeployDagsOnlyFailure() {
...
+		returnedDeploymentID, err := UpdateDeploymentImage(houstonMock, deploymentID, wsID, "", imageName)
+		s.ErrorIs(err, ErrRuntimeVersionNotPassedForRemoteImage)
+		houstonMock.AssertExpectations(s.T())
```
neel-astro, line 1033: Houston mock assertion should be part of the suite's `TearDownSubTest` if it is being called at the end of every test

software/deploy/deploy.go
rujhan-arora-astronomer: It's being called from 2 places. Separated it out so that the API call logic remains here for extensibility and DRY.

neel-astro: there is hardly anything to DRY in here 🤷‍♂️ (it is calling a single function in here), and I am not sure how we would extend it in the future.
I don't think we are gaining anything out of this extra function at the moment.

software/deploy/deploy_test.go
```diff
@@ -113,15 +113,29 @@ func (s *Suite) SetupSuite() {
...
+	testUtil.InitTestConfig(testUtil.SoftwarePlatform)
 	s.mockImageHandler = new(mocks.ImageHandler)
 	imageHandlerInit = func(image string) airflow.ImageHandler {
 		return s.mockImageHandler
 	}
 	s.houstonMock = new(houston_mocks.ClientInterface)
 }
 
+func (s *Suite) TearDownSubTest() {
+	s.houstonMock.AssertExpectations(s.T())
```
neel-astro, line 134: nit: Thanks for incorporating, we could now drop the `s.houstonMock.AssertExpectations(s.T())` and `s.mockImageHandler.AssertExpectations(s.T())` from other tests in this suite as well. Can we clean up the code 🙂

neel-astro (commented):
Left some final comments to squash

software/deploy/deploy.go
rujhan-arora-astronomer: Advantages:
1. Code Reusability
It centralizes the request creation logic
Reduces duplication
Makes the code more DRY

2. Improved Readability
Separates concerns
Makes the calling code cleaner

3. Flexibility
Easier to modify request creation in one place
Supports potential future variations (addition of new request parameters and headers in future)

software/deploy/deploy_test.go
rujhan-arora-astronomer: I can create a seperate ticket and PR for that but I don't want to change the other unrelated functions in this PR itself.

neel-astro: I won't say it's unrelated. Also, it's enclosed to unit tests and there is no functional change. But if you want to do it in a separate issue and PR, I would expect that issue linked here and PR to come up sooner rather than going into a backlog.

software/deploy/deploy.go
neel-astro: Since we are on this topic, think in terms of redundancy and maintainability
- It introduces an unnecessary layer of abstraction. The function merely wraps another function call `(houston.Call(houstonClient.UpdateDeploymentImage)(req))` without adding any significant value or functionality. So if you think about it, that also makes it not so flexible, because if we need to modify the request, then we would need to update this function signature and then both the caller function as well, since everything is just passed on by this function.
- It adds complexity to the codebase. Future maintainers must navigate through an additional function call to understand the logic. So in fact it reduces the reliability of the file. Separation of concern does not come into play here since this is just a wrapper function.

software/deploy/deploy_test.go
rujhan-arora-astronomer: It's unrelated because the goal of this PR is to introduce --remote flag and not to update existing tests (unless those tests are somehow related to the functionality change in this PR).
I've created this - https://github.com/astronomer/issues/issues/6862

neel-astro (approved):

PR #1779, opened 2025-01-06
Fix runtime spinner issues

pritt20:
## Description

We observed couple of issues related to progress spinner in Astro CLI during setting up local project: 

1. During local Astro project startup we do not see any meaningful information or progress to keep users updated.
2. Runtime init messages is too long, causing the spinner to print on multiple lines repeatedly. This is usually caused due to small terminal window size. 

Screenshot:

image

The changes in this PR address above issues and provides user meaningful information as we set-up local astro-project.

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

Tested the changes locally, please find below screen recording for ref:

https://github.com/user-attachments/assets/8c363eff-b5e2-4297-8160-8d649692c5cc

## 📋 Checklist


schnie:
@pritt20 I still see the issue if I shrink my terminal down smaller than the first line. I wonder how hard it would be to somehow hook into the terminal size and wrap the text to multiple lines accordingly. Or maybe cut off the line and end with `...` or something.

pritt20:
> @pritt20 I still see the issue if I shrink my terminal down smaller than the first line. I wonder how hard it would be to somehow hook into the terminal size and wrap the text to multiple lines accordingly. Or maybe cut off the line and end with `...` or something.

Thanks @schnie ! Yeah, I am able to reproduce it as well. I'll try other options to see if we can include terminal size dynamically.

airflow/runtimes/command.go
```diff
@@ -20,3 +23,54 @@ func (p *Command) Execute() (string, error) {
...
+	for scanner.Scan() {
+		line := scanner.Text()
+		handler(line)
+	}
+	// Notify completion or error
+	if err := scanner.Err(); err != nil {
+		doneCh <- fmt.Errorf("error reading output: %w", err)
+	} else {
+		doneCh <- nil
+	}
```
neel-astro, line 75:
```suggestion
	if err := scanner.Err(); err != nil {
		doneCh <- fmt.Errorf("error reading output: %w", err)
		return
	}
	doneCh <- nil
```
nit: just simplifying the block

airflow/runtimes/podman_runtime.go
```diff
@@ -189,9 +182,11 @@ func (rt PodmanRuntime) ensureMachine() error {
 
 	// Otherwise, initialize the machine
 	s.Start()
-	if err := rt.Engine.InitializeMachine(podmanMachineName); err != nil {
+	time.Sleep(1 * time.Second)
```
neel-astro, line 185: nit: can we leave a comment in the code on why we are adding the sleep, so that future folks will have the required context

neel-astro (approved):
overall LGTM, left some minor nits, but approving ahead of time

PR #1336, opened 2023-08-03
remove conflict test

sunkickr:
## Description

Remove conflict test. Test currently finds a conflict with the latest version of runtime. This would be too confusing to users as there tests would fail and they would have no way to fix it

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

> List the functional testing steps to confirm this feature or fix.

## 📸 Screenshots

> Add screenshots to illustrate the validity of these changes.

## 📋 Checklist

- [ ] Rebased from the main (or release if patching) branch (before testing)
- [ ] Ran `make test` before taking out of draft
- [ ] Ran `make lint` before taking out of draft
- [ ] Added/updated applicable tests
- [ ] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary).
- [ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary).
- [ ] Communicated to/tagged owners of respective clients potentially impacted by these changes.
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/)

PR #1797, opened 2025-01-30
Removes dead code for conflict-test

schnie:
## Description

This PR removes some seemingly dead code. This `conflictTest` variable will always evaluate to `false`, so we're removing all the code that would run if we branched on this value being `true`. Seems like it used to be a flag that we've removed along the way. Not sure how else this code would ever run.

Update: I found this digging through the history. Looks like we intentionally removed the functionality, but didn't remove the dead code. @kushalmalani even suggested it [here](https://github.com/astronomer/astro-cli/pull/1336#discussion_r1283441336).

## 🎟 Issue(s)

No issue, just noticed while looking at other things.

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


airflow/docker_image.go
```diff
@@ -30,7 +30,6 @@ import (
 )
 
 const (
-	EchoCmd            = "echo"
```
schnie, line 33: Unrelated, but also unused.

cmd/airflow.go
```diff
@@ -60,7 +60,6 @@ var (
 	envExport              bool
 	noBrowser              bool
 	compose                bool
-	conflictTest           bool
```
schnie, line 63: I don't see anywhere in our code that sets this variable. It used to exist at one time as a flag, but doesn't anymore. This would always evaluate to `false`, so we remove the code that branches if this value is `true`.

airflow/docker.go
```diff
@@ -598,13 +598,6 @@ func (d *DockerCompose) UpgradeTest(newAirflowVersion, deploymentID, newImageNam
 	}
 	newDockerFile := destFolder + "/Dockerfile"
 
-	// check for dependency conflicts
-	if conflictTest {
-		err = d.conflictTest(testHomeDirectory, newImageName, newAirflowVersion)
-		if err != nil {
-			return err
-		}
-	}
```
schnie, line 607: Nowhere in our code base do we seem to set `conflictTest` to `true`. There should be no way for this branch to ever run. This `conflictTest` used to exist as a flag on the CLI commands, but seems to have been removed.

kushalmalani (approved):
Thank you @schnie ❤️

PR #1821, opened 2025-03-28
Add Airflow 3 ruff checks to upgrade-test command

jeremybeard:
## Description

This change adds support to the `astro dev upgrade-test` command so that when upgrading to Airflow 3 (including within Airflow 3 releases) the command will run ruff checks to validate the DAGs in the `dags` directory of the project.

The ruff checks are run [as a container](https://docs.astral.sh/ruff/installation/) using the `ghcr.io/astral-sh/ruff:latest` image. The Airflow 3 rules are available [as part of ruff](https://docs.astral.sh/ruff/rules/#airflow-air) itself. As new rules are added to ruff they will be automatically available to the CLI via the latest tag reference.

## 🧪 Functional Testing

- Unit tests added/updated

## 📸 Screenshots

ruff check failures:
Screenshot 2025-03-28 at 2 28 47 PM

ruff check pass:
Screenshot 2025-03-28 at 2 29 27 PM

Not run for Airflow 2:
Screenshot 2025-03-28 at 2 30 06 PM

## 📋 Checklist

### DIFF
--- a/airflow/docker.go
+++ b/airflow/docker.go
@@ -171,7 +171,7 @@ func DockerComposeInit(airflowHome, envFile, dockerfile, imageName string) (*Doc
 	}
 
 	imageHandler := DockerImageInit(ImageName(imageName, "latest"))
-	ruffImageHandler := DockerImageInit("ghcr.io/astral-sh/ruff:latest")
+	ruffImageHandler := DockerImageInit(config.CFG.RuffImage.GetString())
 
 	// Route output streams according to verbosity.
 	var stdout, stderr io.Writer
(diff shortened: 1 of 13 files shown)


airflow/docker.go
```diff
@@ -605,18 +613,35 @@ func (d *DockerCompose) UpgradeTest(newAirflowVersion, deploymentID, newImageNam
...
+		failed = failed || !dagTestPassed
+	}
+
+	if ruffTest && airflowversions.AirflowMajorVersionForRuntimeVersion(newVersion) != "3" {
+		fmt.Println("Skipping ruff Airflow 3 test because not testing upgrade to Airflow 3.x")
+		ruffTest = false
+	}
+
+	if ruffTest {
+		ruffTestPassed, err := d.ruffTest(testHomeDirectory)
```
neel-astro, line 638: Nit: Should we add a section header, like 
`============= Running ruff tests on the dags folder =============` 
or something along those lines?

As of now, it is slightly confusing with the previous line stating that "No errors detected in your DAGs" and then if there are errors in the ruff test we start with the stacktrace.

neel-astro (approved):
Left some minor questions, but looking good otherwise

airflow/docker.go
jeremybeard: Another good call. I added some header lines to make it clearer what is going on.

PR #1822, opened 2025-03-31
Block unsupported Airflow 3 commands

jeremybeard:
## Description

This change blocks `astro deployment` commands that are not yet supported for Airflow 3 deployments.

These are the commands that match:
- `astro deployment [create|update|logs]`
- `astro deployment [variable|airflow-variable|connection|pool|worker-queue] *` 

## 🧪 Functional Testing

- Unit tested added/updated
- Manually checked each impacted command against Airflow 2 and Airflow 3 deployments

## 📸 Screenshots

Screenshot 2025-03-31 at 3 11 38 PM

## 📋 Checklist

### DIFF
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -240,7 +240,7 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
 
 	// Check if the provided runtime version is for Airflow 3
 	if err := airflowversions.ValidateNoAirflow3Support(runtimeVersion); err != nil {
-		return err
+		return fmt.Errorf("%w. Use --runtime-version to specify an Airflow 2 runtime version", err)
 	}
 
 	// validate workspace
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -165,6 +165,11 @@ func Logs(deploymentID, ws, deploymentName, keyword string, logWebserver, logSch
 		return err
 	}
 
+	// Check if deployment is using Airflow 3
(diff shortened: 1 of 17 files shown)


cloud/deployment/deployment.go
```diff
@@ -233,6 +238,11 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
 		return err
 	}
 
+	// Check if the provided runtime version is for Airflow 3
+	if err := airflowversions.ValidateNoAirflow3Support(runtimeVersion); err != nil {
```
neel-astro, line 242: nit: I haven't tried it, but what if no runtime version is specified in the command? will that cause the command to default to AF 3 version and then get blocked here on day 0?

neel-astro (commented):
Left a minor question, otherwise LGTM

cloud/deployment/deployment.go
jeremybeard: Yes, we are defaulting to Airflow 3 across the CLI when that is released. I only set the flag in the screenshot because it isn't released yet.

neel-astro: in this case, if a user does not set the flag in their command, the defaulting logic would pick AF 3, and then this piece of code would error out, which isn't ideal since CLI is picking the default for the users. I think we might have to use AF 2 default for deployment create logic, wdyt?

jeremybeard: Discussed offline, agreed that this command should exclude Airflow 3 from the default runtime version, so I've pushed a commit for that.

neel-astro (approved):
LGTM ❤️

PR #1829, opened 2025-04-07
Add lint flags to upgrade-test command

jeremybeard:
## Description

This changes the recently added ruff tests of the `astro dev upgrade-test` command:
- The option is now named around linting rather than ruff specifically
- The default linting config only checks for breaking changes in Airflow 3 (`AIR30` rules)
- A new optional flag `--lint-deprecations` also checks for deprecations in Airflow 3 (`AIR31` rules, which do not yet exist but will very soon)
- A new optional flag `--lint-config-file` which overrides the ruff config entirely with a local file

## 🧪 Functional Testing

- Unit tests added/updated
- Manually tested for default configs, with deprecations, and with a provided config file

## 📋 Checklist

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -66,7 +66,9 @@ var (
 	compose                bool
 	versionTest            bool
 	dagTest                bool
-	ruffTest               bool
+	lintTest               bool
+	lintDeprecations       bool
+	lintConfigFile         string
 	waitTime               time.Duration
 	containerRuntime       runtimes.ContainerRuntime
 	RunExample             = `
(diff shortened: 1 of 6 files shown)


cmd/airflow.go
```diff
@@ -203,10 +205,12 @@ func newAirflowUpgradeTestCmd(platformCoreClient astroplatformcore.CoreClient) *
 			return airflowUpgradeTest(cmd, platformCoreClient)
 		},
 	}
-	cmd.Flags().StringVarP(&airflowVersion, "airflow-version", "a", "", "The version of Airflow you want to upgrade to. The default is the latest available version. Tests are run against the equivalent Astro Runtime version. ")
+	cmd.Flags().StringVarP(&airflowVersion, "airflow-version", "a", "", "The version of Airflow you want to upgrade to. The default is the latest available version. Tests are run against the equivalent Astro Runtime version.")
 	cmd.Flags().BoolVarP(&versionTest, "version-test", "", false, "Only run version tests. These tests show you how the versions of your dependencies will change after you upgrade.")
 	cmd.Flags().BoolVarP(&dagTest, "dag-test", "d", false, "Only run DAG tests. These tests check whether your DAGs will generate import errors after you upgrade.")
-	cmd.Flags().BoolVarP(&ruffTest, "ruff-test", "r", false, "Only run ruff tests. These tests check whether your DAGs are compatible with Airflow 3.")
+	cmd.Flags().BoolVarP(&lintTest, "lint-test", "l", false, "Only run ruff lint tests. These tests check whether your DAGs are compatible with Airflow.")
```
neel-astro, line 211:
```suggestion
	cmd.Flags().BoolVarP(&lintTest, "lint-test", "l", false, "Run ruff lint tests. These tests check whether your DAGs are compatible with Airflow.")
```

Sorry I just noticed it now, but I think the flag is set to true would ensure that the lint tests are run, and won't ensure that other tests don't run.
I think we should also update the description for `dag-test` and `version-text` flag, wdyt?

neel-astro (approved):

cmd/airflow.go
jeremybeard: The [way it works](https://github.com/astronomer/astro-cli/blob/3de24616bc190ae9ee3df1d6090588d2b5199324/airflow/docker.go#L563-L568) (not defending it...) is that if all of the `*-test` flags are false then they all run, otherwise it is only the flags that are set to true. So "only run" is not quite right if you set more than one of them to true, but I wouldn't want to imply one only runs if you set its flag. Bit of a UX headache but being existing behavior I'd rather just leave it as is.

PR #1831, opened 2025-04-07
Block tarball members pointing outside tarball

### LINKED ISSUE #1825: Block Airflow 3 DAG deploys where tarball members point outside of tarball
Airflow 3 deployments will not support tarballs that have members which point outside of the tarball, using either relative paths or symlinks.

jeremybeard:
## Description

This changes DAG deploys so that the bundle tarball members cannot symlink to files outside of the tarball. Such external files are not included in DAG deploys and so are not available on the Airflow containers.

This check is only added for deploys to Airflow 3 deployments because in Airflow 3 DAG deploys do not support external symlinks and we want to block those at the source. It is possible that bad symlinks are being deployed to existing Airflow 2 deployments and we do not want to unnecessarily break those deploys.

## 🎟 Issue(s)

Resolves #1825

## 🧪 Functional Testing

- Added unit tests
- Manually checked with symlink in dags dir

## 📸 Screenshots

Screenshot 2025-04-07 at 3 30 52 PM

## 📋 Checklist

### DIFF
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -155,7 +155,7 @@ func shouldIncludeMonitoringDag(deploymentType astroplatformcore.DeploymentType)
 	return !organization.IsOrgHosted() && !deployment.IsDeploymentDedicated(deploymentType) && !deployment.IsDeploymentStandard(deploymentType)
 }
 
-func deployDags(path, dagsPath, dagsUploadURL string, validateSymlinks bool, deploymentType astroplatformcore.DeploymentType) (string, error) {
+func deployDags(path, dagsPath, dagsUploadURL, currentRuntimeVersion string, deploymentType astroplatformcore.DeploymentType) (string, error) {
 	if shouldIncludeMonitoringDag(deploymentType) {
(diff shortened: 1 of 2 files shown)


cloud/deploy/deploy.go
```diff
@@ -276,8 +276,11 @@ func Deploy(deployInput InputDeploy, platformCoreClient astroplatformcore.CoreCl
 			return fmt.Errorf(enableDagDeployMsg, deployInfo.deploymentID) //nolint
 		}
 
+		// only validate symlinks for Airflow 3.x (i.e. don't break backwards compatibility with Airflow 2.x deployments)
+		validateSymlinks := airflowversions.AirflowMajorVersionForRuntimeVersion(deployInfo.currentVersion) == "3"
```
neel-astro, line 280: nit: maybe we could move this outside the if/else block to save ourself from repeating

neel-astro (approved):

cloud/deploy/deploy.go
jeremybeard: Yeah good call. I've centralized it in UploadBundle

PR #1843, opened 2025-04-21
Pull latest Ruff image on upgrade-test

jeremybeard:
## Description

This change adds an image pull of the latest Ruff image before running it to lint the project. Without this the Ruff image will not update itself as new rules are added.

## 🧪 Functional Testing

- Ran without the image locally, it pulls as already the case
- Ran with the image locally, it pulls for the latest

## 📸 Screenshots

Screenshot 2025-04-21 at 4 23 32 PM

Screenshot 2025-04-21 at 4 23 45 PM

## 📋 Checklist

### DIFF
--- a/airflow/docker_image.go
+++ b/airflow/docker_image.go
@@ -432,6 +432,10 @@ func (d *DockerImage) getRegistryToAuth(imageName string) (string, error) {
 }
 
 func (d *DockerImage) Pull(remoteImage, username, token string) error {
+	if remoteImage == "" {
+		remoteImage = d.imageName
+	}
+
 	// Pulling image to registry
 	fmt.Println(pullingImagePrompt)
 	containerRuntime, err := runtimes.GetContainerRuntimeBinary()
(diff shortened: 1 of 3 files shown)


airflow/docker_image.go
```diff
@@ -432,6 +432,10 @@ func (d *DockerImage) getRegistryToAuth(imageName string) (string, error) {
 }
 
 func (d *DockerImage) Pull(remoteImage, username, token string) error {
```
neel-astro, line 434: this method feels slightly wrong to me right now. Mostly I had written it or reviewed it sometime in the past 🙈 . But thinking about it now, why do we even have two separate ways to pass image name?

This is not a blocker at all, just throwing a comment to stir a conversation on the right approach.

I think the right approach here might be to drop the remoteImage arg and force everyone to use the imageName from DockerImage, since the intention of the interface was to club the operation around an image.

jeremybeard: The image handler abstraction is unfortunately a wide mix of "handle one image" and "handle any image" so we'd need a separate effort to try to resolve that.

PR #1847, opened 2025-05-12
Use floating tag for init of Airflow 3 Dockerfile

jeremybeard:
## Description

This changes the `astro dev init` Dockerfile for Airflow 3 to use the minor version floating tag instead of the exact patch tag (e.g. `3.0` instead of `3.0-1`). This will mean that new patches of a minor version will automatically be picked up by the project.

## 📋 Checklist


airflow/docker_registry_test.go
jeremybeard, line 1: Dropped this file because it seems to be interacting with Docker and so can unreliably fail, and is anyway not really unit testing any of the CLI logic.

cmd/airflow_test.go
```diff
@@ -182,10 +182,6 @@ func (s *AirflowSuite) Test_airflowInitNonEmptyDir() {
 		defer testUtil.MockUserInput(s.T(), "y")()
 		err := airflowInit(cmd, args)
 		s.NoError(err)
-
-		b, _ := os.ReadFile(filepath.Join(s.tempDir, "Dockerfile"))
```
jeremybeard, line 186: More non-unit unit testing, this time these fail because now that Airflow 3 is released the default Dockerfile does not contain this string.

PR #1860, opened 2025-05-26
Fix integration tests to be inline with AF3

neel-astro:
## Description

Changes:
- Updated the expected containers in CLI ITs to be inline with AF3 components
- Updated the integration to reduce flakyness due to sleep and teardown component containers even during failures


pritt20 (approved):
Thanks @neel-astro !!

Issue #1895, opened 2025-07-18
`--remote-execution-enabled` flag creates standard deployment instead of failing or prompting for required config


dsgupta02:
**Describe the bug**

When using the `--remote-execution-enabled ` flag with astro deployment create, the CLI proceeds to create a Standard deployment (either with Celery or Astro executor), even though the intention was to create a Remote Execution deployment.

This can lead to unexpected deployment types being created without any warning or guidance.

```
deeksha@Deekshasahus-MacBook-Pro ~ % /usr/local/bin/astro deployment create -n "test_cli_remote_execution"  --remote-execution-enabled
Current Workspace: qa_test


Please select a Region for your Deployment:
 #     REGION
 1     westeurope
 2     westus2
 3     eastus2
 4     australiaeast

> 2
 NAME                          NAMESPACE                CLUSTER     CLOUD PROVIDER     REGION      DEPLOYMENT ID                 RUNTIME VERSION                    DAG DEPLOY ENABLED     CI-CD ENFORCEMENT     DEPLOYMENT TYPE
 test_cli_remote_execution     astral-function-6099     N/A         AZURE              westus2     cmd8i5vyj032801nz220ag8cu     3.0-5 (based on Airflow 3.0.3)     false                  false                 STANDARD

 Successfully created Deployment: test_cli_remote_execution
 Deployment can be accessed at the following URLs

 Deployment Dashboard: cloud.astronomer-dev.io/cm6kqel7k00pv01o9ftlsxvha/deployments/cmd8i5vyj032801nz220ag8cu
 Airflow Dashboard: clztuyh0a01z701mwxjahsza2.astronomer-dev.run/d20ag8cu?orgId=org_bql5Hn1KELMRNPoM
deeksha@Deekshasahus-MacBook-Pro ~ % /usr/local/bin/astro deployment create -n "test_cli_remote_execution"  --remote-execution-enabled --executor Astro
Current Workspace: qa_test


Please select a Region for your Deployment:
 #     REGION
 1     westeurope
 2     australiaeast
 3     westus2
 4     eastus2

> 3
Error: Invalid request: Deployment name 'test_cli_remote_execution' already exists in this workspace
deeksha@Deekshasahus-MacBook-Pro ~ % /usr/local/bin/astro deployment create -n "test_cli_remote_execution_create"  --remote-execution-enabled --executor Astro
Current Workspace: qa_test


Please select a Region for your Deployment:
 #     REGION
 1     eastus2
[...]

PR #1898, opened 2025-07-18
Improve validation of remote execution support

feluelle:
## Description

Currently, we allow passing remote execution fields on both standard and dedicated deployments. However, it is only truly supported at the moment for dedicated. Now, we always set the remote execution spec for hosted deployments, which means that if the user sets it on a standard deployment request, the api will return an error describing what's currently supported, vs silently not setting it.

## 🎟 Issue(s)

Related #1895 

## 🧪 Functional Testing

```
❯ ./astro deployment create --remote-execution-enabled
Current Workspace: felix

Please specify a name for your Deployment

Deployment name: test

Please select a Region for your Deployment:
 #     REGION            
 1     eastus2           
 2     westus2           
 3     australiaeast     
 4     westeurope        

> 1
Error: Invalid field values: field remoteExecution failed check on excluded_unless; remoteExecution is excluded unless Executor is 'ASTRO';,field workerQueues failed check on workerQueuesRequiredForCelery;
```

```
❯ ./astro deployment create --remote-execution-enabled --executor ASTRO
Current Workspace: felix

Please specify a name for your Deployment

Deployment name: test

Please select a Region for your Deployment:
 #     REGION            
 1     westeurope        
 2     westus2           
 3     australiaeast     
 4     eastus2           

> 2
Error: Invalid request: Remote execution is only supported for deployments on dedicated clusters
```

```
❯ ./astro deployment create --remote-execution-enabled --executor ASTRO --type dedicated --cluster-id cm4sauxuo03ig01rfux2aun5d --resource-quota-cpu 2
Current Workspace: felix

Please specify a name for your Deployment

Deployment name: testss
Error: Invalid field values: field resourceQuotaCpu failed check on excluded_if; resourceQuotaCpu is excluded if RemoteExecution.Enabled is 'true';
```

```
❯ ./astro deployment create --remote-execution-enabled --executor ASTRO --type dedicated --cluster-id cm4sauxuo03ig01rfux2aun5d --resource-quota-memory 2
Current Workspace: felix

Please specify a name for your Deployment

Deployment name: asd
Error: Invalid field values: field resourceQuotaMemory failed check on excluded_if; resourceQuotaMemory is excluded if RemoteExecution.Enabled is 'true';
```

- [x] Create Celery deployment with default worker queues
- [x] Create Astro deployment with default worker queues
- [x] Create Astro Remote deployment from file

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -323,17 +323,18 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
 			Name:              "default",
 			WorkerConcurrency: workerConcurrency,
 		}}
-		if defaultTaskPodCpu == "" {
-			defaultTaskPodCpu = configOption.ResourceQuotas.DefaultPodSize.Cpu.Default
-		}
-		if defaultTaskPodMemory == "" {
-			defaultTaskPodMemory = configOption.ResourceQuotas.DefaultPodSize.Memory.Default
-		}
-		if resourceQuotaCpu == "" {
-			resourceQuotaCpu = configOption.ResourceQuotas.ResourceQuota.Cpu.Default
-		}
-		if resourceQuotaMemory == "" {
-			resourceQuotaMemory = configOption.ResourceQuotas.ResourceQuota.Memory.Default
+		defaultTaskPodCPUPtr := CreateDefaultTaskPodCPU(defaultTaskPodCpu, remoteExecutionEnabled, &configOption)
+		defaultTaskPodMemoryPtr := CreateDefaultTaskPodMemory(defaultTaskPodMemory, remoteExecutionEnabled, &configOption)
+		resourceQuotaCPUPtr := CreateResourceQuotaCPU(resourceQuotaCpu, remoteExecutionEnabled, &configOption)
+		resourceQuotaMemoryPtr := CreateResourceQuotaMemory(resourceQuotaMemory, remoteExecutionEnabled, &configOption)
+		var remoteExecution *astroplatformcore.DeploymentRemoteExecutionRequest
+		if remoteExecutionEnabled {
+			remoteExecution = &astroplatformcore.DeploymentRemoteExecutionRequest{
+				Enabled:                true,
+				AllowedIpAddressRanges: allowedIpAddressRanges,
+				TaskLogBucket:          taskLogBucket,
+				TaskLogUrlPattern:      taskLogURLPattern,
+			}
 		}
 		var deplWorkloadIdentity *string
 		if workloadIdentity != "" {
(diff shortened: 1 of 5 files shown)


cloud/deployment/deployment.go
```diff
@@ -374,16 +375,19 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
...
 			}
-			if strings.EqualFold(executor, CeleryExecutor) || strings.EqualFold(executor, CELERY) || strings.EqualFold(executor, AstroExecutor) || strings.EqualFold(executor, ASTRO) {
-				standardDeploymentRequest.WorkerQueues = &defautWorkerQueue
+			if standardDeploymentRequest.RemoteExecution == nil || !standardDeploymentRequest.RemoteExecution.Enabled {
+				if strings.EqualFold(executor, CeleryExecutor) || strings.EqualFold(executor, CELERY) || strings.EqualFold(executor, AstroExecutor) || strings.EqualFold(executor, ASTRO) {
+					standardDeploymentRequest.WorkerQueues = &defautWorkerQueue
+				}
 			}
-			switch schedulerSize {
+			switch schedulerSize { //nolint:dupl
```
neel-astro, line 390: Should this be a utility method instead of using nolint:dupl?

cloud/deployment/deployment.go
```diff
@@ -393,7 +397,11 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
 			case strings.ToLower(string(astrocore.CreateStandardDeploymentRequestSchedulerSizeEXTRALARGE)):
 				standardDeploymentRequest.SchedulerSize = astroplatformcore.CreateStandardDeploymentRequestSchedulerSizeEXTRALARGE
 			case "":
-				standardDeploymentRequest.SchedulerSize = astroplatformcore.CreateStandardDeploymentRequestSchedulerSize(configOption.DefaultValues.SchedulerSize)
+				if standardDeploymentRequest.RemoteExecution != nil && standardDeploymentRequest.RemoteExecution.Enabled {
```
neel-astro: Standard deployment should have remote execution as nil right? or am I missing something?

cloud/deployment/deployment.go
```diff
@@ -1294,6 +1276,92 @@ func updateWorkerQueuesForExecutor(newExecutor astroplatformcore.DeploymentExecu
 	}
 }
 
+func CreateDefaultTaskPodCPU(defaultTaskPodCPU string, isRemoteExecutionEnabled bool, configOption *astrocore.DeploymentOptions) *string {
```
neel-astro, line 1279: nit: should we add unit tests for these new functions? Or are we sure that the existing test cases cover all possible variations?

neel-astro (commented):
Left some minor comments, but otherwise LGTM

cloud/deployment/deployment.go
feluelle: Yes, theoretically, but the API spec doesn't tell us that. I discussed that with @ianbuss, and we think having less business logic (which it is if it is not part of the OpenAPI spec) is better.

feluelle: It is better to surface the raw API error rather than a possible outdated or inaccurate error message.

cloud/deployment/deployment.go
feluelle: I have manually tested some of them, but UTs make sense to me actually, because they are public and shared between `deployment.go` and `from_file.go`. I will add some. Thank you!

cloud/deployment/deployment.go
feluelle: It is not quite the same. There are a lot of dedicated vs standard references.

cloud/deployment/deployment.go
```diff
@@ -415,40 +423,31 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
...
-				dedicatedDeploymentRequest.DefaultTaskPodMemory = &defaultTaskPodMemory
-				dedicatedDeploymentRequest.ResourceQuotaCpu = &resourceQuotaCpu
-				dedicatedDeploymentRequest.ResourceQuotaMemory = &resourceQuotaMemory
+			if !remoteExecutionEnabled {
 				if strings.EqualFold(executor, CeleryExecutor) || strings.EqualFold(executor, CELERY) || strings.EqualFold(executor, AstroExecutor) || strings.EqualFold(executor, ASTRO) {
 					dedicatedDeploymentRequest.WorkerQueues = &defautWorkerQueue
 				}
 			}
-			switch schedulerSize {
+			switch schedulerSize { //nolint:dupl
```
neel-astro, line 450: Nit: if you look at it, the case should be on the dedicated deployment request scheduler size small, and so on for the rest, i.e., dedicated instead of standard. Tiny thing in the existing codebase, but could lead to buggy code someday.

cloud/deployment/deployment.go
neel-astro: ok, I think there is scope to improve still, generics could be one way to look at it. But not a blocker for this PR, as I already have a story incoming from the other nova to make these insane if/else blocks a bit simpler.

neel-astro (approved):

cloud/deployment/deployment.go
feluelle: Oh yeah, the refs are incorrect. Let's handle that on a different PR.

cloud/deployment/deployment.go
feluelle: Yes, 100%, but I don't want to increase the scope of this further. There's too much to clean up.

PR #1907, opened 2025-08-04
Added the support for clusterID for the pre-created namespaces flows in CLI 

rujhan-arora-astronomer:
## Description

Currently, the CLI fetches the list of available pre-created namespaces and applies filters based on the Houston configmap. This approach is problematic for 1.0 because it doesn't account for feature flags or specific configurations that may be enabled on a per-cluster basis.

As a result, a user might see a list of pre-created namespaces that is not accurate for their target environment. 
This PR resolves this issue by ensuring that whenever the CLI requests a list of pre-created namespaces, it passes the relevant clusterId ID. The backend API can then use this ID to evaluate cluster-specific feature flags and return a precise, filtered list of pre-created namespaces that are actually available for that cluster.

## 🎟 Issue(s)

Related https://github.com/astronomer/issues/issues/7590

## 🧪 Functional Testing

Locally
Unit tests

## 📸 Screenshots

Screenshot 2025-08-04 at 3 36 31 PM
Screenshot 2025-08-04 at 4 46 45 PM
Screenshot 2025-08-04 at 4 46 55 PM

## 📋 Checklist


cmd/software/deployment.go
```diff
@@ -55,22 +55,23 @@ var (
 	knowHosts               string
 	runtimeVersion          string
 	desiredRuntimeVersion   string
+	clusterID               string
 	deploymentCreateExample = `
 # Create new deployment with Celery executor (default: celery without params).
-$ astro deployment create --label=new-deployment-name --executor=celery
+$ astro deployment create --label=new-deployment-name --executor=celery --cluster-id=123(mandatory for Software release 1.0.0 and above)
```
pgvishnuram: do we need to put them on brackets ??

cmd/software/deployment.go
```diff
@@ -167,6 +168,15 @@ func newDeploymentCreateCmd(out io.Writer) *cobra.Command {
...
 	cmd.Flags().StringVarP(&cloudRole, "cloud-role", "c", "", "Set cloud role to annotate service accounts in deployment")
+	houstonVersion, err := houstonClient.GetPlatformVersion(nil)
+	if err != nil {
+		fmt.Println("Unable to get Houston version: ", err.Error())
+		return cmd
+	}
+	if houstonVersion >= "1.0.0" {
+		cmd.Flags().StringVarP(&clusterID, "cluster-id", "", "", "Set cluster ID to create deployment in (mandatory for Software release 1.0.0 and above)")
+		_ = cmd.MarkFlagRequired("cluster-id")
+	}
```
pgvishnuram: with this if we set a clusterId and any create deployment we initiate it will use the set clusterId right

houston/app.go
```diff
@@ -68,12 +68,26 @@ var (
...
+	}`,
+		},
+		{
+			version: "1.0.0",
+			query: `
+	query availableNamespaces($clusterId: Uuid!) {
+		availableNamespaces(clusterId: $clusterId) {
+			name
+		}
+	}`,
```
pgvishnuram, line 88: we have to update our api as well in order to make it work insync with dataplanes

cmd/software/deployment.go
rujhan-arora-astronomer: They don't need to add in brackets. Where do you see it?

cmd/software/deployment.go
rujhan-arora-astronomer: Yes, this clusterId will be used for that create call

houston/app.go
rujhan-arora-astronomer: It's unrelated to this PR and we can discuss it seperately.

cmd/software/deployment.go
karankhanchandani: I think what Vishnu is saying that in the cli, you see `(mandatory for Software release 1.0.0 and above)` in the examples section of the command. Do we need it there? We can add this information (cluster ID is mandatory) in the helper of this command.

cmd/software/deployment.go
karankhanchandani: If cluster ID is mandatory, you can specify it using the function https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired. This will fail the command if the flag is not specified and does not hit the backend.

cmd/software/deployment.go
pgvishnuram: since the cli itself makes the field as required explicity adding in the command examples might not make sense

cmd/software/deployment.go
pgvishnuram: do workspace admin will able to make this call like set clusterId

rujhan-arora-astronomer: Anyone can pass the cluster ID in this command. We don't have RBACs at the CLI level.

rujhan-arora-astronomer:
Screenshot 2025-08-11 at 9 25 31 PM

PR #1936, opened 2025-09-15
Feat/add per dp software registry login

pgvishnuram:
## Description

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/software/auth/auth.go
+++ b/software/auth/auth.go
@@ -98,6 +98,11 @@ func registryAuth(client houston.ClientInterface, out io.Writer) error {
 	}
 
 	var registry string
+	if appConfig.Version >= "1.0.0" && !appConfig.Flags.BYORegistryEnabled {
+		logger.Info("skipping registry login")
+		return nil
+	}
+
 	if appConfig.Flags.BYORegistryEnabled {
 		registry = appConfig.BYORegistryDomain
 	} else {
--- a/software/auth/auth.go
+++ b/software/auth/auth.go
@@ -81,8 +81,8 @@ func oAuth(oAuthURL string) string {
 	return input.Text(inputOAuthToken)
 }
 
-// registryAuth authenticates with the private registry
-func registryAuth(client houston.ClientInterface, out io.Writer) error {
+// RegistryAuth authenticates with the private registry
+func RegistryAuth(client houston.ClientInterface, out io.Writer) error {
 	c, err := context.GetCurrentContext()
 	if err != nil {
 		return err
@@ -239,7 +239,7 @@ func Login(domain string, oAuthOnly bool, username, password, houstonVersion str
 		}
 	}
 
-	err = registryAuth(client, out)
+	err = RegistryAuth(client, out)
 	if err != nil {
 		logger.Debugf("There was an error logging into registry: %s", err.Error())
 	}
--- a/software/deploy/deploy.go
+++ b/software/deploy/deploy.go
@@ -19,6 +19,7 @@ import (
 	"github.com/astronomer/astro-cli/pkg/input"
 	"github.com/astronomer/astro-cli/pkg/logger"
 	"github.com/astronomer/astro-cli/pkg/printutil"
+	"github.com/astronomer/astro-cli/software/auth"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


software/deploy/deploy.go
```diff
@@ -202,15 +203,27 @@ func UpdateDeploymentImage(houstonClient houston.ClientInterface, deploymentID,
...
 	var registry, remoteImage, token string
 	if byoRegistryEnabled {
 		registry = byoRegistryDomain
 		remoteImage = fmt.Sprintf("%s:%s", registry, fmt.Sprintf("%s-%s", name, nextTag))
 	} else {
-		registry = registryDomainPrefix + cloudDomain
-		remoteImage = fmt.Sprintf("%s/%s", registry, airflow.ImageName(name, nextTag))
-		token = c.Token
+		platformVersion, _ := houstonClient.GetPlatformVersion(nil)
+		if platformVersion >= "1.0.0" {
```
ianbuss: This is an unsafe comparison as it won't compare semantic versions correctly for say, "1.2.0" and "1.10.0". Same above.

software/auth/auth.go
```diff
@@ -98,6 +98,11 @@ func registryAuth(client houston.ClientInterface, out io.Writer) error {
 	}
 
 	var registry string
+	if appConfig.Version >= "1.0.0" && !appConfig.Flags.BYORegistryEnabled {
```
ianbuss: Do we not support BYO registry in 1.0?

software/deploy/deploy.go
pgvishnuram: In lots of places we have been using this direct comparision thats why i just reused them. post 1.0 the login will be per dataplane logic and the initial login will be skipped automatically. Let me also try with different version scheme to see it works.

software/auth/auth.go
pgvishnuram: right now we have few cases handled as below
1. we are exposing registry url in per deployment response 
2. post 1.0 we are skipping the initial registry login
3. if its 1.0 and above we skip the initial login and directly login registry
4. for BYO registry we still have a plans to query from deployment api or just use the existing flow.

software/deploy/deploy.go
astro-walker: If this pattern is used elsewhere, file a bug report or fix it. Ian is correct that lexicographic comparisons like Python's `==` for `str` types do not correctly handle semantic versions, dates, etc with variable-width fields. It may have worked if the none of the version fields are 10 or higher, otherwise this will result in improper feature gating.

software/deploy/deploy.go
```diff
@@ -323,6 +336,19 @@ func getAirflowUILink(deploymentID string, deploymentURLs []houston.DeploymentUR
 	return ""
 }
 
+func getDeploymentRegistryUrl(deploymentID string, deploymentURLs []houston.DeploymentURL) string {
+	if deploymentID == "" {
```
astro-walker: Is this not an error case? I don't see where you handle empty host names in the calling code.

software/deploy/deploy.go
pgvishnuram: @ianbuss and @astro-walker we have move to semver checker now - this should be good now.

software/auth/auth.go
pgvishnuram: same here
@ianbuss and @astro-walker we have move to semver checker now - this should be good now.

software/deploy/deploy.go
```diff
@@ -202,15 +204,30 @@ func UpdateDeploymentImage(houstonClient houston.ClientInterface, deploymentID,
...
+		platformVersion, _ := houstonClient.GetPlatformVersion(nil)
+		if versions.GreaterThanOrEqualTo(platformVersion, "1.0.0") {
+			registry, err := getDeploymentRegistryUrl(deploymentInfo.Urls)
+			if err != nil {
+				return err
+			}
+			// Switch to per deployment registry login
+			err = auth.RegistryAuth(houstonClient, os.Stdout)
+			if err != nil {
+				logger.Debugf("There was an error logging into registry: %s", err.Error())
```
ianbuss, line 222: Why do we not exit here? It's not going to work if we haven't auth'd right?

ianbuss (commented):
Small comment. We should also try to add at least some description to the PR

software/deploy/deploy.go
pgvishnuram: os.exit(1) right

jeremybeard: Better to return the error up to the command so it fails there.

pgvishnuram: in existing code as well even if the login fails on push it will get into error - thats why we didnt change this, now should we make this to return error with fail even when the auth gets failed

pgvishnuram:
coveralls is failing - tests have already passed

software/deploy/deploy.go
ianbuss: Returning the `err` here is the right call. Better to fail earlier if we know it isn't going to work.

software/deploy/deploy_test.go
```diff
@@ -344,6 +352,26 @@ func (s *Suite) TestGetAirflowUILink() {
 	s.Equal(expectedResult, actualResult)
 }
 
+func (s *Suite) TestGetDeploymentRegistryURL() {
+	mockURLs := []houston.DeploymentURL{
+		{URL: "https://deployments.local.astronomer.io/testDeploymentName/airflow", Type: "airflow"},
+		{URL: "https://deployments.local.astronomer.io/testDeploymentName/flower", Type: "flower"},
+		{URL: "registry.local.astronomer.io", Type: "registry"},
```
ianbuss, line 359: Docker login will work for both `registry.local.astronomer.io` and `https://registry.local.astronomer.io` so might be worth adding both as a test case here in case we ever wanted to change the houston response to be a "real" URL.

ianbuss (approved):
Left a nit but otherwise looks good to me.

software/deploy/deploy_test.go
pgvishnuram: https wont work - we are sending only the url here without https

Issue #1805, opened 2025-02-24
Support providing workspace id on organization switch


jeremybeard:
Running `astro organization switch [organization-id]` where the organization has multiple accessible workspaces, and the context does not have a last used workspace id, will print out a table of accessible workspaces and ask the user to select the workspace they wish to switch to.

For programmatic use cases this is challenging. To handle this we should add an optional `--workspace-id` flag to `astro organization switch` that automatically handles the subsequent workspace switch.

PR #1938, opened 2025-09-16
feat(organization): add --workspace-id flag to org switch for non-interactive use

ronidelacruz-astro:
## Description
Add --workspace-id (-w) to astro organization switch to allow non-interactive workspace switching after changing org.

## 🎟 Issue(s)

Related #1805

## 🧪 Functional Testing

- astro organization switch <org-id> → still prompts if multiple workspaces.
- astro organization switch <org-id> --workspace-id <ws-id> → switches org and workspace without prompt.
- Verified via unit tests (TestOrganizationSwitch).

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cmd/cloud/organization.go
+++ b/cmd/cloud/organization.go
@@ -206,8 +206,7 @@ func organizationSwitch(cmd *cobra.Command, out io.Writer, args []string) error 
 		organizationNameOrID = args[0]
 	}
 
-	err := orgSwitch(organizationNameOrID, astroCoreClient, platformCoreClient, out, shouldDisplayLoginLink)
-	if err != nil {
+	if err := orgSwitch(organizationNameOrID, astroCoreClient, platformCoreClient, out, shouldDisplayLoginLink); err != nil {
 		return err
 	}
 
--- a/cmd/cloud/organization.go
+++ b/cmd/cloud/organization.go
@@ -12,6 +12,7 @@ import (
 	roleClient "github.com/astronomer/astro-cli/cloud/role"
 	"github.com/astronomer/astro-cli/cloud/team"
 	"github.com/astronomer/astro-cli/cloud/user"
+	"github.com/astronomer/astro-cli/cloud/workspace"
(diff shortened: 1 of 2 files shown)


cmd/cloud/organization.go
```diff
@@ -202,7 +206,15 @@ func organizationSwitch(cmd *cobra.Command, out io.Writer, args []string) error
 		organizationNameOrID = args[0]
 	}
 
-	return orgSwitch(organizationNameOrID, astroCoreClient, platformCoreClient, out, shouldDisplayLoginLink)
+	err := orgSwitch(organizationNameOrID, astroCoreClient, platformCoreClient, out, shouldDisplayLoginLink)
+	if err != nil {
```
ianbuss:
```suggestion
	if err := orgSwitch(organizationNameOrID, astroCoreClient, platformCoreClient, out, shouldDisplayLoginLink); err != nil {
```
nit

cmd/cloud/organization_test.go
```diff
@@ -83,13 +83,55 @@ func TestOrganizationList(t *testing.T) {
...
-		return nil
-	}
+	t.Run("workspace flag triggers wsSwitch with provided id", func(t *testing.T) {
+		testUtil.InitTestConfig(testUtil.LocalPlatform)
 
-	cmdArgs := []string{"switch"}
-	_, err := execOrganizationCmd(cmdArgs...)
-	assert.NoError(t, err)
+		origOrgSwitch := orgSwitch
+		orgSwitch = func(orgName string, coreClient astrocore.CoreClient, platformCoreClient astroplatformcore.CoreClient, out io.Writer, shouldDisplayLoginLink bool) error {
```
ianbuss, line 90: Looks like this is an existing pattern in this code base but this is unlikely to be "safe" for parallel test runs if we were to ever use those. Might be worth just adding a comment to that effect. You've done the right thing here @ronidelacruz-astro by adding the defer function to put the function back at the end of this test, which looks like it's missing from other tests.

Issue #1909, opened 2025-08-11
astro deployment pool copy fails on default_pool (missing update_mask)


bmanan7:
**Describe the bug**

When running astro deployment pool copy, the command fails when copying the default_pool from one deployment to another. The CLI attempts to update the existing default_pool in the target deployment and receives a 400 response from the Airflow API:
```
astro deployment pool copy --source-id cmXXXXXX --target-id cmeXXXXXXX

Copying Pool default_pool
Error: API error (400): {"detail":"Only slots and included_deferred can be modified on Default Pool"}
```

I also tried calling the Airflow API directly via curl to confirm, and the same error is returned when updating the default pool:
```
curl -X PATCH \
  "https://$AIRFLOW_URL/api/v2/pools/default_pool" \
  -H "Authorization: Bearer $AIRFLOW_API_KEY" \
  -H "Content-Type: application/json" \
  --data '{"slots":64,"description":"Default pool","include_deferred":false}'
# response:
{"detail":"Only slots and included_deferred can be modified on Default Pool"} 
```
Looking at the astro‑cli implementation, the CopyPool function iterates through all pools and calls airflowAPIClient.UpdatePool() whenever a pool with the same name already exists in the target deployment. It passes the whole pool object, including fields like Description and Name

If the pool is not there it will create it: https://github.com/astronomer/astro-cli/blob/212375d85b93420a91b9ea325ec9c9d082453343/cloud/deployment/deployment_objects.go#L293
 
This problem occurs when the pool is already there. So, in all the deployment we will always have `default_pool` as the user is not allowed to delete it from the UI. Everytime this command will fail with 400 error code as the core airflow code expects `update_mask` field to be present with allowed values(slots and include_deferred) only: https://github.com/astronomer/airflow/blob/3cb49c7c9a645add7d083564f8099f7059773b22/airflow-core/src/airflow/api_fastapi/core_api/routes/public/pools.py#L155
 
In Airflow API, as we can see the PATCH handler for pools explicitly prohibits changing the name or description of default_pool and requires an update_mask that only includes slots and include_deferred. Because the CLI does not handle this special case and does not set an update_mask, the API rejects the request.

**What CLI Version did you experience this bug?**

```
astro version
Astro CLI Version: 1.35.1
```

**This CLI bug is related to which Astronomer Platform?**

- [x] Astro
- [ ] Software
- [ ] None/Unknown

**What Operating System is the above CLI installed on?**

**🪜 Steps To Reproduce**

[...]

PR #1940, opened 2025-09-18
fix-1909: handle default_pool updates with slots/include_deferred mask

bmanan7:
## Description

- Update Airflow client to respect `default_pool` PATCH rules: send only `slots` and `include_deferred` with an update_mask.
- Added a regression test so future changes keep the special-case behavior intact.

## 🎟 Issue(s)

Related https://github.com/astronomer/astro-cli/issues/1909

## 🧪 Functional Testing
Tried copying pool with different slot values and number of pools and each request was successful. Attached screenshots from UI. 
```
./astro deployment pool copy --source-id cme0XXXXXX --target-id cme0XXXXXX

Using an Astro API Token
Copying Pool default_pool
Copying Pool test-pool
Copying Pool second_pool
==============
./astro deployment pool copy --source-id cme0XXXXXX --target-id cme0XXXXXX

Using an Astro API Token
Copying Pool default_pool
Copying Pool test-pool
Copying Pool second_pool
Copying Pool Fourth-pool

````

```
go test -count=1 ./airflow-client
ok  	github.com/astronomer/astro-cli/airflow-client	0.667s
```

## 📸 Screenshots
Confirmed from UI that the default pool slot value is being updated and reflected without any errors. 
image

image

## 📋 Checklist

- [ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer] (https://github.com/astronomer/astronomer/) (if necessary). - NA
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/) - NA

### DIFF
--- a/airflow-client/airflow-client.go
+++ b/airflow-client/airflow-client.go
@@ -181,21 +181,13 @@ func (c *HTTPClient) CreatePool(airflowURL string, pool Pool) error {
 }
 
 func (c *HTTPClient) UpdatePool(airflowURL string, pool Pool) error {
-	payload := interface{}(pool)
 	path := fmt.Sprintf("https://%s/pools/%s", airflowURL, pool.Name)
 
 	if pool.Name == "default_pool" {
-		payload = struct {
-			Slots           int  `json:"slots"`
-			IncludeDeferred bool `json:"include_deferred"`
-		}{
-			Slots:           pool.Slots,
-			IncludeDeferred: pool.IncludeDeferred,
-		}
 		path += "?update_mask=slots&update_mask=include_deferred"
 	}
 
-	varJSON, err := json.Marshal(payload)
+	varJSON, err := json.Marshal(pool)
 	if err != nil {
 		return err
 	}
--- a/airflow-client/airflow-client.go
+++ b/airflow-client/airflow-client.go
@@ -181,14 +181,27 @@ func (c *HTTPClient) CreatePool(airflowURL string, pool Pool) error {
 }
 
 func (c *HTTPClient) UpdatePool(airflowURL string, pool Pool) error {
-	// Convert the connection struct to JSON bytes
-	varJSON, err := json.Marshal(pool)
+	payload := interface{}(pool)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: airflow-client/airflow-client.go [hunk 1]
```diff
@@ -181,14 +181,27 @@ func (c *HTTPClient) CreatePool(airflowURL string, pool Pool) error {
...
+	path := fmt.Sprintf("https://%s/pools/%s", airflowURL, pool.Name)
+
+	if pool.Name == "default_pool" {
+		payload = struct {
+			Slots           int  `json:"slots"`
+			IncludeDeferred bool `json:"include_deferred"`
+		}{
+			Slots:           pool.Slots,
+			IncludeDeferred: pool.IncludeDeferred,
+		}
```
neel-astro: Correct me if I am wrong, but based on the airflow code, I think you just need to pass the update_mask in the path param for the default pool. So no need to exclusively send these two fields, airflow will take care of selecting these two fields from the request body based on the update_mask path param

bmanan7: Thanks for the suggestion. You’re right. Airflow handler only enforces that update_mask is supplied and constrained, it doesn’t object to extra fields in the body. I have updated code to keep it straight forward i.e just appending `update_mask=slots&update_mask=include_deferred` when the pool name is default_pool.

airflow-client/airflow-client.go
```diff
@@ -181,14 +181,19 @@ func (c *HTTPClient) CreatePool(airflowURL string, pool Pool) error {
 }
 
 func (c *HTTPClient) UpdatePool(airflowURL string, pool Pool) error {
-	// Convert the connection struct to JSON bytes
+	path := fmt.Sprintf("https://%s/pools/%s", airflowURL, pool.Name)
+
+	if pool.Name == "default_pool" {
+		path += "?update_mask=slots&update_mask=include_deferred"
+	}
```
neel-astro, line 189:
```suggestion
        // default pool does not allow updating other fields, such as description
	if pool.Name == "default_pool" {
		path += "?update_mask=slots&update_mask=include_deferred"
	}
```

PR #1945, opened 2025-09-26
[Fix]: per deployment docker login

pgvishnuram:
## Description

## 🎟 Issue(s)

https://github.com/astronomer/issues/issues/7903
https://github.com/astronomer/issues/issues/7944

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


software/deploy/deploy.go
```diff
@@ -210,20 +211,25 @@ func pushDockerImage(byoRegistryEnabled bool, deploymentInfo *houston.Deployment
...
 				return err
 			}
 			// Switch to per deployment registry login
-			err = auth.RegistryAuth(houstonClient, os.Stdout)
+			err = auth.RegistryAuth(houstonClient, os.Stdout, registry)
+			registryHandler, _ := registryHandlerInit(registry)
+			if err != nil {
+				return err
+			}
+			err = registryHandler.Login("user", token)
```
ianbuss:
```suggestion
```

pgvishnuram: thx for the catch we dont need them any more

software/auth/auth.go
```diff
@@ -99,19 +99,19 @@ func RegistryAuth(client houston.ClientInterface, out io.Writer) error {
 	}
 
 	var registry string
-	if versions.GreaterThanOrEqualTo(appConfig.Version, "1.0.0") && !appConfig.Flags.BYORegistryEnabled {
-		logger.Info("skipping registry login")
-		return nil
-	}
 
 	if appConfig.Flags.BYORegistryEnabled {
+		logger.Info(appConfig.BYORegistryDomain)
```
ianbuss: Remove this log or add additional context to it.

ianbuss (approved):
Approved with nit.

ianbuss:
> Approved with nit.

Any tests need updating/adding?

software/auth/auth.go
```diff
@@ -99,19 +99,22 @@ func RegistryAuth(client houston.ClientInterface, out io.Writer) error {
...
-	if appConfig.Flags.BYORegistryEnabled {
+	switch {
+	case appConfig.Flags.BYORegistryEnabled:
 		registry = appConfig.BYORegistryDomain
-	} else {
+		registryDomain = strings.Split(registry, "/")[0]
+	case versions.GreaterThanOrEqualTo(appConfig.Version, "1.0.0"):
+		if registry == "" {
+			return nil
+		}
```
ianbuss: I don't understand why this is added -- it will always be true since registry is defined as empty on line 101

pgvishnuram: @ianbuss no this will not always be empty 
case 1 - on first login it will be empty thats expected 
case 2 - on perfoming astro deploy https://github.com/astronomer/astro-cli/blob/feat/fix-perdb-registry-login/software/deploy/deploy.go#L220 this flow will be triggered and it  will pass the registry as string and this process continuous for every astro deploy

software/auth/auth.go
```diff
@@ -99,19 +99,22 @@ func RegistryAuth(client houston.ClientInterface, out io.Writer) error {
...
-	if appConfig.Flags.BYORegistryEnabled {
+	switch {
+	case appConfig.Flags.BYORegistryEnabled:
 		registry = appConfig.BYORegistryDomain
-	} else {
+		registryDomain = strings.Split(registry, "/")[0]
+	case versions.GreaterThanOrEqualTo(appConfig.Version, "1.0.0"):
+		if registryDomain == "" {
+			return nil
+		}
```
neel-astro: nit: to keep the logic simple, I would prefer this check/return at the start of the method, instead of having it from within a switch case.

software/auth/auth_test.go
```diff
@@ -243,7 +243,7 @@ func (s *Suite) TestRegistryAuthFailure() {
 		houstonMock := new(houstonMocks.ClientInterface)
 		houstonMock.On("GetAppConfig", "").Return(nil, errMockHouston).Once()
 
-		err := RegistryAuth(houstonMock, out)
+		err := RegistryAuth(houstonMock, out, "")
```
neel-astro, line 320: nit: todo: need to add test cases for the newly added logic

software/auth/auth.go
neel-astro: But also, if I understand it correctly, there is no case when the app version is greater than 1.0.0 and the registry domain would be an empty string, so not sure this even needs to be here

software/auth/auth.go
ianbuss: To close the loop on this conversation, the code was referencing the wrong variable name `registry` vs `registryDomain`.

software/auth/auth_test.go
ianbuss: Added happy path in https://github.com/astronomer/astro-cli/pull/1945/commits/ee88fe2be0c49c03afc2fc3621231e6d991a8548

Issue #1690, opened 2024-07-26
Debug output to show the docker binary path


stuart23:
- [x] I have checked that a similar [feature request](https://github.com/astronomer/astro-cli/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22) does not already exist.

**✍️ Is your feature request related to a problem? Please describe.**

Cannot tell which `docker` binary is getting called on systems that have multiple docker (podman, rancher etc) flavors.

**🧩 Describe the solution you'd like**

Just output the path to the binary that will be called. This can be just when verbose mode is used if necessary.

**🤔 Describe alternatives you've considered**

**Is your feature request specific to a particular Astronomer Platform?**

- [ ] Astro
- [ ] Software
- [x] None/Unknown

**💡 Additional context**

stuart23:
https://github.com/astronomer/astro-cli/issues/1501

PR #1968, opened 2025-10-22
Verbosity flag provides no extra information (#1501)

### LINKED ISSUE #1501: Verbosity flag provides no extra information
**Describe the bug**

In debugging a CLI error that produces `Error: workspace id is not valid: cannot retrieve organization short name from context`, adding `--verbosity debug` to see the context for which the org short name is missing doesn't actually do anything.

For instance: `astro workspace list --verbosity debug` gives the exact same output as `astro workspace list`.

We need to be able to see the context to be able to debug this issue.

1.21.0

- [x] Astro
- [ ] Software
- [ ] None/Unknown

MacOS

**🪜 Steps To Reproduce**

**📸 Screenshots**

ronidelacruz-astro:
## Description

Fix `--verbosity debug` flag not working in Homebrew-installed and GitHub release binaries.

The `--verbosity` flag was not functioning in binaries built with GoReleaser (used for Homebrew and GitHub releases). When users ran commands like `astro workspace list --verbosity debug`, no debug output was displayed, even though the flag was recognized and validated.

**Root Cause:** GoReleaser uses the `-X` linker flag to inject version information at compile time. This interferes with cobra's package-level variable binding mechanism, preventing the `verboseLevel` variable from receiving the flag value.

**Solution:** Modified `SetupLogging` in `cmd/root_hooks.go` to read the flag value directly from the cobra command object instead of relying on the package-level variable. This bypasses the broken binding and ensures the logger is correctly configured in all build environments.

**Impact:** 
- Fixes verbosity flag for all Homebrew installations
- Fixes verbosity flag for all GitHub release binaries
- Maintains backward compatibility with source builds
- Low risk - defensive change with fallback to original behavior

## 🎟 Issue(s)
Closes #1501 #1690 

## 🧪 Functional Testing

**A/B Testing with Homebrew Build Flags:**
- Built original code with GoReleaser flags: `go build -ldflags="-s -w -X github.com/astronomer/astro-cli/version.CurrVersion=1.37.0"` → No debug output
- Built fixed code with same flags → Debug output works correctly

## 📋 Checklist
- [ ] Tested against Astro-API (if necessary) - N/A
- [ ] Tested against Houston-API and Astronomer (if necessary) - N/A
- [x] Communicated to/tagged owners of respective clients potentially impacted by these changes
- [ ] Updated any related documentation - Will update CLI docs if needed

### DIFF
--- a/cloud/workspace/workspace.go
+++ b/cloud/workspace/workspace.go
@@ -14,6 +14,7 @@ import (
 	"github.com/astronomer/astro-cli/context"
 	"github.com/astronomer/astro-cli/pkg/ansi"
 	"github.com/astronomer/astro-cli/pkg/input"
+	"github.com/astronomer/astro-cli/pkg/logger"
 	"github.com/astronomer/astro-cli/pkg/printutil"
 )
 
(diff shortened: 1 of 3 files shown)


neel-astro (commented):
I believe the problem isn't that the verbosity flag is not set up correctly with those issues, but rather, we do not have any debug logs added for these commands.

Issue #1979, opened 2025-10-31
`astro remote deploy` incorrectly uses `packages.txt` and `requirements.txt` for installing dependencies


feluelle:
**Describe the bug**

When running `astro remote deploy`, it uses `packages.txt` and `requirements.txt` for installing dependencies. It should use `packages-client.txt` and `requirements-client.txt`.

**What CLI Version did you experience this bug?**

1.37.0

**This CLI bug is related to which Astronomer Platform?**

- [ ] Astro
- [ ] Software
- [ ] None/Unknown

**What Operating System is the above CLI installed on?**

**🪜 Steps To Reproduce**

**📸 Screenshots**

PR #1980, opened 2025-11-03
Fix `astro remote deploy` to consider client dependency files for installation during image build

feluelle:
## Description

This fixes the issue in the `astro remote deploy` command, where it does not consider the `requirements-client.txt` and `packages-client.txt` files for dependency installation into the image.

## 🎟 Issue(s)

Fixes: https://github.com/astronomer/astro-cli/issues/1979

## 🧪 Functional Testing

```
❯ cat requirements-client.txt
apache-airflow-providers-amazon
```
```
❯ ../astro remote deploy
Authenticating with base image registry: images.astronomer.cloud
Building client image for host platform
...
```
Validated that the dependencies get installed correctly from client files:
```
❯ docker run --entrypoint pip ... list | grep amazon
...
apache-airflow-providers-amazon          9.16.0
```

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -807,6 +807,78 @@ func WarnIfNonLatestVersion(version string, httpClient *httputil.HTTPClient) {
 	}
 }
 
+// ClientBuildContext represents a prepared build context for client deployment
+type ClientBuildContext struct {
+	// TempDir is the temporary directory containing the build context
+	TempDir string
+	// CleanupFunc should be called to clean up the temporary directory
+	CleanupFunc func()
+}
+
+// prepareClientBuildContext creates a temporary build context with client dependency files
+// This avoids modifying the original project files, preventing race conditions with concurrent deployments.
+func prepareClientBuildContext(sourcePath string) (*ClientBuildContext, error) {
+	// Create a temporary directory for the build context
+	tempBuildDir, err := os.MkdirTemp("", "astro-client-build-*")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temporary build directory: %w", err)
+	}
+
+	// Cleanup function to be called by the caller
+	cleanup := func() {
+		os.RemoveAll(tempBuildDir)
+	}
+
+	// Always return cleanup function if we created a temp directory, even on error
+	buildContext := &ClientBuildContext{
+		TempDir:     tempBuildDir,
+		CleanupFunc: cleanup,
+	}
+
+	// Check if source directory exists first
+	if exists, err := fileutil.Exists(sourcePath, nil); err != nil {
+		return buildContext, fmt.Errorf("failed to check if source directory exists: %w", err)
+	} else if !exists {
+		return buildContext, fmt.Errorf("source directory does not exist: %s", sourcePath)
+	}
+
+	// Copy all project files to the temporary directory
+	err = fileutil.CopyDirectory(sourcePath, tempBuildDir)
+	if err != nil {
+		return buildContext, fmt.Errorf("failed to copy project files to temporary directory: %w", err)
+	}
+
+	// Process client dependency files
+	err = setupClientDependencyFiles(tempBuildDir)
+	if err != nil {
+		return buildContext, fmt.Errorf("failed to setup client dependency files: %w", err)
+	}
+
+	return buildContext, nil
+}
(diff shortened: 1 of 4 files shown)


cloud/deploy/deploy.go
```diff
@@ -807,6 +807,101 @@ func WarnIfNonLatestVersion(version string, httpClient *httputil.HTTPClient) {
...
+		clientPath := filepath.Join(buildDir, clientFile)
+		regularPath := filepath.Join(buildDir, regularFile)
+
+		// Check if client file exists and has content
+		clientExists := false
+		if info, err := os.Stat(clientPath); err == nil && info.Size() > 0 {
+			content, err := os.ReadFile(clientPath)
+			if err == nil && strings.TrimSpace(string(content)) != "" {
+				clientExists = true
+			}
```
neel-astro: I think we can even assume an empty client file as valid, i.e., clientExists as true

cloud/deploy/deploy.go
```diff
@@ -807,6 +807,101 @@ func WarnIfNonLatestVersion(version string, httpClient *httputil.HTTPClient) {
...
+	}
+
+	// Cleanup function to be called by the caller
+	cleanup := func() {
+		os.RemoveAll(tempBuildDir)
+	}
+
+	// Check if source directory exists first
+	if exists, err := fileutil.Exists(sourcePath, nil); err != nil {
+		cleanup()
```
neel-astro: I think we can make this simpler by having the caller handle this cleanup entirely, instead of having to call it for each error case here, as well as having the caller handle it

cloud/deploy/deploy.go
```diff
@@ -879,8 +974,16 @@ func DeployClientImage(deployInput InputClientDeploy) error { //nolint:gocritic
 			fmt.Println("Building client image for host platform")
 		}
 
+		// Prepare build context with client dependency files
+		buildContext, err := prepareClientBuildContext(deployInput.Path)
+		if err != nil {
+			return fmt.Errorf("failed to prepare client build context: %w", err)
+		}
+		defer buildContext.CleanupFunc()
```
neel-astro:
```suggestion
		defer buildContext.CleanupFunc()
		if err != nil {
			return fmt.Errorf("failed to prepare client build context: %w", err)
		}
```
nit: to call the cleanup even in the error case

cloud/deploy/deploy.go
feluelle: I thought the same, but no, we can't. Because there's nothing to clean up in the case of the first error here:
```
	// Create a temporary directory for the build context
	tempBuildDir, err := os.MkdirTemp("", "astro-client-build-*")
	if err != nil {
		return nil, fmt.Errorf("failed to create temporary build directory: %w", err)
	}
```
I was also about to blame the AI, but then realised that this is correct.

cloud/deploy/deploy.go
feluelle: See https://github.com/astronomer/astro-cli/pull/1980#discussion_r2486234556. This cleanup here is purely for the success case, where the build context has been set up correctly, but we clean up after the image build or when the image build itself fails.

cloud/deploy/deploy.go
neel-astro: I think it is ok to even call cleanup in that case, because os.RemoveAll will handle that for us internally (if path does not exist then it fails silently) - https://pkg.go.dev/os#RemoveAll

feluelle: I've found a better way.

cloud/deploy/deploy.go
```diff
@@ -879,8 +951,18 @@ func DeployClientImage(deployInput InputClientDeploy) error { //nolint:gocritic
 			fmt.Println("Building client image for host platform")
 		}
 
+		// Prepare build context with client dependency files
+		buildContext, err := prepareClientBuildContext(deployInput.Path)
+		if buildContext != nil {
+			defer buildContext.CleanupFunc()
+		}
```
neel-astro:
```suggestion
		if buildContext != nil && buildContext.CleanFunc != nil {
			defer buildContext.CleanupFunc()
		}
```
nit: double safety from panic :P

neel-astro (approved):
LGTM

Issue #1957, opened 2025-10-10
Reject `astro remote deploy`s for too-new agent clients


feluelle:
At the time of writing this issue, the remote deploy command does not consume information about the Remote Deployment. We could add a `--deployment-id` flag to read the deployment info, and if the image defined in `Dockerfile.client` is based on an Astro Runtime version newer of what the Remote Deployment uses, we should reject the deploy, and tell the user to either downgrade the client version, or upgrade the Remote Deployment version.

related to: https://github.com/astronomer/astro-agent/issues/673

danalogalbo-astronomer:
https://linear.app/astronomer/issue/AINF-1068/reject-astro-remote-deploys-for-too-new-agent-clients

PR #1983, opened 2025-11-05
feat: add runtime version validation for remote client deployments

feluelle:
## Description

Prevent deployment of client images with runtime versions newer than the target deployment's runtime version to ensure compatibility.

Changes:
- Add --deployment-id flag to 'astro remote deploy' command for validation
- Implement validateClientImageRuntimeVersion() function that:
  - Fetches deployment runtime version from platform API
  - Parses Dockerfile.client to extract client image runtime version
  - Rejects deployment if client version > deployment version
- Enhance ParseImageTag() function to extract runtime versions from image tags like '3.1-1-python-3.12-astro-agent-1.1.0'
- Add comprehensive test coverage for all validation scenarios
- Update docker parsing to support Dockerfile.client files

This ensures client images are compatible with the deployment environment and prevents runtime incompatibility issues.

## 🎟 Issue(s)

closes: https://github.com/astronomer/astro-cli/issues/1957

## 🧪 Functional Testing

## 📸 Screenshots

```
❯ ../astro remote deploy --deployment-id cmieixby600r601mx9snv8pf1
Error: client image runtime version validation failed:

The client image is based on Astro Runtime version 3.1-5, which is newer than the deployment's runtime version 3.1-4.

To fix this issue, you can either:
1. Downgrade the client image version by updating the base image in Dockerfile.client to use runtime version 3.1-4 or earlier
2. Upgrade the deployment's runtime version to 3.1-5 or later

This validation ensures compatibility between your client image and the deployment environment.
```

## 📋 Checklist

### DIFF
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -465,7 +465,7 @@ func getDeploymentInfo(
 	if err != nil {
 		return deploymentInfo{}, err
 	}
-	deployInfo, err := getImageName(deploymentID, c.Organization, platformCoreClient)
+	deployInfo, err := fetchDeploymentDetails(deploymentID, c.Organization, platformCoreClient)
 	if err != nil {
 		return deploymentInfo{}, err
 	}
@@ -547,7 +547,7 @@ func checkPytest(pytest, deployImage, buildSecretString string, containerHandler
 	return err
 }
 
-func getImageName(deploymentID, organizationID string, platformCoreClient astroplatformcore.CoreClient) (deploymentInfo, error) {
+func fetchDeploymentDetails(deploymentID, organizationID string, platformCoreClient astroplatformcore.CoreClient) (deploymentInfo, error) {
 	resp, err := platformCoreClient.GetDeploymentWithResponse(httpContext.Background(), organizationID, deploymentID)
 	if err != nil {
 		return deploymentInfo{}, err
@@ -1018,7 +1018,7 @@ func validateClientImageRuntimeVersion(deployInput InputClientDeploy, platformCo
 	}
 
 	// Get deployment information
-	deployInfo, err := getImageName(deployInput.DeploymentID, c.Organization, platformCoreClient)
+	deployInfo, err := fetchDeploymentDetails(deployInput.DeploymentID, c.Organization, platformCoreClient)
 	if err != nil {
 		return errors.Wrap(err, "failed to get deployment information")
 	}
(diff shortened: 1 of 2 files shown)


cloud/deploy/deploy.go
```diff
@@ -996,3 +1002,89 @@ func DeployClientImage(deployInput InputClientDeploy) error { //nolint:gocritic
...
+	baseImage := docker.GetImageFromParsedFile(cmds)
+	if baseImage == "" {
+		return errors.New("failed to find base image in Dockerfile.client")
+	}
+
+	// Extract runtime version from the base image tag
+	clientRuntimeVersion, err := extractRuntimeVersionFromImage(baseImage)
+	if err != nil {
+		// If we can't extract version, skip validation with a warning
+		fmt.Printf("Warning: Could not extract runtime version from client image %s, skipping version validation\n", baseImage)
```
neel-astro: nit: if we are simply warning in case of error over here, then maybe it makes sense to do the same for other errors before this point as well.

cloud/deploy/deploy.go
```diff
@@ -996,3 +1002,89 @@ func DeployClientImage(deployInput InputClientDeploy) error { //nolint:gocritic
...
+
+	// Compare versions
+	if airflowversions.CompareRuntimeVersions(clientRuntimeVersion, deployInfo.currentVersion) > 0 {
+		return fmt.Errorf(`client image runtime version validation failed:
+
+The client image is based on Astro Runtime version %s, which is newer than the deployment's runtime version %s.
+
+To fix this issue, you can either:
+1. Downgrade the client image version by updating the base image in Dockerfile.client to use runtime version %s or earlier
+2. Upgrade the deployment's runtime version to %s or later
```
neel-astro:
```suggestion
2. Upgrade the deployment's runtime version to %s or higher
```

cloud/deploy/deploy.go
```diff
@@ -996,3 +1002,89 @@ func DeployClientImage(deployInput InputClientDeploy) error { //nolint:gocritic
...
+	}
+
+	// Get current context for organization info
+	c, err := config.GetCurrentContext()
+	if err != nil {
+		return errors.Wrap(err, "failed to get current context")
+	}
+
+	// Get deployment information
+	deployInfo, err := getImageName(deployInput.DeploymentID, c.Organization, platformCoreClient)
```
neel-astro: `getImageName` is a very confusing name for what the method is doing. Should we fix/update the name?

cloud/deploy/deploy.go
feluelle: Good point, but actually, I have changed it to always error since the whole point for the user to define the deployment ID is to use validation, no matter what. If it errors, it should not allow deploying.

PR #1990, opened 2025-12-15
feat: add git metadata to deploys via v1alpha1 API

jlaneve:
## Summary

- Adds support for including git metadata (commit SHA, branch, author, etc.) in deploy requests
- Uses the v1alpha1 API which supports the git metadata field, providing better traceability for deployments
- Reuses existing `retrieveLocalGitMetadata()` logic from bundle deploys for consistency

Tested this against dev, where the API changes are currently deployed and it works (screenshot below). Tested this against prod, where the API changes are not yet deployed, and we disregard the extra fields so no harm in merging this earlier.

Screenshot 2025-12-18 at 1 16 06 PM

## Changes

- Add `deploy.git_metadata` config option (default: `true`)
- Add `ASTRO_DEPLOY_GIT_METADATA` env var support for CI/CD overrides
- Add `createDeployWithGit()` function using v1alpha1 core API
- Fall back to git commit message for deploy description if not provided

## Behavior

When enabled (default), deploys will:
- Include git metadata (provider, account, repo, branch, commit SHA, commit URL, author)
- Auto-populate description from commit message if not specified
- Skip git metadata if repository has uncommitted changes (matches bundle deploy behavior)

Users can opt out:
```bash
# Via config
astro config set deploy.git_metadata false

# Via environment variable
ASTRO_DEPLOY_GIT_METADATA=false astro deploy
```

## Test plan

- [x] Existing deploy tests pass
- [x] Manual test: deploy with git metadata enabled
- [x] Manual test: deploy with git metadata disabled
- [x] Manual test: deploy with uncommitted changes (should skip git metadata)
- [x] Manual test: deploy without description (should use commit message)

### DIFF
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -17,7 +17,6 @@ import (
 	"github.com/astronomer/astro-cli/pkg/fileutil"
 	"github.com/astronomer/astro-cli/pkg/git"
 	"github.com/astronomer/astro-cli/pkg/logger"
-	"github.com/astronomer/astro-cli/pkg/util"
 )
 
 type DeployBundleInput struct {
@@ -57,11 +56,7 @@ func DeployBundle(input *DeployBundleInput) error {
 	// Check if git metadata is enabled (default: true)
 	var deployGit *astrocore.DeployGit
 	var commitMessage string
-	gitMetadataEnabled := config.CFG.DeployGitMetadata.GetBool()
-	if envVal := os.Getenv("ASTRO_DEPLOY_GIT_METADATA"); envVal != "" {
-		gitMetadataEnabled = util.CheckEnvBool(envVal)
-	}
-	if gitMetadataEnabled {
+	if config.CFG.DeployGitMetadata.GetBool() {
 		deployGit, commitMessage = retrieveLocalGitMetadata(input.BundlePath)
 	}
 
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -242,11 +242,7 @@ func Deploy(deployInput InputDeploy, platformCoreClient astroplatformcore.CoreCl
 	// Check if git metadata is enabled (default: true)
 	var deployGit *astrocore.DeployGit
 	var commitMessage string
-	gitMetadataEnabled := config.CFG.DeployGitMetadata.GetBool()
-	if envVal := os.Getenv("ASTRO_DEPLOY_GIT_METADATA"); envVal != "" {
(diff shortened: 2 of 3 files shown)


cloud/deploy/bundle.go
```diff
@@ -53,8 +54,16 @@ func DeployBundle(input *DeployBundleInput) error {
 		return fmt.Errorf(enableDagDeployMsg, input.DeploymentID)
 	}
 
-	// retrieve metadata about the local Git checkout. returns nil if not available
-	deployGit, commitMessage := retrieveLocalGitMetadata(input.BundlePath)
+	// Check if git metadata is enabled (default: true)
+	var deployGit *astrocore.DeployGit
+	var commitMessage string
+	gitMetadataEnabled := config.CFG.DeployGitMetadata.GetBool()
+	if envVal := os.Getenv("ASTRO_DEPLOY_GIT_METADATA"); envVal != "" {
```
neel-astro: The config is enabled by default, so why do we need a separate env var to turn it off, apart from being able to set it via config?
i.e., for anyone who does not need it can, as of now, do it via `astro config set -g deploy.git_metadata false` in CI/CD as well.

cloud/deploy/bundle.go
```diff
@@ -251,12 +260,12 @@ func createBundleDeploy(organizationID string, input *DeployBundleInput, deployG
...
 		request.Git = &astrocore.CreateDeployGitRequest{
 			Provider:   astrocore.CreateDeployGitRequestProvider(deployGit.Provider),
-			Repo:       deployGit.Repo,
 			Account:    deployGit.Account,
+			Repo:       deployGit.Repo,
 			Path:       deployGit.Path,
 			Branch:     deployGit.Branch,
 			CommitSha:  deployGit.CommitSha,
-			CommitUrl:  fmt.Sprintf("https://github.com/%s/%s/commit/%s", deployGit.Account, deployGit.Repo, deployGit.CommitSha),
+			CommitUrl:  deployGit.CommitUrl,
```
neel-astro, line 263: good cleanup ❤️

cloud/deploy/deploy.go
```diff
@@ -369,7 +406,7 @@ func Deploy(deployInput InputDeploy, platformCoreClient astroplatformcore.CoreCl
 			fmt.Println("No DAGs found. Skipping testing...")
 		}
 
-		repository := deploy.ImageRepository
+		repository := imageRepository
```
neel-astro: nit: we may not need this. Declaring the variable and using it only in one place, that too quite far apart, reduces overall code readability.

PR #2000, opened 2026-01-21
fix: allow org switch by ID when org not in paginated list

jlaneve:
## Summary
- Fixes an issue where users with access to more than 100 organizations couldn't switch to orgs not appearing in the first page of results
- Adds explicit pagination limit (100) to `ListOrganizations`
- Adds new `GetOrganization` function to fetch an org directly by ID
- Adds fallback in `Switch` to try fetching by ID when org isn't found in the list

## Test plan
- [x] Unit tests pass (`go test ./cloud/organization/...`)
- [x] New test case added for the fallback scenario

### DIFF
--- a/cmd/cloud/setup_test.go
+++ b/cmd/cloud/setup_test.go
@@ -443,6 +443,9 @@ func TestCheckAPIToken(t *testing.T) {
 			Organizations: []astroplatformcore.Organization{
 				{Name: "test-org", Id: "test-org-id", Product: &mockOrganizationProduct},
 			},
+			TotalCount: 1,
+			Limit:      100,
+			Offset:     0,
 		},
 	}
 	permissions = []string{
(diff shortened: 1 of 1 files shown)


cloud/organization/organization.go
```diff
@@ -173,6 +188,14 @@ func Switch(orgNameOrID string, coreClient astrocore.CoreClient, platformCoreCli
 				targetOrg = &or[i]
 			}
 		}
+		// If not found in list, try to get org by ID directly
```
jeremybeard: It's unfortunate that we have this ambiguous value that could be a name or an ID. I'd prefer if we could deal with this more systematically than a hail mary get org call if the first page of list orgs doesn't produce it. Can we look at the value and if it is a CUID then we do a get org call, otherwise we paginate through list orgs responses until we find it? Then we avoid unnecessary lookups, and if we don't find it we know it is really not there instead of maybe being on another page.

PR #2001, opened 2026-01-27
fix: add --no-dags-base-dir flag for Airflow 3 DAG bundle compatibility

### LINKED ISSUE #1985: DAG folder not on python path in airflow 3.1
**Describe the bug**

 A common pattern we use is packaging extra python files up with the dags for use in multiple dags within a single project.  In airflow 2 this worked, but with the introduction of dag bundles in airflow 3.1 it has become broken.  I've been looking a bit at the code and specifically https://github.com/apache/airflow/pull/50385 where the dags folder has been removed from the path and replaced with the bundle root.  So in my deployment `/tmp/airflow/dag_bundles/astro/main/dags/2025-11-07T15:28:12.8895389Z/` ends up on the path.  However in astronomer this bundle root isn't the same as the dags folder, it contains the dags folder `/tmp/airflow/dag_bundles/astro/main/dags/2025-11-07T15:28:12.8895389Z/dags`.  To me this feels like a bug, when astronomer packaged up the dag folder for deployment, it should have done so such the dags end up at the root of the bundle folder, rather than in the dags subdirectory. 

For reference I'm running astronomer 3.1-3, using dag only deploys and 1.38.1 of the astro-cli

```
Please paste terminal or playground output in a code block here
```

1.38.1

- [x] Astro
- [ ] Software
- [ ] None/Unknown

**🪜 Steps To Reproduce**

**📸 Screenshots**

tayloramurphy:
## Summary

- Adds `--no-dags-base-dir` flag to `astro deploy --dags` for Airflow 3.x compatibility
- By default, DAG bundles include the `dags/` directory prefix (existing behavior)
- With `--no-dags-base-dir`, DAG files are placed at the bundle root

## Usage

```bash
# Default (backward compatible)
astro deploy --dags

# For Airflow 3.x deployments
astro deploy --dags --no-dags-base-dir
```

## Context

Airflow 3 changed the DAG processor to add the bundle root to `sys.path` instead of the `dags/` folder. This caused imports from shared modules in the `dags/` directory to fail for DAG-only deploys.

This flag gives users explicit control over the bundle layout without changing default behavior.

Fixes #1985

### DIFF
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -181,7 +181,7 @@ func deployDags(path, dagsPath, dagsUploadURL, currentRuntimeVersion string, dep
 	}
 
 	// By default, prepend dags/ directory prefix. Use --no-dags-base-dir to place files at bundle root
-	// (required for Airflow 3.x where sys.path includes the bundle root, not dags/).
+	// (needed for some Airflow 3.x deployments where sys.path includes the bundle root, not dags/).
 	prependBaseDir := !noDagsBaseDir
 	versionID, err := UploadBundle(path, dagsPath, dagsUploadURL, prependBaseDir, currentRuntimeVersion)
 	if err != nil {


cloud/deploy/deploy.go
```diff
@@ -179,7 +180,10 @@ func deployDags(path, dagsPath, dagsUploadURL, currentRuntimeVersion string, dep
 		defer os.Remove(monitoringDagPath)
 	}
 
-	versionID, err := UploadBundle(path, dagsPath, dagsUploadURL, true, currentRuntimeVersion)
+	// By default, prepend dags/ directory prefix. Use --no-dags-base-dir to place files at bundle root
+	// (required for Airflow 3.x where sys.path includes the bundle root, not dags/).
```
jeremybeard, line 184: I wouldn't say it's necessarily required, most users on Airflow 3 have been fine as it currently is

cloud/deploy/deploy_test.go
```diff
@@ -2178,3 +2178,29 @@ func TestValidateClientImageRuntimeVersion(t *testing.T) {
 		assert.NoError(t, err)
 	})
 }
+
+func TestDeployDagsBundleLayout(t *testing.T) {
```
jeremybeard, line 2182: This is really just testing that the negation operator still works but I guess you can never be too careful

PR #2006, opened 2026-02-05
Add `astro api` command

jeremybeard:
## Description

Add the `astro api` top-level command with two subcommands for making authenticated API requests to Astronomer services:

**`astro api airflow`** — make requests to the Airflow REST API (local or deployed), with automatic version detection and OpenAPI spec resolution.

**`astro api cloud`** — make requests to the Astro Cloud platform API using the current context's bearer token.

Both subcommands support:
- Endpoint discovery via `ls` and `describe` (parsed from OpenAPI specs)
- Calling endpoints by operation ID or raw path
- Pagination (per-page streaming or `--slurp` into a single array)
- `jq` filters and Go-template output formatting
- Colored JSON output
- `--curl` flag to emit equivalent curl commands
- Magic field syntax for request bodies (`@file`, `:=json`, `=string`)
- Custom headers and path-parameter overrides

**Supporting packages:**
- `pkg/openapi`: OpenAPI spec fetching, caching, endpoint indexing, Airflow version-range mapping, and schema introspection.

## 🧪 Functional Testing

1. **Airflow local**: Run `astro dev start`, then `astro api airflow ls` to list available endpoints. Call an endpoint by operation ID, e.g. `astro api airflow get_health`.
2. **Airflow deployed**: Run `astro api airflow -d <deployment-id> ls` to list endpoints on a deployed Airflow instance. Call an endpoint, e.g. `astro api airflow -d <deployment-id> get_dags`.
3. **Cloud API**: Run `astro api cloud ls` to list Astro platform endpoints. Call an endpoint, e.g. `astro api cloud list-deployments --path-param organizationId=<org-id>`.
4. **Describe**: Run `astro api airflow describe get_dags` or `astro api cloud describe list-deployments` to see endpoint details including parameters and response schemas.
5. **Output formatting**: Test `--jq '.dags[]'`, `--template '{{range .dags}}{{.dag_id}}{{end}}'`, `--silent`, and `--slurp` flags.
6. **Curl generation**: Add `--curl` to any request to verify the generated curl command.
7. **Error handling**: Call a nonexistent endpoint and verify the CLI exits non-zero with the error body printed.

## 📋 Checklist

### DIFF
--- a/cmd/api/api_test.go
+++ b/cmd/api/api_test.go
@@ -12,19 +12,19 @@ import (
 
 // newRootWithAPI builds a minimal root -> api command tree for testing.
 // The root command's PersistentPreRunE is set to rootHook.
-func newRootWithAPI(rootHook func(cmd *cobra.Command, args []string) error) (*cobra.Command, *cobra.Command) {
+func newRootWithAPI(rootHook func(cmd *cobra.Command, args []string) error) *cobra.Command {
 	root := &cobra.Command{
 		Use:               "astro",
 		PersistentPreRunE: rootHook,
 	}
 	apiCmd := NewAPICmdWithOutput(new(bytes.Buffer))
 	root.AddCommand(apiCmd)
-	return root, apiCmd
+	return apiCmd
 }
 
 func TestPersistentPreRunE_CallsRootHook(t *testing.T) {
 	called := false
-	_, apiCmd := newRootWithAPI(func(cmd *cobra.Command, args []string) error {
+	apiCmd := newRootWithAPI(func(cmd *cobra.Command, args []string) error {
 		called = true
 		return nil
 	})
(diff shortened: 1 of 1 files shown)


pkg/openapi/schema.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if schema.Ref != "" {
+		refName := extractRefName(schema.Ref)
+		resolved := r.lookupSchema(refName)
+		if resolved != nil {
+			return resolved, refName
+		}
+	}
+
+	return schema, ""
```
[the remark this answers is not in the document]
jeremybeard: I think this is fine as it is

pkg/openapi/types.go
```diff
@@ -0,0 +1,120 @@
+// Package openapi provides utilities for working with OpenAPI specifications.
```
jlaneve, line 1: chatted about this offline but it'd be good to see if we can find a package for this instead of maintaining it ourselves

jeremybeard: Switched it to kin-openapi, which was actually already a transitive dependency

PR #2007, opened 2026-02-06
Update code owners

jeremybeard:
Moves the code owners to Astronomer-controlled teams instead of listing individuals.


neel-astro (approved):
Thanks for cleaning this up

PR #2012, opened 2026-02-13
feat: add standalone mode for Docker-free local development

jlaneve:
## Summary

Adds a **standalone mode** for running Airflow locally without Docker, unified into the existing `astro dev` command tree via mode-based dispatch.

### How it works

- A `dev.mode` project config (`"docker"` or `"standalone"`) controls which backend is used
- `--standalone` / `--docker` persistent flags allow per-call overrides
- All existing commands (`start`, `stop`, `logs`, `ps`, `kill`, `restart`) work in both modes
- `run`, `bash`, `pytest`, `parse`, `object import`, `object export` all work in standalone mode
- `logs` supports component filtering (`--scheduler`, `--api-server`, etc.) in both modes
- `build`, `upgrade-test`, and `compose-export` return clear "not available" errors in standalone mode

### Key changes

- **Config**: `dev.mode` config with `"docker"` default, validated to accept only `"docker"` or `"standalone"`
- **Flags**: `--standalone`/`--docker` mutually exclusive persistent flags on the `dev` root command
- **Dispatch**: `resolveDevMode()` resolves mode (flag → config priority), `resolveHandlerInit()` returns the right handler
- **Hooks**: `ConfigureContainerRuntime` assigns a `noOpContainerRuntime` in standalone mode, so downstream hooks (`EnsureRuntime`, `SetRuntimeIfExists`, `KillPreRunHook`, `KillPostRunHook`) work without nil-pointer guards or mode checks
- **Start/Restart**: `--foreground/-f` and `--port/-p` flags added for standalone mode
- **Grouped help**: `start` and `restart` use a custom usage template that organizes flags into Common / Docker Mode / Standalone Mode sections
- **Log filtering**: `astro dev logs` component flags (`--scheduler`, `--api-server`, `--triggerer`, `--dag-processor`, `--webserver`) filter standalone log lines by prefix. `--webserver` maps to `api-server` since AF3 standalone has no separate webserver. Unfiltered shows all lines including standalone orchestrator messages.
- **Port conflict detection**: Pre-flight TCP check before starting — `airflow standalone` doesn't crash when the port is taken (only the api-server subprocess fails), so without this check the health check would pass against whichever service already occupies the port
- **Custom port**: Uses `AIRFLOW__API__PORT` (AF3 config key) instead of the deprecated `AIRFLOW__WEBSERVER__WEB_SERVER_PORT`, reads from both `--port` flag and `api-server.port` config
- **Command dispatch**: `build` and `upgrade-test` now route through `resolveHandlerInit()` instead of hardcoded `containerHandlerInit`, so they properly error in standalone mode instead of silently running Docker operations
- **Removed**: The separate `astro dev local` command tree is replaced by the unified approach

### Standalone command implementations

| Command | Standalone behavior |
|---------|-------------------|
| `astro dev run <cmd>` | Executes command in the venv environment (e.g. `airflow dags list`) |
| `astro dev bash` | Opens interactive shell with venv env (`AIRFLOW_HOME`, `PATH` set) |
| `astro dev object import` | Imports connections/variables/pools from `airflow_settings.yaml` via venv |
| `astro dev object export` | Exports to settings file or `.env` via venv |
| `astro dev pytest` | Runs pytest on DAGs with file path resolution and args passthrough |
| `astro dev parse` | Validates DAGs by running the default integrity test |
| `astro dev logs` | Streams log file with optional component filtering |
| `astro dev build` | Returns clear "not available in standalone mode" error |
| `astro dev compose-export` | Returns clear "not available in standalone mode" error |
| `astro dev upgrade-test` | Returns clear "not available in standalone mode" error |

### Usage

**Setting the mode via config:**

```bash
# Set standalone as the default mode for a project
$ astro config set dev.mode standalone
Setting dev.mode to standalone successfully

# Check the current mode
$ astro config get dev.mode
dev.mode: standalone

# Switch back to Docker mode
$ astro config set dev.mode docker
Setting dev.mode to docker successfully

# Invalid values are rejected
$ astro config set dev.mode invalid
Error: invalid value for dev.mode: dev.mode must be "docker" or "standalone", got "invalid"
```

**Using per-command flags:**

```bash
# Override mode for a single command
astro dev start --standalone
astro dev start --standalone --foreground --port 9090
astro dev logs --standalone -f
astro dev stop --standalone
astro dev kill --standalone
```

**Standalone-specific commands:**

```bash
# Run airflow CLI commands
astro dev run dags list
astro dev run connections list
astro dev run version

# Open an interactive shell with venv environment
astro dev bash

# Import/export airflow objects
astro dev object import
astro dev object export
astro dev object export --env-export

# Run tests
astro dev pytest
astro dev pytest dags/test_my_dag.py
astro dev pytest --args "-v --tb=short"
astro dev parse

# Filter logs by component
astro dev logs -f --scheduler
astro dev logs --api-server
astro dev logs --triggerer --dag-processor
astro dev logs --webserver  # maps to api-server in standalone
```

### What the flags look like

```
$ astro dev start --help

Start a local Airflow environment. In Docker mode (default), this spins up
containers for each Airflow component. In standalone mode, this runs Airflow
directly on your machine without Docker.

Usage:
  astro dev start [flags]

Flags:
  -d, --deployment-id string   ID of the Deployment to retrieve environment connections from
  -e, --env string             Location of file containing environment variables (default ".env")
  -h, --help                   help for start
  -s, --settings-file string   Settings file from which to import airflow objects (default "airflow_settings.yaml")
      --wait duration          Duration to wait for webserver to get healthy. (default 1m0s)
  -w, --workspace-id string    ID of the Workspace to retrieve environment connections from.

Docker Mode Flags:
      --build-secrets strings   Mimics docker build --secret flag.
      --compose-file string     Location of a custom compose file to use for starting Airflow
  -i, --image-name string       Name of a custom built image to start airflow with
  -n, --no-browser              Don't bring up the browser once the Webserver is healthy
      --no-cache                Do not use cache when building container image

Standalone Mode Flags:
  -f, --foreground    Run in the foreground
  -p, --port string   Port for the Airflow webserver

Global Flags:
      --docker             Run in Docker mode
      --standalone         Run in standalone mode (without Docker)
      --verbosity string   Log level (debug, info, warn, error, fatal, panic (default "warning")
```

### Known limitation

`astro dev run --standalone` does not work because the `run` command uses `DisableFlagParsing` (pre-existing cobra behavior to pass args through to airflow). Use `astro config set dev.mode standalone` instead, which works correctly for all commands including `run`.

## Test plan

- [x] Unit tests for all new standalone commands and helpers
- [x] Unit test for port conflict detection (`TestStandaloneStart_PortInUse`)
- [x] Existing Docker-mode tests still pass
- [x] E2E: fresh project init → start → run/parse/pytest/import/export/logs → restart → stop
- [x] E2E: port conflict between two standalone projects blocked immediately
- [x] E2E: port conflict with non-Airflow process blocked immediately
- [x] E2E: custom port via `astro config set api-server.port` works correctly
- [x] E2E: `build`, `upgrade-test`, `compose-export` show proper errors in standalone mode


airflow/standalone.go
```diff
@@ -0,0 +1,701 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	standaloneDir          = ".astro/standalone"
+	standalonePIDFile      = "airflow.pid"
+	standaloneLogFile      = "airflow.log"
+	defaultStandalonePort  = "8080"
+	standaloneIndexURL     = "https://pip.astronomer.io/v2/"
+	standalonePythonVer    = "3.12"
```
jedcunningham: You'd want to capture the python version from the image they have in their dockerfile as well.

airflow/standalone.go
```diff
@@ -0,0 +1,701 @@
...
+	if err != nil {
+		sp.Stop()
+		return fmt.Errorf("error creating virtual environment: %w", err)
+	}
+
+	// 6. Install dependencies
+	requirementsPath := filepath.Join(s.airflowHome, "requirements.txt")
+	installArgs := []string{
+		"pip", "install",
+		fmt.Sprintf("apache-airflow==%s", airflowVersion),
```
jedcunningham, line 234: This alone isn't enough - you'd need to capture the full set of "stuff" we install by default. And that varies a bit.

Airflow 2 has the concept of a slim and "full" image. The latter has a number of extra providers installed by default. Both have a set of providers. From memory, it also varies a bit based on the Runtime version as well, so you likely can't have a static list across the board here.

Airflow 3 is only slim, but there is a bit of risk here that changes in the default set would cause environment drift. Ideally we aren't hard coding those there to get stale.

airflow/standalone.go
jeremybeard, line 1: nit: rename this and the test file to `standalone[_test].go`

airflow/standalone.go
```diff
@@ -0,0 +1,1089 @@
...
+	cmds, err := standaloneParseFile(filepath.Join(s.airflowHome, "Dockerfile"))
+	if err != nil {
+		return fmt.Errorf("error parsing Dockerfile: %w", err)
+	}
+	_, tag := standaloneGetImageTag(cmds)
+	if tag == "" {
+		return errors.New("could not determine runtime version from Dockerfile")
+	}
+
+	baseTag, tagPython := parseRuntimeTagPython(tag)
```
jeremybeard, line 171: This needs to capture floating tags, e.g. "3.1", which currently are blocked in the validation below

jeremybeard: We may need a general "unknown Airflow version" error, or some way to override it, since you could have a project Dockerfile with your own image/tag that you built from a runtime image

airflow/local.go
```diff
@@ -0,0 +1,1089 @@
...
+	if err := s.applySettings(settingsFile, envConns); err != nil {
+		fmt.Printf("Warning: could not apply airflow settings: %s\n", err.Error())
+	}
+
+	bullet := ansi.Cyan("\u27A4") + " "
+	uiURL := "http://localhost:" + s.webserverPort()
+	fmt.Printf("\n%s Airflow is ready! (PID %d)\n", ansi.Green("\u2714"), cmd.Process.Pid)
+	fmt.Printf("%sAirflow UI: %s\n", bullet, ansi.Bold(uiURL))
+	if user, pass := s.readCredentials(); user != "" {
+		fmt.Printf("%sUsername:   %s\n", bullet, ansi.Bold(user))
```
jeremybeard: Should we use the "all admins" option that is used for Docker mode, so you don't need to log in at all?

airflow/standalone.go
```diff
@@ -0,0 +1,1089 @@
...
+	freezePath, airflowVersion, taskSDKVersion, err := s.getConstraints(baseTag, pythonVersion)
+	if err != nil {
+		return err
+	}
+
+	sp := spinner.NewSpinner("Setting up standalone environment…")
+	sp.Start()
+
+	// 5. Create venv
+	err = runCommand(s.airflowHome, "uv", "venv", "--python", pythonVersion)
```
jeremybeard, line 224: Can we hide all the uv logs by default, but show them in verbose mode, like what we do with image build logs? It's not something you'd typically read

airflow/standalone.go
jeremybeard: A complication with the floating tag is how do you know which patch version to get the constraints for?

airflow/standalone.go
jlaneve: done in 5dc0df77

airflow/standalone.go
jlaneve: if resolution fails and the tag isn't a recognized format, it now gives a specific error: "could not determine runtime version from Dockerfile image tag '...'. standalone mode requires a pinned astronomer runtime image (e.g., astro-runtime:3.1-12)". done in 342affd7

airflow/local.go
jlaneve: switched to `SIMPLE_AUTH_MANAGER_ALL_ADMINS=True` to match docker mode, removed the credentials file and login prompt. done in 534429bb

airflow/standalone.go
jlaneve: done, output is now suppressed unless `--verbose` is passed. same pattern as docker image build logs. done in 6c8bc46c

airflow/standalone.go
jlaneve: the floating tag gets resolved to the latest pinned version via the runtime versions json before fetching constraints, so e.g. "3.1" resolves to "3.1-12" and then constraints are fetched for "3.1-12". done in 342affd7

airflow/container.go
```diff
@@ -23,8 +23,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+// StartOptions is re-exported from airflow/types for use by callers.
+type StartOptions = types.StartOptions
+
 type ContainerHandler interface {
 	Start(imageName, settingsFile, composeFile, buildSecretString string, noCache, noBrowser bool, waitTime time.Duration, envConns map[string]astrocore.EnvironmentObjectConnection) error
+	SetStartOpts(opts types.StartOptions)
```
jeremybeard: Is this to avoid adding even more params to `Start`? Should we just have `types.StartOptions` as single param to `Start` instead?

Issue #601, opened 2022-06-13
astro dev parse Fails when parsing the SnowflakeOperator in a dag


ReadytoRocc:
**Describe the bug**
astro dev parse Fails when parsing the SnowflakeOperator in a dag. Error returned below:

```
8eecb7-test-1  | ______________ test_file_imports[dags/snowflake-dag.py] ______________
8eecb7-test-1  | 
8eecb7-test-1  | rel_path = 'dags/snowflake-dag.py'
8eecb7-test-1  | rv = 'Traceback (most recent call last):\n  File "/usr/local/lib/python3.9/logging/__init__.py", line 1146, in __init__\n  ...r: [Errno 2] No such file or directory: \'/usr/local/airflow/NON_DEFAULT_OS_ENV_VALUE/snowflake_ssm_rt_telemetry.log\''
8eecb7-test-1  | 
8eecb7-test-1  |     @pytest.mark.parametrize("rel_path,rv", get_import_errors(), ids=[x[0] for x in get_import_errors()])
8eecb7-test-1  |     def test_file_imports(rel_path,rv):
8eecb7-test-1  |        """ Test for import errors on a file """
8eecb7-test-1  |        if rel_path and rv : #Make sure our no op test doesn't raise an error
8eecb7-test-1  | >              raise Exception(f"{rel_path} failed to import with message \n {rv}")
8eecb7-test-1  | E     Exception: dags/snowflake-dag.py failed to import with message 
8eecb7-test-1  | E      Traceback (most recent call last):
8eecb7-test-1  | E       File "/usr/local/lib/python3.9/logging/__init__.py", line 1146, in __init__
8eecb7-test-1  | E         StreamHandler.__init__(self, self._open())
8eecb7-test-1  | E       File "/usr/local/lib/python3.9/logging/__init__.py", line 1175, in _open
8eecb7-test-1  | E         return open(self.baseFilename, self.mode, encoding=self.encoding,
8eecb7-test-1  | E     FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/airflow/NON_DEFAULT_OS_ENV_VALUE/snowflake_ssm_rt_telemetry.log'
```

**What CLI Version did you experience this bug?**

Astro CLI Version: 1.0.2

**What Operating System is the above CLI installed on?**
macOS

**🪜 Steps To Reproduce**

Run `astro dev parse` with the following DAG in your dags directory:
```
from airflow import DAG
[...]

PR #2028, opened 2026-03-05
fix(parse): handle PYTEST_THEME env var in DAG integrity test

josh-fell:
Fixes a bug where `astro dev parse` fails when projects specify `--color=yes` in their pytest configuration. The monkeypatched `os.getenv()` in the DAG integrity test templates returns `MOCKED_PYTEST_THEME_VALUE` for unset environment variables, which pytest validates against known Pygment styles and rejects. This change adds special handling for `PYTEST_THEME` (similar to the existing `JENKINS_HOME` fix from #601), returning `None` when the variable is unset instead of a mocked value.

## 🎟 Issue(s)

Resolves a bug where pytest fails with:
```
ERROR: PYTEST_THEME environment variable had an invalid value: 'MOCKED_PYTEST_THEME_VALUE'.
Only valid pygment styles are allowed.
```

Related to #601 (similar fix pattern for JENKINS_HOME)

## 🧪 Functional Testing

1. Create an Astro project with pytest configured to use `--color=yes`:
   ```bash
   astro dev init
   ```

2. Add to `pytest.ini` or `pyproject.toml`:
   ```ini
   [tool.pytest.ini_options]
   addopts = "--color=yes"
   ```

3. Run parse command:
   ```bash
   astro dev parse
   ```

4. Verify it completes successfully without PYTEST_THEME error

**Before this fix:** Command fails with "PYTEST_THEME environment variable had an invalid value"
**After this fix:** Command succeeds and DAGs parse correctly

## 📸 Screenshots

N/A - This is a bug fix that prevents an error message, no visual changes

## 📋 Checklist

- [ ] Added/updated applicable tests (N/A - no existing tests for monkeypatch functions)
- [ ] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary). (N/A - local CLI fix only)
- [ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary). (N/A - local CLI fix only)
- [ ] Communicated to/tagged owners of respective clients potentially impacted by these changes. (N/A - internal fix)
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/) (N/A - fixes existing behavior)

## Files Changed

- `airflow/include/airflow2/dagintegritytestdefault.py` - Added PYTEST_THEME special case
- `airflow/include/airflow3/dagintegritytestdefault.py` - Added PYTEST_THEME special case
- `go.mod` - Fixed go version format from `1.24.0` to `1.24`

## Notes for Reviewers

This follows the exact same pattern as the JENKINS_HOME fix that was added for issue #601. The monkeypatch templates already have infrastructure for handling environment variables that need special treatment - this simply adds PYTEST_THEME to that list.

### DIFF
--- a/airflow/include/airflow2/dagintegritytestdefault.py
+++ b/airflow/include/airflow2/dagintegritytestdefault.py
@@ -46,6 +46,10 @@ def os_getenv_monkeypatch(key: str, *args, **kwargs):
         key == "JENKINS_HOME" and default is None
     ):  # fix https://github.com/astronomer/astro-cli/issues/601
         return None
+    if (
+        key == "PYTEST_THEME" and default is None
+    ):  # fix pytest validation of pygment styles
+        return None
     if default:
         return default  # otherwise return whatever default has been passed
     return f"MOCKED_{key.upper()}_VALUE"  # if absolutely nothing has been passed - return the mocked value
--- a/airflow/include/airflow3/dagintegritytestdefault.py
+++ b/airflow/include/airflow3/dagintegritytestdefault.py
@@ -46,6 +46,10 @@ def os_getenv_monkeypatch(key: str, *args, **kwargs):
         key == "JENKINS_HOME" and default is None
     ):  # fix https://github.com/astronomer/astro-cli/issues/601
         return None
+    if (
+        key == "PYTEST_THEME" and default is None
+    ):  # fix pytest validation of pygment styles
+        return None
     if default:
         return default  # otherwise return whatever default has been passed
     return f"MOCKED_{key.upper()}_VALUE"  # if absolutely nothing has been passed - return the mocked value


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/astronomer/astro-cli
 
-go 1.24.0
+go 1.24
```
josh-fell: When running `make test` locally I was running into this issue. Let me know if this change isn't necessary though; just trying to accomplish the PR checklist. Long-time listener, first-time caller here.

```
❯ make test
go test -count=1 -cover -coverprofile=coverage.txt -covermode=atomic ./... -test.v
go: errors parsing go.mod:
/Users/joshfell/conductor/workspaces/astro-cli/moscow-v1/go.mod:3: invalid go version '1.24.0': must match format 1.23
make: *** [test] Error 1
```

jlaneve: we should undo this, it's breaking CI

PR #2032, opened 2026-03-05
fix: replace kin-openapi with libopenapi for OpenAPI 3.1 support

jeremybeard:
## Summary

- Replaces `kin-openapi` with `pb33f/libopenapi` v0.34.0 to fix OpenAPI 3.1 spec parsing for Airflow 3.1+
- Airflow 3.1+ uses numeric `exclusiveMinimum` (OpenAPI 3.1), which kin-openapi v0.131.0 silently fails to parse, resulting in "no endpoints found"
- Introduces self-contained types in `pkg/openapi/types.go` so the library dependency is isolated to the parsing layer, making tests simpler and future library swaps trivial

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./pkg/openapi/... ./cmd/api/...` all tests pass
- [x] `astro api airflow ls --airflow-version 3.1.7` lists endpoints successfully
- [x] `astro api airflow ls --airflow-version 3.0.3` still works for 3.0.x
- [x] `astro api airflow ls --airflow-version 2.10.0` still works for 2.x
- [x] `astro api airflow describe get_dags --airflow-version 3.1.7` schema details display correctly
- [x] `astro api airflow ls` works with live local Airflow instance (auto-detect version)


go.mod
```diff
@@ -130,6 +132,7 @@ require (
 	github.com/fsnotify/fsevents v0.2.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
 	github.com/fzipp/gocyclo v0.6.0 // indirect
+	github.com/getkin/kin-openapi v0.131.0 // indirect
```
schnie, line 135: Why does kin still get pulled in?

jeremybeard: "indirect" means it's a transitive dependency of something else. It was already there before I added the feature, this is why I used it in the first place, turns out it wasn't the best choice!

PR #2039, opened 2026-03-18
[AFE-53] feat(airflow-client): paginate Get* requests and add short retries

nickypy:
## Description

This PR fixes an issue around copying pools, variables, and connections between deployments:
- paginate across all resources to make sure everything is copied, not just the first page of the results
- add retries for scenarios when deployments are still spinning up (e.g. deploy previews)

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

1. Create a test deployment with >100 pools, variables and connections
2. Use that to spin up a preview deployment
3. Make sure that **all** pools, variables and connections are copied over

## 📸 Screenshots

## 📋 Checklist

### DIFF
--- a/airflow-client/airflow-client.go
+++ b/airflow-client/airflow-client.go
@@ -17,12 +17,10 @@ import (
 var errDecode = errors.New("failed to decode response from API")
 
 const (
-	pageLimit  = 100
-	maxRetries = 10
+	pageLimit    = 100
+	maxRetries   = 10
+	retryBackoff = time.Second
 )
-
-// retryBackoff is a var so tests can override it to avoid real sleeps.
-var retryBackoff = time.Second
 
 type Client interface {
 	// connections
--- a/airflow-client/airflow-client_test.go
+++ b/airflow-client/airflow-client_test.go
@@ -8,7 +8,7 @@ import (
 	"io"
 	"net/http"
 	"testing"
-	"time"
+	"testing/synctest"
 
 	"github.com/astronomer/astro-cli/pkg/httputil"
 	testUtil "github.com/astronomer/astro-cli/pkg/testing"
@@ -28,13 +28,6 @@ func (f errorRoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error)
 
 func TestAirflowClient(t *testing.T) {
 	suite.Run(t, new(Suite))
-}
-
-func (s *Suite) SetupSuite() {
-	// Use minimal backoff so retry tests don't sleep.
-	original := retryBackoff
-	retryBackoff = time.Millisecond
-	s.T().Cleanup(func() { retryBackoff = original })
 }
 
 func (s *Suite) TestNew() {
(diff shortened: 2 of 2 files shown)


airflow-client/airflow-client.go
```diff
@@ -39,18 +51,36 @@ func NewAirflowClient(c *httputil.HTTPClient) *HTTPClient {
 	}
 }
 
-func (c *HTTPClient) GetConnections(airflowURL string) (Response, error) {
-	doOpts := &httputil.DoOptions{
-		Path:   "https://" + airflowURL + "/connections",
-		Method: http.MethodGet,
+// fetchAllPages fetches paginated results from an Airflow REST endpoint, accumulating
+// all items across pages. When TotalEntries is absent (0), it stops on a short page.
```
nickstenning:
```suggestion
// all items across pages. It stops on the first page with fewer than pageLimit entries, or when
// the offset exceeds the response's TotalEntries value (if present).
```

airflow-client/airflow-client.go
```diff
@@ -219,22 +252,37 @@ func (c *HTTPClient) DoAirflowClient(doOpts *httputil.DoOptions) (*Response, err
 		}
 	}
 
-	response, err := c.Do(doOpts)
-	if err != nil {
-		return nil, err
-	}
-	defer response.Body.Close()
+	var lastErr error
+	for attempt := 0; attempt <= maxRetries; attempt++ {
```
nickstenning: Hmmm, I wonder if we're doing retries whether we should be using something like [`github.com/hashicorp/go-retryablehttp`](https://github.com/hashicorp/go-retryablehttp) here. Not a hard requirement, but I do always get a bit suspicious when HTTP clients implement retries themselves.. 😆

airflow-client/airflow-client.go
```diff
@@ -1,17 +1,29 @@
...
 
 	"github.com/astronomer/astro-cli/context"
 	"github.com/astronomer/astro-cli/pkg/httputil"
+	"github.com/sirupsen/logrus"
 )
 
 var errDecode = errors.New("failed to decode response from API")
 
+// sleepFn is a variable so tests can replace it with a no-op to avoid real sleeps.
+var sleepFn = time.Sleep
```
nickstenning: The "modern Go" way to do this would probably be to use [`testing/synctest`](https://pkg.go.dev/testing/synctest). Maybe this is a moot point if we're going to adopt go-retryablehttp as mentioned in a later comment, but just flagging it for the future.

airflow-client/airflow-client.go
```diff
@@ -1,17 +1,29 @@
 package airflowclient
 
 import (
+	stdctx "context"
```
nickstenning, line 4: Not your issue, but naming a local package context is a bit of a yikes from me. I wonder if it's too late to fix that... 🤔 (not in this PR).

airflow-client/types.go
```diff
@@ -29,7 +29,8 @@ type Pool struct {
 }
 
 type Response struct {
-	Connections []Connection `json:"connections"`
-	Variables   []Variable   `json:"variables"`
-	Pools       []Pool       `json:"pools"`
+	Connections  []Connection `json:"connections"`
+	Variables    []Variable   `json:"variables"`
+	Pools        []Pool       `json:"pools"`
+	TotalEntries int          `json:"total_entries"`
```
nickstenning, line 35: Is this already returned by the server and we just weren't capturing it?

nickypy: Correct, `total_entries` was added alongside Airflow 2

airflow-client/airflow-client.go
nickypy: Done. I agree, and I'd also rather not re-invent retries here

airflow-client/airflow-client_test.go
```diff
@@ -17,10 +19,24 @@ type Suite struct {
...
+func (f errorRoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return nil, f(req)
+}
+
 func TestAirflowClient(t *testing.T) {
 	suite.Run(t, new(Suite))
 }
 
+func (s *Suite) SetupSuite() {
+	// Use minimal backoff so retry tests don't sleep.
```
nickstenning: Is there a reason we're not using `synctest`?

nickypy: Once [this](https://github.com/astronomer/astro-cli/pull/2051) is merged in I should be able to migrate the tests to use `synctest`

nickypy: Done: https://github.com/astronomer/astro-cli/pull/2039/commits/fcf72faf02f0b202478480bf4d5cfaa36959a90b

PR #2042, opened 2026-03-20
Add `astro api registry` command

kaxil:
## Summary

Adds `astro api registry` for querying the public [Airflow Provider Registry](https://airflow.apache.org/registry), following the same patterns and infrastructure as `astro api airflow` and `astro api cloud`.

Supports both **path-based** and **operation ID-based** queries, plus `ls` and `describe` for endpoint discovery — identical UX to the other `astro api` subcommands.

## Design Rationale

**Consistent with airflow/cloud:** Reuses `RequestOptions`, `executeRequest`, `resolveOperationID`, `applyPathParams`, `findMissingPathParams`, `generateCurl`, the shared output pipeline, and `openapi.Cache`. The only registry-specific code is URL resolution and the connection error message.

**No convenience subcommands:** The operation ID approach (`getProviderModulesLatest -p providerId=amazon`) is consistent with how airflow/cloud work and covers all 9 endpoints without hardcoding any of them. Convenience commands can be added later under `astro registry` if there's demand.

**Only addition to shared code:** `SetStripPrefix` on `openapi.Cache` — the registry's OpenAPI spec uses `/api` path prefixes that need stripping for `ls`/`describe` display.

## Usage Examples

```bash
# Discover endpoints
astro api registry ls
astro api registry describe getProviderModulesLatest

# Query by operation ID
astro api registry listProviders
astro api registry getProviderModulesLatest -p providerId=amazon
astro api registry getProviderModulesByVersion -p providerId=amazon -p version=9.22.0
astro api registry getProviderVersions -p providerId=google

# Query by path
astro api registry /providers.json
astro api registry /providers/amazon/modules.json
astro api registry /providers/amazon/9.22.0/modules.json
astro api registry /modules.json

# Same flags as airflow/cloud
astro api registry listProviders --jq '.providers[0].id'
astro api registry /providers.json --template '{{range .providers}}{{.id}}{{"\n"}}{{end}}'
astro api registry /providers.json --generate
astro api registry /providers.json --verbose
astro api registry /providers.json -i
astro api registry /providers.json --silent
astro api registry /providers.json -H "X-Custom:value"

# Override registry URL
astro api registry --registry-url https://custom.example.com listProviders
ASTRO_REGISTRY_URL=https://custom.example.com astro api registry listProviders
```

## Gotchas / Tradeoffs

- The `/api` prefix is added automatically for path-based access, so users pass `/providers.json` not `/api/providers.json`
- `ls` and `describe` strip the `/api` prefix from paths for readability (via `SetStripPrefix`)
- No auth required — the registry is a public API, so the `token` parameter to `executeRequest` is always empty
- `--registry-url` precedence: flag > `ASTRO_REGISTRY_URL` env > default
- Pagination flags (`--paginate`, `--slurp`) are intentionally omitted since the registry API doesn't paginate

### DIFF
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -28,8 +28,9 @@ func NewAPICmdWithOutput(out io.Writer) *cobra.Command {
 The 'astro api' command provides direct access to Astronomer's REST APIs.
 
 Available subcommands:
-  airflow  Make requests to the Airflow REST API
-  cloud    Make requests to the Astro Cloud API (api.astronomer.io)
+  airflow   Make requests to the Airflow REST API
+  cloud     Make requests to the Astro Cloud API (api.astronomer.io)
+  registry  Query the Airflow Provider Registry
 
 Use "astro api [command] --help" for more information about a command.`,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
@@ -69,6 +70,7 @@ Use "astro api [command] --help" for more information about a command.`,
 
 	cmd.AddCommand(NewAirflowCmd(out))
 	cmd.AddCommand(NewCloudCmd(out))
+	cmd.AddCommand(NewRegistryCmd(out))
 
 	return cmd
 }
--- a/cmd/api/registry.go
+++ b/cmd/api/registry.go
@@ -1,0 +1,332 @@
+package api
+
+import (
+	"fmt"
+	"io"
+	"os"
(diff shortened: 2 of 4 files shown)


kaxil:
`api` sounds like a wrong subcommand given the current ones directly use the OpenAPI spec but I didn't want to put `registry` at top level. Suggestions welcomed

PR #2053, opened 2026-03-24
fix(airflow_versions): correct astro-agent default image ordering

feluelle:
## Summary

Default **astro-agent** image selection (from the client-versions manifest) now ranks candidates with **Astro runtime** rules first, then **Python** and **astro-agent** versions using `golang.org/x/mod/semver` (`cmp.Or` + `semver.Compare` on `v`-prefixed fields). Runtime in the tag still uses `CompareRuntimeVersions`; only the Python/agent tie-breaks move from string compare to semver.

This fixes incorrect ordering where string comparison disagrees with semver (e.g. `3.10` vs `3.9`) and ensures a **newer agent** wins when image runtime and Python match across different client versions (e.g. `1.3.7` over `1.3.6` for the same `3.1-13-python-3.12-...` shape).

The redundant **sort of stable client-version keys** before scanning tags was removed; iteration order does not change the outcome once `isBetterImage` compares runtime, Python, and agent for every candidate.

## Test plan

- [x] `go test ./airflow_versions/...`
- [ ] CI / pre-commit

### DIFF
--- a/airflow_versions/airflow_versions.go
+++ b/airflow_versions/airflow_versions.go
@@ -1,12 +1,14 @@
 package airflowversions
 
 import (
+	"cmp"
 	"fmt"
 	"regexp"
 	"sort"
 	"strings"
 
 	"github.com/astronomer/astro-cli/pkg/logger"
+	"golang.org/x/mod/semver"
 )
 
 const (
@@ -63,8 +65,8 @@ func ParseImageTag(imageTag string) (*ImageTagInfo, error) {
 	}, nil
 }
 
-// isBetterImage compares two ImageTagInfo objects to determine if the candidate has higher runtime version or python version than the current
-// Priority: Runtime version first, then Python version
+// isBetterImage reports whether candidate should replace current for default astro-agent image selection.
+// Priority: Astro runtime (CompareRuntimeVersions), then Python semver, then astro-agent semver.
 func isBetterImage(candidate, current *ImageTagInfo) bool {
 	if candidate == nil {
 		return false
@@ -72,18 +74,11 @@ func isBetterImage(candidate, current *ImageTagInfo) bool {
 	if current == nil {
 		return true
 	}
-
-	// Compare runtime versions first using existing function
-	runtimeComparison := CompareRuntimeVersions(candidate.RuntimeVersion, current.RuntimeVersion)
-	if runtimeComparison > 0 {
-		return true
-	}
-	if runtimeComparison < 0 {
-		return false
-	}
-
-	// Same runtime version, compare Python versions (string comparison works for semantic versions like "3.11" vs "3.12")
-	return candidate.PythonVersion > current.PythonVersion
+	return cmp.Or(
+		CompareRuntimeVersions(candidate.RuntimeVersion, current.RuntimeVersion),
+		semver.Compare("v"+candidate.PythonVersion, "v"+current.PythonVersion),
+		semver.Compare("v"+candidate.AgentVersion, "v"+current.AgentVersion),
+	) > 0
 }
 
 // GetDefaultImageTag returns default airflow image tag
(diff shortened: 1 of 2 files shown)


airflow_versions/airflow_versions.go
```diff
@@ -63,8 +64,46 @@ func ParseImageTag(imageTag string) (*ImageTagInfo, error) {
 	}, nil
 }
 
-// isBetterImage compares two ImageTagInfo objects to determine if the candidate has higher runtime version or python version than the current
-// Priority: Runtime version first, then Python version
+// dottedVersionToSemver normalizes Python/agent strings from image tags (e.g. "3.12", "1.3.7")
+// for golang.org/x/mod/semver. Two-part X.Y is treated as X.Y.0.
+func dottedVersionToSemver(s string) string {
```
nickstenning: Is this needed? The package you're using already handles missing minor/patch/etc.

airflow_versions/airflow_versions.go
```diff
@@ -63,8 +64,46 @@ func ParseImageTag(imageTag string) (*ImageTagInfo, error) {
...
+}
+
+// compareSemverLike compares dotted version strings from image tags using semver rules.
+func compareSemverLike(a, b string) int {
+	va := dottedVersionToSemver(a)
+	vb := dottedVersionToSemver(b)
+	okA := va != ""
+	okB := vb != ""
+	if !okA && !okB {
+		return strings.Compare(a, b)
```
nickstenning: Should we even fall back to this? It seems to me that `semver.Compare` already basically does what we want, with the one exception of making sure there's a `v` on the front of the strings first.

airflow_versions/airflow_versions.go
```diff
@@ -82,8 +121,14 @@ func isBetterImage(candidate, current *ImageTagInfo) bool {
 		return false
 	}
 
-	// Same runtime version, compare Python versions (string comparison works for semantic versions like "3.11" vs "3.12")
-	return candidate.PythonVersion > current.PythonVersion
```
nickstenning, line 86: 😬

airflow_versions/airflow_versions.go
```diff
@@ -63,8 +64,46 @@ func ParseImageTag(imageTag string) (*ImageTagInfo, error) {
...
+func dottedVersionToSemver(s string) string {
+	s = strings.TrimSpace(strings.TrimPrefix(s, "v"))
+	if s == "" {
+		return ""
+	}
+	candidate := "v" + s
+	if semver.IsValid(candidate) {
+		return candidate
+	}
+	if strings.Count(s, ".") == 1 && !strings.ContainsAny(s, "-+") {
```
neel-astro: curious q: Is the `-` or `+` to help identify the nightlies or build images? I am guessing in that case we are preferring non-nightly image  tag always, might want to comment that behaviour or include that in the test

airflow_versions/airflow_versions.go
```diff
@@ -176,7 +221,7 @@ func getAstroAgentTag(clientVersions map[string]ClientVersion, runtimeVersion st
 		return "", fmt.Errorf("no client versions found")
 	}
 
-	// sort the available versions by runtime version in descending order
+	// Deterministic iteration order (JSON map key order is not defined). Selection does not depend on this sort once isBetterImage compares agent versions.
 	sort.Slice(availableVersions, func(i, j int) bool {
```
nickstenning: Why are we doing this?

airflow_versions/airflow_versions.go
feluelle: This is actually not much of an issue for Python versions for quite some time :D

airflow_versions/airflow_versions.go
```diff
@@ -176,7 +221,7 @@ func getAstroAgentTag(clientVersions map[string]ClientVersion, runtimeVersion st
...
 		return CompareRuntimeVersions(availableVersions[i], availableVersions[j]) < 0
 	})
```
neel-astro: I think we can get away from needing this `sort` given that we are anyway iterating all the versions, and the best image tag considers all three of runtime, python, and agent version

airflow_versions/airflow_versions.go
```diff
@@ -63,8 +64,46 @@ func ParseImageTag(imageTag string) (*ImageTagInfo, error) {
...
+	}
+	candidate := "v" + s
+	if semver.IsValid(candidate) {
+		return candidate
+	}
+	if strings.Count(s, ".") == 1 && !strings.ContainsAny(s, "-+") {
+		candidate = "v" + s + ".0"
+		if semver.IsValid(candidate) {
+			return candidate
+		}
```
neel-astro:
```suggestion
    // Normalize X.Y → X.Y.0 before single validation
    if strings.Count(s, ".") == 1 && !strings.ContainsAny(s, "-+") {
        s += ".0"
    }
    if v := "v" + s; semver.IsValid(v) {
        return v
    }
```

nit: would this be the same? avoids the double semver.IsValid call

PR #2071, opened 2026-04-07
Move shared request editor and API error logic into pkg/httputil

ashb:
## Description

Consolidates the duplicated TrueString constant, ErrorRequest/ErrorBaseURL/
HTTPStatus200/HTTPStatus204 vars, NormalizeAPIError function, and request
editor function from all three astro-client-*-core packages into a new
pkg/httputil/request_editor.go. Each client package now re-exports the
vars/const and delegates NormalizeAPIError to httputil, and constructs its
request editor via httputil.NewRequestEditorFn with its API path argument.

Also fixes the double x-astro-client-version header bug in
astro-client-core/client.go. The original code called Header.Add (which
appends, not overwrites) unconditionally after the switch that already set
it -- so every request carried two x-astro-client-version values. In the
deploy-action case the server would see both the deploy action version and
CurrVersion.

### DIFF
--- a/airflow/standalone_windows.go
+++ b/airflow/standalone_windows.go
@@ -5,10 +5,11 @@ package airflow
 import (
 	"errors"
 
+	"github.com/pkg/browser"
+
 	"github.com/astronomer/astro-cli/airflow/types"
 	airflowversions "github.com/astronomer/astro-cli/airflow_versions"
 	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
-	"github.com/pkg/browser"
 )
 
 var (
--- a/pkg/httputil/request_editor.go
+++ b/pkg/httputil/request_editor.go
@@ -1,0 +1,68 @@
+package httputil
+
+import (
+	"bytes"
+	httpContext "context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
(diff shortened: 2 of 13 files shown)


pkg/httputil/request_editor.go
```diff
@@ -0,0 +1,68 @@
...
+		requestURL, err := url.Parse(baseURL + req.URL.String())
+		if err != nil {
+			return fmt.Errorf("%w, %s", ErrorBaseURL, baseURL)
+		}
+		req.URL = requestURL
+		req.Header.Add("authorization", token)
+		switch {
+		case os.Getenv("DEPLOY_ACTION") == "true" && os.Getenv("GITHUB_ACTIONS") == "true":
+			req.Header.Add("x-astro-client-identifier", "deploy-action")
+			req.Header.Add("x-astro-client-version", os.Getenv("DEPLOY_ACTION_VERSION"))
```
neel-astro, line 56: good catch on `x-astro-client-version` trailing overwrite in the existing client ❤️

neel-astro (approved):
LGTM

airflow/standalone_windows.go
ashb, line 1: Yes, this is unrelated change, but it was from `golangci-lint fmt` -- I don't know why that wasn't checked in CI

PR #2072, opened 2026-04-07
AFE-85: Pin dependencies (astro-cli)

jaketf:
## Description
* pin actions with pinact (see example PR)

* fix curl | sh in CircleCI and GitHub Actions workflows

* update Dockerfiles with digest pins

* ensure `uv --frozen` is used everywhere in CI builds

## Motivation 
 [AFE-85: Pinning dependencies temporarily](https://linear.app/astronomer/issue/AFE-85)


.circleci/config.yml
```diff
@@ -56,7 +56,15 @@ jobs:
       - setup-cross-build-libs
       - run:
           name: goreleaser
-          command: curl -sL https://raw.githubusercontent.com/goreleaser/get/master/get | bash
+          command: |
+            GORELEASER_VERSION=v2.15.2
```
nickstenning: Let's use `goreleaser/goreleaser-action` instead.

nickstenning (approved):
A suggestion to use an existing action rather than our own script, but otherwise LGTM.

neel-astro (approved):

PR #2075, opened 2026-04-08
Support local spec files for --spec-url

jeremybeard:
## Summary
- Allow `--spec-url` to accept local file paths (absolute, relative, `~/`, `file://`) in addition to HTTP URLs
- Local specs are read fresh on every `Load()` call with no caching, ideal for development workflows
- `--spec-token-env-var` is silently ignored for local files since auth is meaningless for local access

## Test plan
- [x] New unit tests for `IsLocalSpec`, `ResolveLocalPath`, `NewCacheForLocalFile`
- [x] New unit tests for `Load` with local files (success, not found, invalid, always-fresh)
- [x] New unit tests for `initCloudSpecCache` with local paths, `file://` URLs, and token env var
- [x] All existing tests pass
- [x] Manual test: `astro api cloud ls --spec-url file:///path/to/spec.yaml`

### DIFF
--- a/cmd/api/cloud.go
+++ b/cmd/api/cloud.go
@@ -131,7 +131,7 @@ To pass nested values as arrays, declare multiple fields with key[]=value1.`,
 
 	// Other flags
 	cmd.Flags().BoolVar(&opts.GenerateCurl, "generate", false, "Output a curl command instead of executing the request")
-	cmd.PersistentFlags().StringVar(&opts.SpecURL, "spec-url", "", "OpenAPI spec URL (overrides default Cloud API spec)")
+	cmd.PersistentFlags().StringVar(&opts.SpecURL, "spec-url", "", "OpenAPI spec URL or file path (overrides default Cloud API spec)")
 	cmd.PersistentFlags().StringVar(&opts.SpecTokenEnvVar, "spec-token-env-var", "", "Environment variable containing auth token for fetching the spec")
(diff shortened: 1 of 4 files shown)


nickstenning (approved):
Very nice. LGTM.

cmd/api/cloud.go
```diff
@@ -394,9 +394,19 @@ func initCloudSpecCache(opts *CloudOptions, ctx *config.Context) error {
 		return nil
 	}
 
-	// When --spec-url is provided, use it. If --spec-token-env-var is also set,
-	// read the token from that env var and send it as auth when fetching the spec.
+	// When --spec-url is provided, use it. It can be a remote URL or a local file path.
 	if opts.SpecURL != "" {
+		// Local file: read directly from disk, no caching or auth needed.
+		if openapi.IsLocalSpec(opts.SpecURL) {
+			localPath, err := openapi.ResolveLocalPath(opts.SpecURL)
```
neel-astro, line 401: nit: should we log that we are treating the spec URL as a local path, because the `IsLocalSpec` method defaults to true for anything that is not starting with http or https, could even be a user typo in worst case

neel-astro (approved):
LGTM 🚀

cmd/api/cloud.go
jeremybeard: I think the error should be clear enough about what went wrong

PR #2077, opened 2026-04-09
Store auth credentials in the OS keychain (macOS + Linux)

ashb:
Move auth tokens and refresh tokens out of the plaintext config.yaml
into the OS-native credential store via 99designs/keyring:

- macOS: Keychain Services with per-app ACL (re-prompts after binary
  updates when another process tries to read the item)
- Linux: Secret Service (GNOME Keyring / KWallet) when a D-Bus
  session is available
- Windows support to follow

Credentials are managed through a SecureStore interface (pkg/keychain)
with Get/Set/DeleteCredentials methods. A TokenHolder (pkg/httputil) is
populated once during PersistentPreRunE and read by API client request
editors on every outbound request — replacing the previous per-request
context lookup for the token.

On first run, MigrateLegacyCredentials copies token/refreshtoken fields
from existing config.yaml contexts into the secure store and clears
them from the config file.

On Linux, when no Secret Service daemon is reachable (headless CI,
containers without D-Bus), the store falls back to a plaintext JSON
file at ~/.astro/credentials.json (mode 0600). This matches the
previous config.yaml security posture and is acceptable because CI
environments typically use ASTRO_API_TOKEN rather than interactive
login credentials.

Windows uses the same plaintext file fallback for now; the Credential
Manager backend is added in the follow-up commit.

After building and running this, the config file looks like:

Screenshot 2026-04-09 at 13 11 54

And we can see the item in keychain access.

Screenshot 2026-04-09 at 13 09 10

As of this Pr. `astro auth token` still works without further prompting (but I have a separate PR to put that, at least on macOS behind a TouchID et al prompt).

A similar approach is possible on Windows (and the code is written, but not yet tested)

### DIFF
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -14,6 +14,7 @@ import (
 	astroplatformcore "github.com/astronomer/astro-cli/astro-client-platform-core"
 	"github.com/astronomer/astro-cli/cloud/deployment"
 	"github.com/astronomer/astro-cli/config"
+	"github.com/astronomer/astro-cli/pkg/credentials"
 	"github.com/astronomer/astro-cli/pkg/fileutil"
 	"github.com/astronomer/astro-cli/pkg/git"
 	"github.com/astronomer/astro-cli/pkg/logger"
@@ -29,6 +30,7 @@ type DeployBundleInput struct {
 	WaitTime           time.Duration
 	PlatformCoreClient astroplatformcore.CoreClient
 	CoreClient         astrocore.CoreClient
+	Creds              *credentials.CurrentCredentials
 }
 
 func DeployBundle(input *DeployBundleInput) error {
@@ -44,7 +46,7 @@ func DeployBundle(input *DeployBundleInput) error {
 	}
 
 	// if CI/CD is enforced, check the subject can deploy
-	if currentDeployment.IsCicdEnforced && !canCiCdDeploy("Bearer "+os.Getenv("ASTRO_API_TOKEN")) {
+	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(input.Creds) {
 		return fmt.Errorf(errCiCdEnforcementUpdate, currentDeployment.Name)
 	}
 
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -44,7 +44,7 @@ func DeployBundle(input *DeployBundleInput) error {
 	}
 
 	// if CI/CD is enforced, check the subject can deploy
-	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(c.Token) {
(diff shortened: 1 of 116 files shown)


cloud/deploy/bundle.go
```diff
@@ -44,7 +44,7 @@ func DeployBundle(input *DeployBundleInput) error {
 	}
 
 	// if CI/CD is enforced, check the subject can deploy
-	if currentDeployment.IsCicdEnforced && !canCiCdDeploy(c.Token) {
+	if currentDeployment.IsCicdEnforced && !canCiCdDeploy("Bearer "+os.Getenv("ASTRO_API_TOKEN")) {
```
neel-astro: curious q: why do we directly refer api token env var here?

Edit: Ahh, because CI/CD deploy mode can only use api token and not session tokens, can we simplify this fetching env var logic a bit into some method, because it is being used in bunch of places, and I think we should error out early with appropriate error message if env var is not set, rather than passing empty value to `canCiCdDeploy` and failing on parsing token

pkg/keychain/keychain_linux.go
```diff
@@ -0,0 +1,39 @@
...
+// If no Secret Service daemon is available (e.g. headless CI environments),
+// falls back to a plaintext JSON file at ~/.astro/credentials.json with
+// 0600 permissions. This matches the current plaintext config.yaml behavior
+// and is intentional — encrypted file fallback is not worth the complexity
+// given that CI environments use ASTRO_API_TOKEN anyway.
+//
+// NOTE: if 99designs/keyring fails to connect to Secret Service in environments
+// that DO have it running, replace with godbus/dbus directly:
+// https://github.com/godbus/dbus — the SecureStore interface is the only
+// change boundary.
```
neel-astro: Is this a comment for future devs or just for you while you are running through tests in various envs

software/auth/auth.go
```diff
@@ -257,20 +261,13 @@ func Login(domain string, oAuthOnly bool, username, password, houstonVersion str
...
-	c, err := context.GetContext(domain)
-	if err != nil {
-		return
-	}
-
-	err = c.SetContextKey("token", "")
-	if err != nil {
-		return
+func Logout(domain string, store keychain.SecureStore) {
+	if err := store.DeleteCredentials(domain); err != nil {
```
neel-astro: nit: probably need the `store == nil` check here as well, similar to what we have on the cloud side?

neel-astro (commented):
Mostly looks great apart from tiny comments, this would be a great addition to CLI posture, so thanks for all the effort 🙇 
Also, it would be worth checking basic functionality with APC platform or ask APC team to check the same

pkg/keychain/keychain_linux.go
ashb: Yeah not really relevant anymore, was a note to myself. Will remove.

neel-astro (approved):

ashb:
I'm testing/asking for testers with this on APC before we merge.

PR #2078, opened 2026-04-09
Batch import Airflow settings to speed up dev start

jlaneve:
## Summary

`astro dev start` and `astro dev object import` were very slow because they loaded each variable, connection, and pool from `airflow_settings.yaml` by shelling into the Airflow container and running a separate `airflow variables set` / `airflow connections add` / `airflow pools set` command per item. Every command paid the full Airflow CLI startup cost (~2-5 seconds), so a project with 10 connections, 5 variables, and 2 pools would spend **1-2+ minutes** just on settings injection.

This PR replaces the per-item CLI approach with direct calls to the Airflow REST API. For each object we POST to create, and if the object already exists (409 Conflict) we PATCH to update — matching the previous overwrite behavior without needing to list + delete + recreate.

### Settings import (`settings/settings.go`)

- `ConfigSettings`, `AddVariables`, `AddConnections`, `AddPools` now take an Airflow API URL + optional auth header instead of a container ID
- Each function makes HTTP calls against `/api/v1/{resource}` (Airflow 2) or `/api/v2/{resource}` (Airflow 3)
- URI-only connections (e.g. `conn_uri: postgres://user:pass@host:5432/db`) are parsed into individual fields since the REST API doesn't accept a `uri` field
- Connection extras are serialized to a JSON string as the API expects
- `default_pool` uses `?update_mask=slots&update_mask=include_deferred` on PATCH since Airflow only allows those two fields to be updated on the default pool
- Airflow 1 codepaths are removed entirely since Astronomer no longer ships Airflow 1
- Tests use `httptest.NewServer` to mock the Airflow API

### Caller updates (`airflow/docker.go`, `airflow/standalone.go`)

- **Docker mode**: constructs `http://localhost:<port>/api/v<version>` and fetches auth — Basic Auth (`admin:admin`) for Airflow 2, JWT token via `/auth/token` for Airflow 3's SimpleAuthManager
- **Port detection**: when proxy mode allocates a random port (not the config default), the code now queries the running container's port via `composeService.Ps()` so `object import` finds the right port. `printStatus` / `printProxyStatus` get the port from the overrides they already receive
- **Standalone mode**: persists the allocated port to `.astro/standalone/port` on start and reads it back in `object import`. Also fetches a JWT token the same way Docker mode does

## Performance impact

Previously: O(n) Airflow CLI invocations per object type, each ~2-5s overhead.
Now: O(n) HTTP requests with ~10-50ms overhead each. A typical settings file goes from >1 minute to a few seconds.

## Test plan

Unit tests:
- [x] `go test ./settings/... ./airflow/...` passes (36 settings tests including new API mocking tests, airflow tests updated)
- [x] `golangci-lint run ./settings/... ./airflow/...` clean

E2E tests (done manually with real Airflow containers/venvs):
- [x] **Docker + Airflow 3** (runtime 3.1-12): `dev start`, `dev object import`, `dev restart`, verified every value round-trips through the REST API
- [x] **Docker + Airflow 2** (runtime 13.6.0 / Airflow 2.11.2): `dev start`, `dev object import`, `dev restart`, verified via `/api/v1` with Basic Auth
- [x] **Standalone** (runtime 3.1-12): `dev start --standalone`, `dev object import --standalone`, `dev restart --standalone`, verified JWT auth + port persistence
- [x] **Port overrides**: verified that proxy-allocated random ports (e.g. 13425, 18107) are detected and used correctly
- [x] **Edge cases**: special characters in passwords (`p@ss'w0rd&special=chars!@#$%`), JSON values in variables, Unicode, URI-only connections parsed correctly, invalid items skipped with existing warning messages, empty settings file, upsert on restart
- [x] **conn_extra types**: map[any]any (YAML nested keys), JSON strings, nil

## Out of scope

There's a pre-existing crash in `settings.ExportConnections` when running against Airflow 3 — it parses the YAML output of `airflow connections list -o yaml` which has a different format in Airflow 3. Not related to this PR; filed separately.

### DIFF
--- a/settings/settings.go
+++ b/settings/settings.go
@@ -159,7 +159,8 @@ func airflowAPIRequest(method, requestURL, authHeader string, body []byte) (resp
 }
 
 // airflowAPIUpsert creates a resource via POST, and if it already exists (409), updates via PATCH.
-func airflowAPIUpsert(baseURL, resource, id, authHeader string, body []byte) error {
+// updateSuffix is appended to the PATCH URL (e.g., "?update_mask=slots" for default_pool).
+func airflowAPIUpsert(baseURL, resource, id, authHeader string, body []byte, updateSuffix string) error {
 	createURL := fmt.Sprintf("%s/%s", baseURL, resource)
 	respBody, status, err := airflowAPIRequest(http.MethodPost, createURL, authHeader, body)
 	if err != nil {
@@ -170,7 +171,7 @@ func airflowAPIUpsert(baseURL, resource, id, authHeader string, body []byte) err
 	}
 	if status == http.StatusConflict {
 		// Already exists — update via PATCH
-		updateURL := fmt.Sprintf("%s/%s/%s", baseURL, resource, id)
+		updateURL := fmt.Sprintf("%s/%s/%s%s", baseURL, resource, id, updateSuffix)
 		respBody, status, err = airflowAPIRequest(http.MethodPatch, updateURL, authHeader, body)
 		if err != nil {
 			return err
@@ -204,7 +205,7 @@ func AddVariables(airflowURL, authHeader string) error {
 		if err != nil {
 			return fmt.Errorf("error marshaling variable %s: %w", variable.VariableName, err)
 		}
-		if err := airflowAPIUpsert(airflowURL, "variables", variable.VariableName, authHeader, body); err != nil {
+		if err := airflowAPIUpsert(airflowURL, "variables", variable.VariableName, authHeader, body, ""); err != nil {
 			return fmt.Errorf("error adding variable %s: %w", variable.VariableName, err)
 		}
 		fmt.Printf("Added Variable: %s\n", variable.VariableName)
@@ -232,7 +233,7 @@ func AddConnections(airflowURL, authHeader string, envConns map[string]astrocore
 		if err != nil {
 			return fmt.Errorf("error marshaling connection %s: %w", conn.ConnID, err)
 		}
-		if err := airflowAPIUpsert(airflowURL, "connections", conn.ConnID, authHeader, body); err != nil {
+		if err := airflowAPIUpsert(airflowURL, "connections", conn.ConnID, authHeader, body, ""); err != nil {
 			return fmt.Errorf("error adding connection %s: %w", conn.ConnID, err)
 		}
 		fmt.Printf("Added Connection: %s\n", conn.ConnID)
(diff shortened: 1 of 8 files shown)


settings/settings.go
```diff
@@ -108,24 +108,68 @@ func InitSettings(settingsFile string) error {
...
+func buildBatchImportCommand(tmpFile, importCmd, jsonContent string) string {
+	return fmt.Sprintf("cat > %s <<'__ASTRO_CLI_EOF__'\n%s\n__ASTRO_CLI_EOF__\n%s %s; _ret=$?; rm -f %s; exit $_ret",
+		tmpFile, jsonContent, importCmd, tmpFile, tmpFile)
+}
+
 // AddVariables is a function to add Variables from settings.yaml
 func AddVariables(id string, version uint64) error {
 	variables := settings.Airflow.Variables
+
+	if version >= AirflowVersionTwo {
```
jeremybeard: We've long moved on from Airflow 1 so can we use this refactoring opportunity to rip all that out?

settings/settings.go
```diff
@@ -143,21 +187,133 @@ func AddConnections(id string, version uint64, envConns map[string]astrocore.Env
...
-		baseListCmd = baseCmd + "-l "
-		connIDArg = "--conn_id"
+		return addConnectionsBatch(id, connections)
+	}
+	return addConnectionsLegacy(id, connections)
+}
+
+// connectionImportObject builds the JSON value for a single connection in the
+// airflow connections import format. Returns nil if the connection should be skipped.
+func connectionImportObject(conn *Connection) interface{} {
```
jeremybeard: iirc a lot of the below translation logic already exists in the CLI?

settings/settings.go
jlaneve: done, ripped out all the airflow 1 codepaths - removed `addVariablesLegacy`, `addConnectionsLegacy`, `addPoolsLegacy`, `prepareAirflowConnectionAddCommand`, and the `version` parameter from all the Add*/ConfigSettings signatures + callers

settings/settings.go
jlaneve: yeah, `connectionImportObject` does similar field-by-field mapping as the old `prepareAirflowConnectionAddCommand`, but the output format is different enough that i didn't see a clean way to share it - the old code built CLI arg strings while this builds a json dict for `airflow connections import`. the old code is now deleted anyway with the airflow 1 removal so it's no longer duplicated

settings/settings.go
```diff
@@ -108,165 +108,169 @@ func InitSettings(settingsFile string) error {
 	return nil
 }
 
+// buildBatchImportCommand builds a compound shell command that writes jsonContent
+// to tmpFile via heredoc, runs importCmd against that file, then cleans up.
+func buildBatchImportCommand(tmpFile, importCmd, jsonContent string) string {
+	return fmt.Sprintf("cat > %s <<'__ASTRO_CLI_EOF__'\n%s\n__ASTRO_CLI_EOF__\n%s %s; _ret=$?; rm -f %s; exit $_ret",
```
jeremybeard: It seems like this will get in the way of us trying to support standalone mode for Windows. Is it possible to do it just with Airflow API calls?

jeremybeard (approved):

PR #2081, opened 2026-04-10
Add Airflow 2 support for standalone mode

pritt20:
## Description

Extends astro dev start --standalone to support Airflow 2 (Runtime 13.x) projects. The existing standalone flow was AF3-only; this PR removes that restriction and adds the conditional logic needed to handle AF2 differences.

Key changes:

- Detect Airflow major version from the runtime constraint file and branch accordingly
- Use pip freeze constraints (AF2 images don't ship uv)
- Configure LocalExecutor with SQLite for AF2 local dev
- On macOS, replace the default Gunicorn webserver with a Python shim that runs airflow webserver --debug (Flask dev server), avoiding fork-related SIGSEGV crashes
- Install a .pth startup patch in the venv that disables os.fork and no-ops setproctitle on macOS, preventing 100% CPU spins in forked children caused by corrupted ObjC/CoreFoundation/Network framework state
- Auto-deploy a plugin to fix QueuedLocalWorker pickling under macOS spawn-mode multiprocessing

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

Tested these changes locally:

image

## 📋 Checklist


airflow/standalone.go
```diff
@@ -924,8 +1172,10 @@ func (s *Standalone) ExportSettings(settingsFile, envFile string, connections, v
 	origExec := settings.SetExecAirflowCommand(s.standaloneExecAirflowCommand)
 	defer settings.SetExecAirflowCommand(origExec)
 
+	afVersion := s.airflowMajorVersionUint()
```
kaxil, line 1250: `ExportSettings` was correctly updated to use `airflowMajorVersionUint()`, but `applySettings` (line 846) and `ImportSettings` (line 1149) were missed. Both still hardcode `/api/v2` and call `fetchAirflowJWTToken`. AF2 has no `/api/v2` endpoint and no JWT auth -- after a successful AF2 startup, settings import will silently fail.

The repo already has version-aware `/api/v1` + Basic auth fallback in `cmd/api/airflow.go` (lines 291-345, 375-446). Same pattern should apply here.

airflow/standalone.go
```diff
@@ -88,15 +251,16 @@ var resolvePythonVersion = func(baseTag, tagPython string) string {
...
+	airflowHome         string
+	envFile             string
+	dockerfile          string
+	foreground          bool   // if true, run in foreground (stream output, block on Wait)
+	noBrowser           bool   // if true, don't open the browser after startup
+	port                string // webserver port; defaults to defaultStandalonePort
+	useProxy            bool   // whether the reverse proxy is active
+	proxyHostname       string // e.g. "my-project.localhost"
+	proxyPort           string // proxy listener port (default 6563)
+	airflowMajorVersion string // "2" or "3", determined from the runtime tag at Start()
```
kaxil, line 272: This field is only set inside `Start()`. Subcommands like `Logs`, `ImportSettings`, `ExportSettings`, `PS` all create a fresh `Standalone` via `StandaloneInit()` which leaves this empty.

When empty, `webserverPort()` takes the else branch and checks `WebserverPort` config even for AF3 projects. `buildEnv()` treats `""` as AF2 since the `if == "3"` check fails. `buildLogPrefixes` falls back to AF3 by coincidence.

Consider persisting this to a file in `.astro/standalone/` during `Start()` and reading it back in `StandaloneInit()`.

airflow/standalone.go
```diff
@@ -159,22 +341,22 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
-		// If it's an old-format tag (e.g., "12.0.0"), fall through to the AF3 check
 	}
-	if airflowversions.AirflowMajorVersionForRuntimeVersion(baseTag) != "3" {
+	afMajor := airflowversions.AirflowMajorVersionForRuntimeVersion(baseTag)
+	if afMajor != "2" && afMajor != "3" {
 		return errUnsupportedAirflowVersion
 	}
+	s.airflowMajorVersion = afMajor
 
 	pythonVersion := resolvePythonVersion(baseTag, tagPython)
```
kaxil, line 418: `GetDefaultPythonVersion(baseTag)` only searches `RuntimeVersionsV3` (see `airflow_versions.go:297`). AF2 runtimes like `12.0.0` or `11.14.0` won't match, so it falls through to the hardcoded `3.12`. AF2 runtimes 4-11 default to 3.9/3.10/3.11 -- fetching `runtime-12.0.0-python-3.12.txt` from CDN will 404.

The acceptance test masks this by pre-seeding `constraints-12.0.0-python-3.12.txt` instead of exercising real resolution.

airflow/standalone.go
```diff
@@ -67,6 +71,165 @@ var (
...
+        setproctitle.setthreadtitle = lambda *a, **kw: None
+    except ImportError:
+        pass
+`
+		if err := os.WriteFile(patchPy, []byte(pyContent), filePermissions); err != nil {
+			return err
+		}
+
+		patchPth := filepath.Join(sitePackages, "_fix_setproctitle.pth")
+		return os.WriteFile(patchPth, []byte("import _fix_setproctitle\n"), filePermissions)
```
kaxil: This `return` exits the loop after patching the first `pythonX.Y` directory. If the venv was recreated with a different Python version without cleaning `lib/`, a stale dir could sort before the active one. The patch lands in the wrong site-packages and macOS users still hit the crash.

Either patch all matching directories or resolve the active version from the `python` symlink and target only that one.

airflow/standalone.go
```diff
@@ -67,6 +71,165 @@ var (
...
+        sys.stdout.buffer.flush()
+    proc.stdout.close()
+
+def main():
+    af = [sys.executable, "-m", "airflow"]
+
+    # DB migrations
+    print("standalone | Running DB migrations...")
+    subprocess.run(af + ["db", "migrate"], check=True,
+                   stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
kaxil: `stdout=DEVNULL, stderr=DEVNULL` with `check=True` means the user sees a bare `CalledProcessError` when migration fails. Schema mismatches from a previous install are a common first-run failure -- the actual output from `airflow db migrate` is exactly what you need to debug it. At minimum, don't suppress stderr.

airflow/standalone.go
```diff
@@ -67,6 +71,165 @@ var (
...
+              "--password", "admin",
+              "--firstname", "Admin",
+              "--lastname", "User",
+              "--email", "admin@example.com"],
+        capture_output=True, text=True,
+    )
+    if result.returncode == 0:
+        print("standalone | Created admin user.  Login: admin / admin")
+    else:
+        print("standalone | Admin user already exists.  Login: admin / admin")
```
kaxil: Every non-zero exit prints "already exists". If the command fails for a real reason (DB connection error, bad install), the user gets a misleading success message and startup continues with no admin user. Check `result.stderr` for "already exists" rather than treating all failures as idempotent.

airflow/standalone.go
```diff
@@ -300,7 +482,38 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
+		shimPath := filepath.Join(venvBin, "_standalone_macos.py")
+		if err := os.WriteFile(shimPath, []byte(af2DarwinShim), filePermissions); err != nil {
+			return fmt.Errorf("error writing macOS standalone shim: %w", err)
+		}
+
+		// Disable os.fork and patch setproctitle on macOS.  After fork()
+		// the ObjC/CF/Network runtimes are corrupt, causing children to
+		// spin at 100 % CPU on getaddrinfo or setproctitle calls.
+		if err := writeDarwinForkSafetyPatch(filepath.Join(s.airflowHome, ".venv")); err != nil {
+			logger.Debugf("Warning: could not write Darwin fork-safety patch: %v", err)
```
kaxil: `logger.Debugf` is invisible at default verbosity. If this patch fails to install, AF2 on macOS will hit the exact 100% CPU fork crash it was designed to prevent, with no visible hint. Worth at least `logger.Warnf` or `fmt.Fprintf(os.Stderr, ...)`.

airflow/standalone.go
```diff
@@ -300,7 +482,38 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
+		}
+
+		// Write the LocalExecutor pickle fix plugin so QueuedLocalWorker
+		// can be serialized across macOS spawn-mode multiprocessing.
+		pluginsDir := filepath.Join(s.airflowHome, "plugins")
+		if err := os.MkdirAll(pluginsDir, dirPermissions); err == nil {
+			pickleFix := filepath.Join(pluginsDir, "fix_local_executor_pickle.py")
+			if _, err := os.Stat(pickleFix); os.IsNotExist(err) {
```
kaxil, line 566: Both the `MkdirAll` error (line 504) and the `WriteFile` error here are silently swallowed. If this file can't be written, macOS AF2 will fail at runtime with a cryptic pickling error that's much harder to diagnose than a clear "couldn't write plugin file" message at startup.

airflow/standalone.go
pritt20: Hey @kaxil ,

Thank you !!

I looked into https://updates.astronomer.io/astronomer-runtime and I see that we do not populate default python version for Airflow 3 runtimes too in json. Could it be that even AF3 versions are defaulting to 3.12 ? Is there anywhere I can find default python versions for AF2 and AF3 runtimes and I'll refactor this piece of code.

airflow/standalone.go
```diff
@@ -300,7 +560,42 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
+		shimPath := filepath.Join(venvBin, "_standalone_macos.py")
+		if err := os.WriteFile(shimPath, []byte(af2DarwinShim), filePermissions); err != nil {
+			return fmt.Errorf("error writing macOS standalone shim: %w", err)
+		}
+
+		// Disable os.fork and patch setproctitle on macOS.  After fork()
+		// the ObjC/CF/Network runtimes are corrupt, causing children to
+		// spin at 100 % CPU on getaddrinfo or setproctitle calls.
+		if err := writeDarwinForkSafetyPatch(filepath.Join(s.airflowHome, ".venv")); err != nil {
+			logger.Warnf("could not write Darwin fork-safety patch: %v", err)
```
kaxil: If the fork-safety patch silently fails the whole AF2/macOS path is pointless. User still gets SIGSEGV or 100% CPU spin but the run looks successful. Same concern for the pickle-plugin warning below. Should be hard errors, or at minimum logger.Errorf with the health check surfacing the failure.

airflow/standalone.go
```diff
@@ -559,19 +858,39 @@ func (s *Standalone) buildEnv() []string {
 	overrides["ASTRONOMER_ENVIRONMENT"] = "local"
 	overrides["AIRFLOW__CORE__LOAD_EXAMPLES"] = "False"
 	overrides["AIRFLOW__CORE__DAGS_FOLDER"] = filepath.Join(s.airflowHome, "dags")
-	overrides["AIRFLOW__CORE__SIMPLE_AUTH_MANAGER_ALL_ADMINS"] = "True"
+	overrides["AIRFLOW__CORE__PLUGINS_FOLDER"] = filepath.Join(s.airflowHome, "plugins")
```
kaxil: Setting this unconditionally changes behavior for AF3 too. Previously AF3 fell back to $AIRFLOW_HOME/plugins (which resolves to .astro/standalone/plugins); now it reads from <project>/plugins. Worth calling out in the PR description, or gating to AF2-only.

pkg/airflowrt/health.go
```diff
@@ -8,9 +8,15 @@ import (
 )
 
 // CheckHealth polls the Airflow health endpoint until it responds with 200 or the timeout is reached.
+// The airflowMajorVersion parameter selects the correct endpoint: "/api/v2/monitor/health" for AF3,
+// "/health" for AF2. An empty string defaults to AF3.
 // The provided context can be used to cancel the health check before the timeout expires.
-var CheckHealth = func(ctx context.Context, port string, timeout time.Duration) error {
-	url := fmt.Sprintf("http://localhost:%s/api/v2/monitor/health", port)
+var CheckHealth = func(ctx context.Context, port string, timeout time.Duration, airflowMajorVersion string) error {
```
kaxil: Adding airflowMajorVersion string to the exported CheckHealth signature is a breaking API change. If intentional, fine, but consider a functional option or making it a field on a receiver so we don't break callers every time we add a version-specific knob.

airflow/standalone_test.go
```diff
@@ -37,13 +38,45 @@ func (s *Suite) TestStandaloneHandlerInit() {
...
+	constraintsDir := filepath.Join(tmpDir, ".astro", "standalone")
+	err = os.MkdirAll(constraintsDir, 0o755)
+	s.NoError(err)
+	err = os.WriteFile(filepath.Join(constraintsDir, "constraints-12.0.0-python-3.12.txt"), []byte("apache-airflow==2.7.3+astro.1\n"), 0o644)
+	s.NoError(err)
+	err = os.WriteFile(filepath.Join(constraintsDir, "freeze-12.0.0-python-3.12.txt"), []byte("apache-airflow==2.7.3+astro.1\n"), 0o644)
+	s.NoError(err)
+
+	// Create fake airflow and python binaries (python needed for macOS shim path)
+	venvBin := filepath.Join(tmpDir, ".venv", "bin")
```
kaxil, line 57: Test doesn't pre-create .venv/lib/python*/site-packages, so on macOS writeDarwinForkSafetyPatch fails, gets logged as a warning, and the test still passes. Never actually asserts the fork patch installed. Either stub writeDarwinForkSafetyPatch or pre-create the site-packages dir and assert _fix_setproctitle.py exists.

airflow/standalone.go
```diff
@@ -596,6 +915,29 @@ func (s *Standalone) buildEnv() []string {
 	return env
 }
 
+// standaloneAPIURL returns the versioned API URL for the running standalone instance.
+// AF2 exposes /api/v1, AF3 exposes /api/v2.
+func (s *Standalone) standaloneAPIURL(port string) string {
```
kaxil, line 837: These new helpers (standaloneAPIURL, standaloneAuthHeader, airflowMajorVersionUint, readPersistedVersion, detectVersionFromDockerfile) don't have direct unit tests. They're the version-branching points that'll regress silently. Worth adding table-driven tests.

airflow/standalone.go
```diff
@@ -596,6 +915,29 @@ func (s *Standalone) buildEnv() []string {
...
+		return fmt.Sprintf("http://localhost:%s/api/v1", port)
+	}
+	return fmt.Sprintf("http://localhost:%s/api/v2", port)
+}
+
+// standaloneAuthHeader returns the Authorization header for the running standalone instance.
+// AF2 uses Basic auth (admin:admin); AF3 uses JWT from /auth/token.
+func (s *Standalone) standaloneAuthHeader(port string) string {
+	if s.airflowMajorVersion == "2" {
+		return "Basic " + base64.StdEncoding.EncodeToString([]byte("admin:admin"))
```
kaxil, line 853: admin:admin is hardcoded here and matches --password admin in the shim. If the user changes the admin password after first start, import/export will silently fail against the real credentials. OK for local dev but worth a comment noting the coupling.

airflow/standalone.go
```diff
@@ -159,22 +413,28 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
-			return fmt.Errorf("could not determine runtime version from Dockerfile image tag '%s'.\nStandalone mode requires a pinned Astronomer Runtime image (e.g., astro-runtime:3.1-12)", tag)
+			return fmt.Errorf("could not determine runtime version from Dockerfile image tag '%s'.\nStandalone mode requires a pinned Astronomer Runtime image (e.g., astro-runtime:3.1-12 or astro-runtime:13.5.0)", tag)
 		}
-		// If it's an old-format tag (e.g., "12.0.0"), fall through to the AF3 check
 	}
-	if airflowversions.AirflowMajorVersionForRuntimeVersion(baseTag) != "3" {
+	afMajor := airflowversions.AirflowMajorVersionForRuntimeVersion(baseTag)
+	if afMajor == "2" {
+		rtMajor := airflowversions.RuntimeVersionMajor(baseTag)
+		if rtMajor != "11" && rtMajor != "12" && rtMajor != "13" {
```
kaxil, line 343: nit: the {11, 12, 13} allow-list is hardcoded. Worth a comment pointing at where to bump this when runtime 14.x ships so it doesn't silently fall through to errUnsupportedAirflowVersion.

airflow/standalone.go
```diff
@@ -300,7 +560,42 @@ func (s *Standalone) Start(opts *types.StartOptions) error {
...
+		// `airflow webserver --debug` (Flask dev server, no fork).
+		shimPath := filepath.Join(venvBin, "_standalone_macos.py")
+		if err := os.WriteFile(shimPath, []byte(af2DarwinShim), filePermissions); err != nil {
+			return fmt.Errorf("error writing macOS standalone shim: %w", err)
+		}
+
+		// Disable os.fork and patch setproctitle on macOS.  After fork()
+		// the ObjC/CF/Network runtimes are corrupt, causing children to
+		// spin at 100 % CPU on getaddrinfo or setproctitle calls.
+		if err := writeDarwinForkSafetyPatch(filepath.Join(s.airflowHome, ".venv")); err != nil {
```
kaxil, line 490: nit: runs on every Start(). It's idempotent, but wasteful. Consider guarding with os.Stat the way the pickle plugin does right below.

pkg/airflowrt/health.go
pritt20: Hey @kaxil ,

I have added a new `HealthCheckConfig` config struct, so that in future if we ever have to add new config, we can just add to this struct without modifying signature.

airflow/docker.go
```diff
@@ -383,7 +383,7 @@ func (d *DockerCompose) Start(opts *airflowTypes.StartOptions) error {
 			wsPort = portOvr.WebserverPort
 		}
 		healthURL = fmt.Sprintf("http://localhost:%s/health", wsPort)
-		healthComponent = "webserver"
+		healthComponent = WebserverDockerContainerName
```
kaxil, line 386: Nit: unrelated to the AF2 work, pure `"webserver"` to `WebserverDockerContainerName` constant cleanup. Fine to keep, but ideally lives in its own refactor PR so the diff stays focused on the feature.

airflow/standalone.go
```diff
@@ -67,10 +73,192 @@ var (
...
+	for _, e := range entries {
+		if !e.IsDir() || !strings.HasPrefix(e.Name(), "python") {
+			continue
+		}
+
+		sitePackages := filepath.Join(libDir, e.Name(), "site-packages")
+
+		// Skip if already patched.
+		if _, err := os.Stat(filepath.Join(sitePackages, "_fix_setproctitle.pth")); err == nil {
+			continue
```
kaxil, line 118: Nit: the guard only checks `_fix_setproctitle.pth`. If someone deletes `_fix_setproctitle.py` but leaves the `.pth` behind, the next `Start()` skips re-installing. Edge case (users shouldn't be editing venv internals), but worth either checking both files or stating the invariant in the comment.

pkg/airflowrt/health.go
```diff
@@ -7,10 +7,24 @@ import (
...
+	AirflowMajorVersion string
+}
+
 // CheckHealth polls the Airflow health endpoint until it responds with 200 or the timeout is reached.
+// cfg.AirflowMajorVersion selects the correct endpoint: "/api/v2/monitor/health" for AF3,
+// "/health" for AF2. An empty AirflowMajorVersion defaults to AF3.
 // The provided context can be used to cancel the health check before the timeout expires.
-var CheckHealth = func(ctx context.Context, port string, timeout time.Duration) error {
-	url := fmt.Sprintf("http://localhost:%s/api/v2/monitor/health", port)
+var CheckHealth = func(ctx context.Context, port string, timeout time.Duration, cfg HealthCheckConfig) error {
```
kaxil, line 22: `airflowrt.CheckHealth` has no in-repo callers. `standalone.go` and `docker.go` both use the local `checkWebserverHealth` wrapper in `airflow/health_check.go`. The `HealthCheckConfig` refactor is the right shape, but worth either wiring this into the actual health-check path or removing it so we don't maintain two parallel health checkers.

kaxil (approved):
LGTM. All 8 of my prior review comments were addressed substantively. The `HealthCheckConfig` struct is a nice forward-compatible refactor, the error handling on the macOS fork-safety path is now hard-failing where it should, and the new table-driven tests cover the version-branching helpers well.

Leaving a couple of non-blocking nits inline that can be picked up here or in a follow-up.

airflow/docker.go
pritt20: Yeah, I included it here just to make linter job happy and it was just a 1 line change. 😅

airflow/standalone.go
```diff
@@ -67,10 +73,192 @@ var (
...
+`
+
+// af2DarwinShim is a self-contained Python script that replaces
+// `airflow standalone` on macOS for Airflow 2.  AF2's webserver uses
+// gunicorn which always forks workers; on macOS the fork inherits
+// corrupted Objective-C runtime state, causing SIGSEGV crashes and
+// the "Python quit unexpectedly" dialog.  This shim runs individual
+// components and passes --debug to the webserver so it uses Flask's
+// built-in dev server (no gunicorn, no fork).
+const af2DarwinShim = `#!/usr/bin/env python3
```
schnie: Might be worth pulling this into a standalone py file that we can exercise in test or two.

pritt20: Thanks @schnie! that make sense. I have captured all python scripts in separate files and referenced these in code.

pkg/airflowrt/health.go
pritt20: Yeah, this one was never in use, I'll clean this up in follow up PR if we are not planning to use this in future changes.

schnie (approved):

PR #2090, opened 2026-04-16
fix: skip Dockerfile parse when --image-name is provided

odaneau-astro:
## Summary

When `astro deploy --image-name ` is used, the image is pre-built and supplied externally. The CLI was unconditionally parsing the `Dockerfile` to extract the base image string for a warning — even though:

1. The build is skipped (`Skipping building image...`)
2. The runtime version is always obtained from the **image label**, not the `FROM` instruction

This caused a hard failure for projects that don't maintain a `Dockerfile` locally (e.g. repos where the image is built in a separate pipeline and referenced by digest/tag):

```
Skipping building image...
Error: failed to parse dockerfile: /path/to/Dockerfile: open /path/to/Dockerfile: no such file or directory
```

## Change

Move `docker.ParseFile` inside the `if version == ""` warning guard, and only execute it when no `--image-name` was given (i.e. a local build). When a pre-built image is provided, the missing label is referenced in the warning instead of the Dockerfile's `FROM` image.

```go
// before: always parsed
cmds, err := docker.ParseFile(filepath.Join(path, dockerfile))  // ← fails if no Dockerfile

// after: only parsed when needed for the warning
if imageName == "" {
    cmds, err := docker.ParseFile(...)
    warningImage = docker.GetImageFromParsedFile(cmds)
}
```

## Test plan

- [x] `TestBuildImageFailure` updated: the "dockerfile parsing error" sub-case now sets `GetLabel` to return `""` so the warning branch is entered (exercising the parse path)
- [x] New sub-case added: passes `imageName` with `Dockerfile.invalid` set — asserts `NoError`, proving the parse is skipped
- [x] Full `cloud/deploy` package test suite passes (`go test ./cloud/deploy/...`)## Summary

When `astro deploy --image-name ` is used, the image is pre-built and supplied externally. The CLI was unconditionally parsing the `Dockerfile` to extract the base image string for a warning — even though:

1. The build is skipped (`Skipping building image...`)
2. The runtime version is always obtained from the **image label**, not the `FROM` instruction

This caused a hard failure for projects that don't maintain a `Dockerfile` locally (e.g. repos where the image is built in a separate pipeline and referenced by digest/tag):

```
Skipping building image...
Error: failed to parse dockerfile: /path/to/Dockerfile: open /path/to/Dockerfile: no such file or directory
```

## Change

Move `docker.ParseFile` inside the `if version == ""` warning guard, and only execute it when no `--image-name` was given (i.e. a local build). When a pre-built image is provided, the missing label is referenced in the warning instead of the Dockerfile's `FROM` image.

```go
// before: always parsed
cmds, err := docker.ParseFile(filepath.Join(path, dockerfile))  // ← fails if no Dockerfile

// after: only parsed when needed for the warning
if imageName == "" {
    cmds, err := docker.ParseFile(...)
    warningImage = docker.GetImageFromParsedFile(cmds)
}
```

## Test plan

- [x] `TestBuildImageFailure` updated: the "dockerfile parsing error" sub-case now sets `GetLabel` to return `""` so the warning branch is entered (exercising the parse path)
- [x] New sub-case added: passes `imageName` with `Dockerfile.invalid` set — asserts `NoError`, proving the parse is skipped
- [x] Full `cloud/deploy` package test suite passes (`go test ./cloud/deploy/...`)

### DIFF
--- a/cloud/deploy/deploy.go
+++ b/cloud/deploy/deploy.go
@@ -41,7 +41,8 @@ const (
 	composeSkipImageBuildingPromptMsg = "Skipping building image..."
 	deploymentHeaderMsg               = "Authenticated to %s \n\n"
 
-	warningInvalidImageNameMsg = "WARNING! The image in your Dockerfile '%s' is not based on Astro Runtime and is not supported. Change your Dockerfile with an image that pulls from 'quay.io/astronomer/astro-runtime' to proceed.\n"
+	warningInvalidImageNameMsg         = "WARNING! The image in your Dockerfile '%s' is not based on Astro Runtime and is not supported. Change your Dockerfile with an image that pulls from 'quay.io/astronomer/astro-runtime' to proceed.\n"
+	warningInvalidPrebuiltImageNameMsg = "WARNING! The image '%s' does not appear to be based on Astro Runtime (the '%s' label is missing). Ensure your image is built FROM quay.io/astronomer/astro-runtime to proceed.\n"
 
 	allTests                 = "all-tests"
 	parseAndPytest           = "parse-and-all-tests"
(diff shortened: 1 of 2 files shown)


jlaneve (approved):
looks good — clean fix, the new prebuilt warning is more accurate than the old wording. tests cover both the parse-skipped and parse-still-runs paths. shipping.

PR #2093, opened 2026-04-17
Migrate CLI to v1 public API; retire v1beta1 and v1alpha1 (except IDE)

jeremybeard:
## Summary

Retires the `astro-client-core` (v1alpha1), `astro-client-iam-core` (iam/v1beta1), and `astro-client-platform-core` (platform/v1beta1) generated clients in favor of a single unified `astro-client-v1` client built from the public `versioned/v1.0` spec released in astronomer/astro#37986.

- New: `astro-client-v1/` — unified client, tags for Organization/Workspace/Cluster/Deployment/Deploy/User/Team/ApiToken/Environment/Authorization/etc.
- Renamed: `astro-client-core/` → `astro-client-v1alpha1/`, shrunk to IDE-only (`Polaris Project`, `Polaris Session` tags).
- Deleted: `astro-client-iam-core/`, `astro-client-platform-core/`, `cloud/platformclient/`.
- Two endpoints stay on v1alpha1 because the server has no v1 equivalent: `private/v1alpha1/cli/auth-config` (`pkg/astroauth/config.go`) and `/v1alpha1/telemetry` (`pkg/telemetry/telemetry.go`).
- Legacy tests (22 files) fully migrated; dead cmd-layer test paths covering collapsed split-client APIs were dropped (service-layer coverage remains).

### Notable production fixes surfaced during migration
- `cloud/deployment/inspect/inspect.go`: format `DeploymentEnvironmentVariable.UpdatedAt` as RFC3339 before mapstructure decode (v1 changed the field from `string` to `time.Time`; inspect was broken for any deployment with env vars).
- `cmd/cloud/deployment.go`: uppercase `--cloud-provider` input before validating against the `ClusterCloudProvider` enum (AWS/GCP/AZURE) so lowercase input keeps working.
- `cloud/deploy/bundle.go`: wire v1 `CreateDeployGitRequest` (GITHUB + GENERIC providers) into `CreateDeployRequest.Git` now that the v1 spec exposes it.
- `cloud/deployment/deployment.go` + `cloud/deployment/fromfile/fromfile.go`: stop synthesizing client-side defaults for optional deployment fields (scheduler size, hybrid scheduler AU/replicas, default worker queue, task-pod node pool). The server applies defaults; sending them from the CLI caused schema drift.
- `scripts/patch-v1-gen/` (Go): function-scope-aware rewrite of `oneOf` discriminator setters that oapi-codegen (≤ v2.6.0) emits with untyped string literals into pointer fields.

## Test plan
- [x] `go build ./...` clean
- [x] `make lint` clean
- [x] `go test ./...` passes
- [x] Live smoke against local astro (`localhost` context, organization `Sample Organization`, workspace `Sample Workspace`, cluster `Sample AWS Cluster`):
  - [x] `astro login localhost` — v1 `/users/self`
  - [x] `astro organization list` / `astro workspace list`
  - [x] `astro deployment list`, `astro deployment inspect <id>` (full + `--key`), `astro deployment logs <id>`
  - [x] `astro deployment create --type dedicated` — confirmed server applies defaults (scheduler size MEDIUM, default worker queue, task-pod cpu/memory) that the CLI no longer synthesizes
  - [x] `astro deployment update --description …` — inspect confirms change
  - [x] `astro deployment inspect --template` → `astro deployment create --deployment-file` (fromfile create), then `astro deployment update --deployment-file` (fromfile update); env-var `updated_at` decoded correctly (RFC3339 mapstructure fix)
  - [x] `astro deployment variable create/update/list` (including `--secret`)
  - [x] `astro deployment worker-queue create/update/delete`
  - [x] `astro deployment token create/list/update/rotate/delete` (identified by `-t <name>`)
  - [x] `astro workspace token create/list/update/rotate/delete`
  - [x] `astro organization token create/list/update/rotate/delete`
  - [x] `astro organization team create/list/delete`; `astro workspace team add/update/list`; `astro deployment team add/update/list` — exercises the new `workspaceId` / `deploymentId` query params on `GET /organizations/{orgId}/teams`
  - [x] `astro organization user list/invite/update`; `astro workspace user add/update/remove/list`; `astro deployment user add/update/remove/list`
  - [x] `astro deployment delete` — cleaned up both smoke deployments


cloud/deployment/deployment.go
```diff
@@ -72,6 +71,10 @@ var (
 	dagDeployEnabled bool
 )
 
+// defaultWorkerMachineNam is the machine type UpdateDeployment falls back to when
+// an executor change forces it to synthesize a replacement worker queue.
+const defaultWorkerMachineNam = "A5"
```
akshayarora: I assume this is a typo? Should be `defaultWorkerMachineName`??

akshayarora (approved):
Thanks for the migration work. Left one comment.

cloud/deployment/deployment.go
```diff
@@ -425,27 +428,13 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
 		} else {
 			developmentModeValue = false
 		}
-		var workerConcurrency int
```
aliotta, line 438: im assuming this deletion is intentional?

cloud/deployment/deployment.go
```diff
@@ -458,69 +447,58 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
...
-				Type:                 astroplatformcore.CreateStandardDeploymentRequestTypeSTANDARD,
+				Type:                 new(astrov1.CreateStandardDeploymentRequestTypeSTANDARD),
 				DefaultTaskPodCpu:    defaultTaskPodCPUPtr,
 				DefaultTaskPodMemory: defaultTaskPodMemoryPtr,
 				ResourceQuotaCpu:     resourceQuotaCPUPtr,
 				ResourceQuotaMemory:  resourceQuotaMemoryPtr,
 				WorkloadIdentity:     deplWorkloadIdentity,
 				RemoteExecution:      remoteExecution,
 			}
-			if !remoteExecutionEnabled {
```
aliotta, line 511: im assuming this deletion is intentional?

cloud/deployment/deployment.go
```diff
@@ -529,55 +507,44 @@ func Create(name, workspaceID, description, clusterID, runtimeVersion, dagDeploy
...
-				Type:                 astroplatformcore.CreateDedicatedDeploymentRequestTypeDEDICATED,
+				Type:                 new(astrov1.CreateDedicatedDeploymentRequestTypeDEDICATED),
 				DefaultTaskPodCpu:    defaultTaskPodCPUPtr,
 				DefaultTaskPodMemory: defaultTaskPodMemoryPtr,
 				ResourceQuotaCpu:     resourceQuotaCPUPtr,
 				ResourceQuotaMemory:  resourceQuotaMemoryPtr,
 				WorkloadIdentity:     deplWorkloadIdentity,
 				RemoteExecution:      remoteExecution,
 			}
-			if !remoteExecutionEnabled {
```
aliotta, line 571: im assuming this deletion is intentional?

aliotta:
I wonder how difficult it would be to wire up some component type integration tests that use the cli such that refactors like this do not require the full suite of manual tests you had to run.

cloud/deployment/deployment.go
jeremybeard: Yes, very recently the create deployment API was changed to make the worker queues optional, so dropping this logic. There's more we could do later here to have the server apply even more defaults.

aliotta (approved):

cloud/deployment/deployment.go
```diff
@@ -1256,76 +1217,76 @@ func Update(deploymentID, name, ws, description, deploymentName, dagDeploy, exec
 			schedulerReplicas = currentDeployment.SchedulerReplicas
 		}
 		if workloadIdentity == "" {
-			if currentDeployment.WorkloadIdentity != nil {
-				workloadIdentity = *currentDeployment.WorkloadIdentity
+			if currentDeployment.EffectiveWorkloadIdentity != nil {
```
jlaneve: heads up, the swap from `WorkloadIdentity` to `EffectiveWorkloadIdentity` is forced (v1's `Deployment` only exposes the effective one, per the field doc: "either user-configured or Astronomer-provided default"), but it does shift behavior in this fallback path

previously, if the deployment had no user-configured identity, `currentDeployment.WorkloadIdentity` was nil and we skipped the assignment, so the Update sent no identity and the server kept managing it. now, `EffectiveWorkloadIdentity` is non-nil whenever a default is in effect, so an Update without `--workload-identity` on a previously-server-managed deployment will stamp the Astronomer-provided default as the requested value going forward

is that intentional? inspect.go:279-280 has the matching shift so the inspect template also bakes the effective identity in, which means `inspect --template` -> `create --deployment-file` round-trips pin the default too

longer-term, probably worth flagging back to the api team that the read model could expose the user-configured identity separately (or a flag indicating whether the effective value was user-set vs defaulted) so we can restore the prior nil-means-skip behavior. either way a short code comment noting the semantic difference would help future readers

jlaneve (approved):
lgtm, awesome cleanup. left one non-blocking note re the `EffectiveWorkloadIdentity` semantics shift, can be followed up separately

cloud/deployment/deployment.go
jeremybeard: Good call, have changed it so the effective value is not carried through to the update request, but the user can still override with `--workload-identity`.

PR #2115, opened 2026-04-30
feat(env): add astro env command tree (var, conn, airflow-var, metrics-export)

jlaneve:
## Summary

Closes [AI-475](https://linear.app/astronomer/issue/AI-475/add-astro-env-command-tree-for-environment-manager-env-objects).

Adds a new `astro env` command tree for managing **Environment Objects** through the Astro Environment Manager from the CLI. Covers all four object types the public API exposes: environment variables, connections, Airflow variables, and Universal Metrics Exports. Each type supports list / get / create / update / delete (env vars also support `export` to a dotenv file), scoped to either a workspace or a deployment, with `--auto-link` for the workspace-scoped "auto-link to all deployments" toggle.

Motivation: a public commenter asked for variable export from a workspace and built their own wrapper because the CLI did not expose the env-manager API. The existing `astro deployment variable` and `astro deployment connection` commands target different stores and do not see workspace-level objects, so this is additive: the new tree fills the gap, and the legacy commands remain untouched aside from a one-line cross-reference in their help text.

Worth calling out: deployment-scoped `.env` import/export already exists today via `astro deployment variable list --save` / `create --load`. **Workspace-scoped** export is genuinely new — Astronomer's own docs note that "bulk export of workspace-level environment variables is not supported" today, which is exactly the gap users are filing tickets about.

## Commands

```
astro env var             list | get | create | update | delete | export
astro env conn            list | get | create | update | delete
astro env airflow-var     list | get | create | update | delete
astro env metrics-export  list | get | create | update | delete
```

Persistent flags on each subroot:

```
--workspace-id      Defaults to the current workspace from context
--deployment-id     Mutually exclusive with --workspace-id
--include-secrets   Surfaces secret values (subject to org policy)
--resolve-linked    Default true; pass false to see addressable IDs
```

Type-specific flags on `create` / `update`:
- `var` and `airflow-var`: `--key`, `--value`, `--secret`
- `conn`: `--type`, `--host`, `--login`, `--password`, `--port`, `--schema`, `--extra` (JSON)
- `metrics-export`: `--endpoint`, `--exporter-type`, `--auth-type`, `--basic-token`, `--username`, `--password`, `--sigv4-assume-arn`, `--sigv4-sts-region`, `--header KEY=VAL` (repeatable), `--label KEY=VAL` (repeatable)

Shared on every create / update:
- `--auto-link` — Workspace scope only. Auto-links the object to every deployment in the workspace, including future ones. Rejected at deployment scope with a clear error. **Note**: due to a server-side bug ([AINF-1792](https://linear.app/astronomer/issue/AINF-1792)), the platform currently clears `autoLinkDeployments` whenever an update body omits the field, so users wanting to keep the flag set must re-pass `--auto-link` on every update. Once AINF-1792 lands, our tri-state plumbing (omit = no change) will work as designed.

`update` is upsert by default; pass `--strict` to fail when the target does not exist.

### Pull workspace env vars into a local `.env`

The originating ask. Workspace export is the genuinely new capability here; deployment export already exists via `astro deployment variable list --save`.

```bash
astro workspace switch my-ws
astro env var export > .env
astro dev start
```

Add `--include-secrets` to materialize secret values too (subject to the org-level **Environment Secrets Fetching** policy; the CLI prints a sensitive-data warning to stderr so the warning never lands in the dotenv file).

### Set a workspace-level connection auto-linked to every deployment

The recommended pattern for sharing a connection across all deployments in a workspace, in one command:

```bash
echo "$DB_PASSWORD" | astro env conn create \
  --workspace-id ws-prod \
  --key main_postgres \
  --type postgres \
  --host db.example.com \
  --port 5432 \
  --login analytics \
  --schema public \
  --auto-link
```

Drop `--auto-link` to create the workspace object without auto-linking (you can then link to specific deployments via the UI; explicit `astro env link` lands in a follow-up).

### Toggle auto-link on an existing object

```bash
astro env var update SHARED_API_KEY --workspace-id ws-prod --value $(pass api/key) --auto-link
astro env var update SHARED_API_KEY --workspace-id ws-prod --value $(pass api/key) --auto-link=false
```

### Deployment view that includes objects linked from the workspace

Useful for verifying what a deployment will actually see at runtime. The platform's resolve-linked view returns both explicitly-linked and auto-linked objects.

```bash
# resolve-linked is the default; explicit here for clarity.
astro env var list --deployment-id dep-abc --resolve-linked

# To get addressable IDs (e.g. so you can update by ID rather than key),
# disable resolution. Linked-from-workspace rows will not appear.
astro env var list --deployment-id dep-abc --resolve-linked=false
```

### Agent-friendly inventory across types

Stable JSON shape, predictable exit codes, no interactive prompts when flags are set.

```bash
WS=ws-prod
for kind in var conn airflow-var metrics-export; do
  echo "## $kind"
  astro env "$kind" list --workspace-id "$WS" --format json | jq -r '.[] | .objectKey'
done
```

### Bulk variable import for connections / workspace-scope vars

The platform already supports `astro deployment variable create --load --env` for **deployment-scoped** env vars, and the UI supports bulk Airflow-variable import. Connection bulk import is intentionally not supported through the UI for security reasons. This loop fills the workspace-env-var gap, with the security consideration that values flow through stdin (not shell history):

```bash
while IFS='=' read -r k v; do
  [[ -z "$k" || "$k" =~ ^# ]] && continue
  printf '%s' "$v" | astro env var create --workspace-id ws-prod --key "$k" --auto-link
done < .env.production
```

### Rotate a secret in place

Astronomer's docs don't currently describe a rotation flow (secret values are write-only once created). `--strict` keeps `update` from accidentally creating a new var if the original was deleted between rotations:

```bash
read -sp "new value: " v && echo "$v" | astro env var update API_TOKEN --strict
```

### Update a connection partially

Connection updates are partial: flags you don't pass are left alone (the platform's documented "override values for the fields you want to edit" model). Note that `--type` is required by the API on every update.

```bash
echo "$NEW_PW" | astro env conn update main_postgres --type postgres
```

### Promote a deployment-only var up to workspace scope

There's no in-platform "promote" or "migrate" affordance today. This three-step works:

```bash
VAL=$(astro env var get SHARED_KEY --deployment-id dep-abc --include-secrets --format json | jq -r '.environmentVariable.value')
printf '%s' "$VAL" | astro env var create --workspace-id ws-prod --key SHARED_KEY --secret --auto-link
astro env var delete SHARED_KEY --deployment-id dep-abc --yes
```

### Cross-workspace promotion (bridge pattern)

Astronomer's preferred sharing model is **per-deployment overrides on a single workspace-scoped object**, not duplication across workspaces. This flow is a bridge for setups where dev and prod live in separate workspaces:

```bash
astro env var get API_TOKEN --workspace-id ws-dev --include-secrets --format json \
  | jq -r '.environmentVariable.value' \
  | astro env var create --workspace-id ws-prod --key API_TOKEN --secret
```

## Security choices

1. `--value` and `--password` accept stdin (piped) or a TTY prompt with echo disabled when the flag is omitted, so secret values do not land in shell history. The flag still works for scripts but is documented as discouraged. This matches `wrangler secret put`, `gh secret set`, and `op` conventions.
2. `export --include-secrets` and any list with `--include-secrets` print a sensitive-data warning to stderr, leaving stdout clean for piping into a file.
3. `delete` refuses non-TTY callers without `--yes`; TTY callers get an interactive y/N prompt. CI scripts must be explicit about destructive operations.
4. Connection passwords are opt-in: omitting `--password` on a TTY does not block on a hidden prompt, since many connection types are passwordless.

## API quirks worth knowing for review

- The platform's list endpoint blanks IDs when `resolveLinked=true` (the default for list/export) because those rows reference a resolved view rather than addressable objects. The CLI's table view auto-hides the ID column when none of the rows have IDs. Pass `--resolve-linked=false` to see them.
- There is no GET-by-key endpoint, so key lookups go through list with the `objectKey` filter and `resolveLinked=false`, then match client-side. `update` and `delete` skip the lookup entirely when the user passes a CUID.
- `create` returns only an ID. The CLI synthesizes the full response object from the create inputs + returned ID for all four types, avoiding an extra GET round-trip.
- `--auto-link` only makes sense at workspace scope (deployment-scoped objects are already pinned to a single deployment). The CLI rejects `--auto-link` at deployment scope before any network call.

## Reuse

- All HTTP calls go through the auto-generated `astro-client-core` client (oapi-codegen). No hand-rolled HTTP code.
- Exported `IsCUID` from `cloud/organization` (now wraps `lucsky/cuid.IsCuid`).
- Exported `IsSecretsFetchingNotAllowedError` and `SecretsFetchingNotAllowedErrMsg` from `cloud/environment`. Single source of truth for the "secrets fetching not enabled" guidance.
- `readSecretValue` and `confirmTTY` delegate to `pkg/input.Password` and `pkg/input.Confirm` for the TTY paths.

## What is not in this PR

Tracked separately for follow-ups:

- **Per-deployment overrides** of a workspace object. The platform's preferred sharing model and a UI-first concept; not yet exposed via the public env-object API in a clean way.
- **Restart/propagation warnings**: docs note that deployment-level env-var saves restart scheduler/webserver/workers and take "up to two minutes." Worth printing on writes once this PR lands.
- **`astro env link` / `unlink` / `exclude` commands** for explicit (non-auto) linking.
- **First-class `--from-file` bulk import**.
- **`astro env var run -- <cmd>` to inject env vars into a subprocess** (the convergent feature across Vercel, Railway, Doppler, and 1Password).

## Test plan

- [x] `go test ./...` green
- [x] `gofumpt` and `golangci-lint run` clean
- [x] Adversarial smoke against a throwaway workspace covering all four types: list (empty and populated), create (flag value, piped stdin, prompt for password on connections), get (by key and by ID), update (value-only and partial), delete (`--yes` and interactive), export (dotenv stdout, secrets masked, `--include-secrets` warns to stderr). Edge cases included empty values, multiline values, unicode, 5000-char values, conflict on duplicate key, `--strict` upsert refusal, dotenv on connection (rejected), bad JSON in `--extra`, file output to non-existent dir, `--include-secrets` against an org without policy enabled.
- [x] `--auto-link` end-to-end: create with the flag sets `autoLinkDeployments: true`, omitting it leaves the field unset, `update --auto-link` toggles on, `update --auto-link=false` toggles off, deployment-scope create with `--auto-link` is rejected pre-API
- [x] `coalesceWorkspace` fallback verified: every command honors the current workspace from context when neither scope flag is set
- [x] Cross-reference help text rendered correctly in `astro deployment variable --help`
- [x] Workflow examples cross-checked against [Manage env vars](https://www.astronomer.io/docs/astro/manage-env-vars), [Create and link connections](https://www.astronomer.io/docs/astro/create-and-link-connections), [Create and link variables](https://www.astronomer.io/docs/astro/create-and-link-variables), and [Import and export Airflow connections and variables](https://www.astronomer.io/docs/astro/import-export-connections-variables); terminology aligned to "Environment Objects", "linked", "Universal Metrics Export", "Environment Secrets Fetching"

---


cmd/cloud/env_airflow_var.go
```diff
@@ -0,0 +1,235 @@
...
+  # Update
+  astro env airflow-var update MY_VAR --workspace-id <ws-id> --value new-value
+
+  # Delete
+  astro env airflow-var delete MY_VAR --workspace-id <ws-id> --yes
+`
+
+func newEnvAirflowVarRootCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "airflow-var",
```
neel-astro: Just a self-preference: I would prefer airflow-variable, connection, variable instead of airflow-var, conn, and var as the default command. We already have them as aliases, so not much harm, but it feels like the right thing in terms of CLI documentation for first-time Airflow or Astro users

cloud/env/objects.go
```diff
@@ -0,0 +1,190 @@
...
+// includeSecrets requests secret values from the server (subject to org policy).
+func listObjects(scope Scope, objectType astrocore.ListEnvironmentObjectsParamsObjectType, resolveLinked, includeSecrets bool, coreClient astrocore.CoreClient) ([]astrocore.EnvironmentObject, error) {
+	if err := scope.Validate(); err != nil {
+		return nil, err
+	}
+	c, err := config.GetCurrentContext()
+	if err != nil {
+		return nil, err
+	}
+	params := buildListParams(scope, objectType, nil, resolveLinked, includeSecrets, defaultListLimit)
```
neel-astro: Can we please add pagination support for these new API calls

cmd/cloud/env_conn.go
```diff
@@ -0,0 +1,275 @@
...
+	if envConnPort != 0 {
+		in.Port = &envConnPort
+	}
+	if envConnExtra != "" {
+		var extra map[string]any
+		if err := json.Unmarshal([]byte(envConnExtra), &extra); err != nil {
+			return env.ConnInput{}, fmt.Errorf("--extra is not valid JSON: %w", err)
+		}
+		in.Extra = &extra
+	}
```
neel-astro: I think this would conflict with a user trying to unset extra params on a connection. Cobra allows us to distinguish between a flag not passed in the command vs a flag passed as a default value explicitly, which we could utilize to make the right call for all the optional flags. I think we have already done the same for the password flag below, which uses the `Changed` method

cmd/cloud/env_metrics.go
```diff
@@ -0,0 +1,282 @@
...
+		return errors.New("aborted: pass --yes (or confirm interactively) to delete")
+	}
+	if err := env.DeleteMetricsExport(idOrKey, scope, astroCoreClient); err != nil {
+		return err
+	}
+	fmt.Fprintf(out, "Deleted %s\n", idOrKey)
+	return nil
+}
+
+func buildMetricsInput(cmd *cobra.Command) (*env.MetricsInput, error) {
```
neel-astro, line 242: Same as the previous comment in buildConnInput, can we make the optional fields be set explicitly as empty?

cmd/cloud/env_var.go
```diff
@@ -0,0 +1,270 @@
...
+	_ = cmd.MarkFlagRequired("key")
+	return cmd
+}
+
+func newEnvVarUpdateCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "update <id-or-key>",
+		Aliases: []string{"up"},
+		Short:   "Update an environment variable's value",
+		Long:    "Update the value of an existing environment variable. The platform API does not allow toggling the secret flag; delete and recreate to change it.",
```
neel-astro: The default behaviour is Upsert and not update, which feels slightly confusing because we are naming the command as update. I think for other similar cases, we end up calling out that the resource does not exist and ask the user for input if they would like to move forward with creating the resource.

cloud/env/objects.go
```diff
@@ -0,0 +1,190 @@
...
+		if err := astrocore.NormalizeAPIError(resp.HTTPResponse, resp.Body); err != nil {
+			return nil, err
+		}
+		return resp.JSON200, nil
+	}
+
+	if err := scope.Validate(); err != nil {
+		return nil, err
+	}
+	params := buildListParams(scope, objectType, &idOrKey, false, includeSecrets, 2)
```
neel-astro, line 123: curious Q: Why the limit 2 in this case? Also, should we try to reuse `listObjects` over here instead of re-implementing the list logic in the `getObject` method again?

cloud/env/objects.go
```diff
@@ -0,0 +1,190 @@
...
+	case s.WorkspaceID != "" && s.DeploymentID != "":
+		return ErrScopeAmbiguous
+	}
+	return nil
+}
+
+// listObjects returns env-objects of the given type within the scope.
+// resolveLinked includes inherited workspace objects when listing at deployment scope.
+// includeSecrets requests secret values from the server (subject to org policy).
+func listObjects(scope Scope, objectType astrocore.ListEnvironmentObjectsParamsObjectType, resolveLinked, includeSecrets bool, coreClient astrocore.CoreClient) ([]astrocore.EnvironmentObject, error) {
```
neel-astro, line 63: This and `listEnvironmentObjects` in `cloud/environment/environment.go` are almost doing the same thing. Can we, in this PR or a subsequent one, clean up to use just one of them?

cloud/env/conn.go
```diff
@@ -0,0 +1,144 @@
...
+	Login               *string
+	Password            *string
+	Schema              *string
+	Port                *int
+	Extra               *map[string]any
+	AutoLinkDeployments *bool
+}
+
+// ListConns returns CONNECTION objects for the given scope.
+func ListConns(scope Scope, resolveLinked, includeSecrets bool, coreClient astrocore.CoreClient) ([]astrocore.EnvironmentObject, error) {
```
neel-astro, line 31: Same here, we should replace the usage of `ListConnections` from `cloud/environment/environment.go` with this method

cloud/env/format.go
```diff
@@ -0,0 +1,387 @@
...
+		o := &envObjs[i]
+		if o.EnvironmentVariable == nil {
+			continue
+		}
+		isSecret := o.EnvironmentVariable.IsSecret
+		switch {
+		case isSecret && !includeSecrets:
+			fmt.Fprintf(out, "%s=  # secret, use --include-secrets\n", o.ObjectKey)
+		default:
+			fmt.Fprintf(out, "%s=%s\n", o.ObjectKey, o.EnvironmentVariable.Value)
```
neel-astro: I think we would have to escape the value over here for edge cases when it has `=` or `\`. Also, you can look into the writer from https://github.com/joho/godotenv to help with that, instead of re-implementing the wheel

cloud/env/var.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := validateAutoLink(scope, autoLink); err != nil {
+		return nil, err
+	}
+	c, err := config.GetCurrentContext()
+	if err != nil {
+		return nil, err
+	}
+	scopeType, scopeEntityID := scopeRequest(scope)
+
+	body := astrocore.CreateEnvironmentObjectJSONRequestBody{
```
neel-astro, line 44: I think all the create and update logic in connection, metrics, env var, and airflow var files are almost the same, with the only diff being the object value and object type, which is just a pass-through in this logic. So we can maybe keep the logic in a single common function and invoke that instead of maintaining 4 separate copies.

neel-astro (commented):
Thanks for adding the support ❤️ , left a few comments, but it looks like a solid base

cmd/cloud/env_var.go
jlaneve: kept upsert as the default but reworded the Short and Long to make that explicit. short now reads "Set ... (creates if missing; use --strict to require existing)". I think prompting on TTY would break script usage and --strict already gives the opt-out, plus this matches what astro deployment variable update already does (also upsert). open to revisiting if you'd rather have the prompt path

cloud/env/objects.go
jlaneve: trimmed the comment to drop the misleading "signal duplicates" framing - the code just picks the first match. didn't refactor getObject to call listObjects since they need different param shapes (objectKey filter, forced resolveLinked=false, custom limit) and the helper would end up just as long with more parameters. the duplication is mostly the request boilerplate which is small

cloud/env/var.go
jlaneve: thought about this one and I want to push back. the four files are intentionally parallel, and the //nolint:dupl at the top documents that choice. a helper would need two callbacks per call site (typed body + typed response) plus per-type validation pulled out separately, since conn requires type, metrics requires endpoint + exporter, and vars require neither. also worth noting the request and response enums are two different generated types (CreateEnvironmentObjectRequestObjectType vs EnvironmentObjectObjectType), so centralizing means trusting the strings round-trip rather than catching drift at compile time. net savings would be ~15 lines and each call site gets harder to read top to bottom. happy to revisit if a fifth type lands

neel-astro (approved):
LGTM 🚀

PR #2117, opened 2026-05-01
fix(otto): Refresh update cache before auto-update, not after

kaxil:
## Summary

`astro otto` did not auto-download v0.1.1 the morning it released. The 24h TTL was a contributor, but the underlying bug was ordering: the cache refresh ran in a fire-and-forget goroutine *after* `autoUpdate` had already read stale data. The cache only caught up during the session for the *next* invocation's benefit, so the first launch after every release missed the new version.

Fixes [AI-485](https://linear.app/astronomer/issue/AI-485/otto-does-not-auto-download-latest-version).

## What changed

- **Move the refresh to before the check, synchronously.** `refreshUpdateCacheIfStale` now runs in `Start()` *before* `autoUpdate` / `hintUpdateAvailable` read the cache, instead of as a post-launch goroutine. The CDN endpoint returns a tiny version string bounded by `LatestVersion`'s existing 5s timeout.
- **Lower TTL from 24h → 1h.** Even when the cache is "fresh," a 24h window could hide a recent release for a full day. 1h shrinks the worst-case miss window to a single hour.
- **Negative-cache failures.** On network failure, `LastCheck` still advances (preserving `LatestKnown` so the hint keeps working). Otherwise a blocked CDN would impose a 5s wait on every launch.

## Design rationale

- *Why synchronous and not just shorten the goroutine timer?* Because the bug isn't TTL length — it's that the goroutine runs *after* the decision it should inform. Any value of "refresh later for next time" still misses the current launch.
- *Why keep the cache at all instead of always fetching?* The cache lets us skip the network call on rapid relaunches and serves as a graceful fallback when offline. Dropping it entirely would mean a 5s timeout on every launch when the CDN is blocked.
- *Why 1h and not 5min?* 1h is what the issue thread suggested. It's short enough that releases are picked up promptly while still avoiding chatty CDN traffic from users who launch Otto many times per hour.
- *Why bump `LastCheck` on failure?* Without it, a stale-cache + blocked-CDN environment would re-pay the 5s timeout on every launch indefinitely. With it, that user sees one 5s wait per hour.

## Tradeoffs

- Worst-case startup penalty in a blocked-CDN environment is now visible (5s once per hour), where previously it was invisible because the network call lived in a goroutine. Acceptable: the alternative (the bug we're fixing) was that releases never arrived without manual `astro otto update`.
- Users who launched within the last hour of a release still won't get it on the very next invocation. That window is unavoidable without dropping the cache entirely.

### DIFF
--- a/pkg/otto/start.go
+++ b/pkg/otto/start.go
@@ -45,6 +45,14 @@ func Start(args []string) error {
 		return fmt.Errorf("setting up otto: %w", err)
 	}
 
+	// Refresh the cached latest version up-front so autoUpdate / hint paths
+	// see today's release, not yesterday's. Previously this ran post-autoUpdate
+	// in a goroutine, which meant the very first launch after a release missed
+	// the new version (cache only caught up by the *next* invocation). Bounded
+	// by LatestVersion's 5s timeout; offline launches soft-fall-through to the
+	// previously cached value.
+	refreshUpdateCacheIfStale()
+
 	// Apply auto-update or fall back to the hint. Both must run before
 	// redirectCLILogs — post-redirect, writes would land in a log file the
 	// user never opens. autoUpdate downloads synchronously when the cache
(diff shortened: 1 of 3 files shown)


jlaneve (approved):
lgtm, also ran /jl-simplify and pushed a small follow-up trim

PR #2119, opened 2026-05-04
feat(env): add --from-file bulk import for env and airflow variables

jlaneve:
## Summary

Adds `--from-file <path>` to `astro env variable create/update` and `astro env airflow-variable create/update`, completing the round-trip with the `export` command shipped in #2115.

## Behavior

```bash
# Bulk create from a dotenv file
astro env variable create --workspace-id <ws> --from-file .env

# Bulk upsert; pass --strict to require existing
astro env variable update --workspace-id <ws> --from-file .env

# Pipe from stdin (Unix `-` convention)
some-secret-mgr export | astro env variable create --workspace-id <ws> --from-file -

# Cross-workspace promotion
astro env variable export --workspace-id ws-dev | \
  astro env variable update --workspace-id ws-prod --from-file -
```

`--secret`, `--auto-link`, `--strict` apply uniformly. `--from-file` is mutually exclusive with `--key`/`--value` and the positional `<id-or-key>`. Parse errors abort before any creates run; per-entry runtime errors fail-fast and name the offending key.

## Parser choice

Uses `subosito/gotenv`. Empirically: `joho/godotenv` 1.5.1 mishandles `\"` and `\\`, so the export→import round-trip is broken for values containing quotes or backslashes. `subosito/gotenv` round-trips every escape form `dotenvQuote` produces.

## Real-environment round-trip (verified)

Sandbox workspace, full edge-case set: bulk create (8 entries), export → byte-equal diff, bulk upsert (8 existing + 1 new), export again → byte-equal diff, `--strict` rejection of missing key, stdin pipe `export | update --from-file -`. All green.

## Known limitations

- **POSIX-style keys only** (dotenv format constraint, enforced by `subosito/gotenv`). Airflow variables with non-POSIX keys (e.g. `team.config`, `dag-name`) can't be bulk-imported here; use single-key `--key`/`--value` create instead, or wait for the JSON/YAML `--from-file` follow-up.
- **No batch endpoint**. N entries = N sequential API calls.
- **Round-trip via dotenv is `variable`-only**. `astro env airflow-variable list --format json` (or `yaml`) already produces a portable export today, but airflow-vars explicitly reject dotenv format, so the dotenv `--from-file` we ship here doesn't pair symmetrically with an `export` command for that type. A future JSON/YAML `--from-file` would close that gap.

## Test plan

- [x] `go test ./...` green
- [x] `go vet ./...` clean
- [x] `golangci-lint run` clean
- [x] 37+ binary smoke cases (parser, CLI validation, stdin, edge cases)
- [x] Real round-trip against a live workspace
- [x] Examples in `--help` updated
- [x] No regression on single-key path

### DIFF
--- a/cloud/env/parse_test.go
+++ b/cloud/env/parse_test.go
@@ -31,7 +31,8 @@ func (s *Suite) TestParseDotenvRoundTripsExport() {
 
 	parsed, err := ParseDotenvFile(path)
 	s.NoError(err)
-	for _, o := range objs {
+	for i := range objs {
+		o := &objs[i]
 		s.Equal(o.EnvironmentVariable.Value, parsed[o.ObjectKey], "mismatch for %s", o.ObjectKey)
 	}
 }
--- a/cloud/env/parse_test.go
+++ b/cloud/env/parse_test.go
@@ -1,0 +1,84 @@
+package env
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+
+	astrocore "github.com/astronomer/astro-cli/astro-client-core"
+)
+
+func (s *Suite) TestParseDotenvRoundTripsExport() {
+	// Round-trip: write a set of values via the dotenv writer, read them back
+	// via the parser, assert byte-identical values. Covers every escape path
+	// dotenvQuote produces (whitespace, newline, quote, $, #, \, export prefix).
+	objs := []astrocore.EnvironmentObject{
+		{ObjectKey: "PLAIN", EnvironmentVariable: &astrocore.EnvironmentObjectEnvironmentVariable{Value: "simple"}},
+		{ObjectKey: "WITH_SPACES", EnvironmentVariable: &astrocore.EnvironmentObjectEnvironmentVariable{Value: "two words"}},
+		{ObjectKey: "WITH_NL", EnvironmentVariable: &astrocore.EnvironmentObjectEnvironmentVariable{Value: "line1\nline2"}},
+		{ObjectKey: "WITH_QUOTE", EnvironmentVariable: &astrocore.EnvironmentObjectEnvironmentVariable{Value: `say "hi"`}},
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -49,6 +49,7 @@ require (
 	github.com/golang-jwt/jwt/v4 v4.5.2
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/itchyny/gojq v0.12.18
+	github.com/joho/godotenv v1.5.1
```
neel-astro, line 52: I don't think we are using godotenv directly in this change, maybe an outdated go.mod

go.mod
```diff
@@ -353,7 +354,7 @@ require (
 	github.com/fsnotify/fsnotify v1.8.0 // indirect
 	github.com/fvbommel/sortorder v1.1.0 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
-	github.com/gofrs/flock v0.12.1 // indirect
+	github.com/gofrs/flock v0.12.1
```
neel-astro, line 357: Same here, I don't think flock is being used directly in this change, running `go mod tidy` should fix all these inconsistencies

cloud/env/parse_test.go
```diff
@@ -0,0 +1,85 @@
...
+func (s *Suite) TestParseDotenvFileMissing() {
+	_, err := ParseDotenvFile("/does/not/exist/.env")
+	s.Error(err)
+	s.ErrorContains(err, "opening")
+}
+
+func (s *Suite) TestParseDotenvSkipsCommentsAndBlanks() {
+	dir := s.T().TempDir()
+	path := filepath.Join(dir, ".env")
+	body := []byte("# leading comment\n\nFOO=bar\n  # indented comment treated as blank by godotenv\nBAZ=qux\n")
```
neel-astro, line 49:
```suggestion
	body := []byte("# leading comment\n\nFOO=bar\n  # indented comment treated as blank by gotenv\nBAZ=qux\n")
```
nit

neel-astro (approved):
Left a comment to fix the go.mod file otherwise LGTM, pre-approving

PR #2120, opened 2026-05-04
feat(env): add link / unlink / links commands for env variables

jlaneve:
## Summary

Adds explicit per-deployment linking — including per-link **value overrides** — for workspace-scoped env variables.

Today, the only way to attach a workspace var to specific deployments via the CLI is `--auto-link` (all-or-nothing). To get "shared default with per-deployment differences" you have to use the UI. This PR closes that gap.

Variables only this PR; airflow-variable / connection / metrics-export linking will follow in focused PRs once this shape is validated.

## Mental model

The platform stores overrides **on the link itself**, not as separate objects:

```
WorkspaceVar { value: "default", autoLinkDeployments: bool }
  ├── Link    → Deployment A  [override?: { value: "for-A" }]
  ├── Link    → Deployment B  [no override → uses workspace default]
  └── Exclude → Deployment C  (only meaningful with autoLinkDeployments=true)
```

When a deployment resolves the linked var, the override (if present) replaces the workspace default — but only for that deployment. The workspace value and other deployments are unaffected.

## New commands

```
astro env variable link create --variable-key <key> --workspace-id <ws> --deployment-id <dep> [--value <override>]
astro env variable link create --variable-key <key> --workspace-id <ws> --deployment-id <dep> --exclude
astro env variable link delete --variable-key <key> --workspace-id <ws> --deployment-id <dep> [--exclude]
astro env variable link list   --variable-key <key> --workspace-id <ws> [--format table|json|yaml] [--include-secrets]
```

Following the CLI's `<resource> <action>` convention, the variable (the parent resource) is identified by exactly one of `--variable-id` / `--variable-key` rather than a positional arg.

`--value` and `--exclude` are mutually exclusive (excludes don't carry override values).

### Upsert semantics

`link create` is **idempotent and partial-update**:

- If not yet linked, the link is created (with override if `--value` is passed).
- If already linked, the override is replaced when `--value` is passed.
- Re-running without `--value` is a no-op for an existing link's override (the platform preserves fields omitted *within* a present link entry; omitted Links/ExcludeLinks *arrays* are dropped, which is why every update body round-trips them).
- To remove an existing override: `link delete` then `link create` without `--value`. Two-step, explicit. (Tracked upstream as [AINF-1809](https://linear.app/astronomer/issue/AINF-1809) — the API doesn't currently support "explicit clear" in a single PATCH.)

`link create --exclude` is also idempotent (excluding an already-excluded deployment is a no-op success). `link delete` and `link delete --exclude` keep reject-on-not-present, since removal of something that isn't there is usually a typo.

### 1. Dev / staging / prod from one workspace var

```bash
# Link prod with an override
astro env variable link create --variable-key DATABASE_URL --workspace-id $WS \
  --deployment-id prod-dep --value "postgres://prod-host:5432/app"

# Confirm the link has the override
astro env variable link list --variable-key DATABASE_URL --workspace-id $WS

# Verify what the prod deployment actually sees
astro env variable list --deployment-id prod-dep
# → DATABASE_URL    postgres://prod-host:5432/app
```

### 2. Rotate the override

`link create` is upsert, so just re-run with the new value:

```bash
astro env variable link create --variable-key DATABASE_URL --workspace-id $WS \
  --deployment-id prod-dep --value "postgres://new-host:5432/app"
```

### 3. Shared-by-default, opt out one deployment

```bash
# Var is auto-linked to every deployment
astro env variable create --workspace-id $WS --key LOG_LEVEL --value DEBUG --auto-link

# Exclude the noisy one
astro env variable link create --variable-key LOG_LEVEL --workspace-id $WS \
  --deployment-id low-noise-dep --exclude

# That deployment no longer sees the var
astro env variable list --deployment-id low-noise-dep | grep LOG_LEVEL
# → (no match)
```

### 4. Auto-link with one specific override

Keep auto-link on so every deployment gets the workspace default, but give one deployment a different value.

```bash
astro env variable create --workspace-id $WS --key API_BASE_URL \
  --value "https://api.example.com" --auto-link

astro env variable link create --variable-key API_BASE_URL --workspace-id $WS \
  --deployment-id staging-dep --value "https://staging-api.example.com"

# Every deployment sees the workspace default EXCEPT staging-dep
astro env variable list --deployment-id staging-dep
# → API_BASE_URL    https://staging-api.example.com
```

This is one of the cases the platform investigation explicitly confirmed: explicit `Links` and `autoLinkDeployments=true` coexist, and the explicit link's override wins for that deployment.

### 5. Auditing

```bash
# Human-readable
astro env variable link list --variable-key DATABASE_URL --workspace-id $WS

# Machine-readable for scripts
astro env variable link list --variable-key DATABASE_URL --workspace-id $WS --format json | \
  jq '.links[] | {dep: .deploymentId, override: .overrideValue}'
```

### 6. Cleanup

```bash
# Remove a single explicit link (and its override)
astro env variable link delete --variable-key DATABASE_URL --workspace-id $WS --deployment-id prod-dep

# Remove a single exclude (re-include in auto-link)
astro env variable link delete --variable-key LOG_LEVEL --workspace-id $WS --deployment-id low-noise-dep --exclude

# Clear an override but keep the link: delete the link, then re-create it without --value
astro env variable link delete --variable-key DATABASE_URL --workspace-id $WS --deployment-id prod-dep
astro env variable link create --variable-key DATABASE_URL --workspace-id $WS --deployment-id prod-dep
```

## Wire-level details

For reviewers who want to dig in:

- **`link create` and `link delete`** PATCH `/environment-objects/{id}` with the modified `Links` array. The platform doesn't expose an incremental "add one link" endpoint, so we GET-modify-PATCH.
- **`link create --exclude`** uses the dedicated `POST /environment-objects/{id}/exclude-linking` endpoint (single-shot, incremental).
- **`link delete --exclude`** PATCHes with the trimmed `excludeLinks` array (no incremental endpoint exists for removal).
- **AINF-1792 mitigation**: every PATCH echoes `autoLinkDeployments` back, including plain `update` commands for all four object types (a value-only update previously cleared the flag and dropped links). The platform clears that field when an update body omits it. Stress-tested with 6 mixed link/unlink/exclude/unexclude operations on an `autoLinkDeployments=true` var; the flag survived all 6.
- **Platform requires the typed body**: every PATCH must include `environmentVariable: { value: ... }` even when only mutating links. The CLI re-sends the existing workspace value so this is a no-op for the value while satisfying the type-discriminator check.

## Validation surface

| Input | Result |
|-------|--------|
| `--deployment-id ""` | rejected pre-API: "deployment-id cannot be empty" |
| `--deployment-id "garbage"` | rejected pre-API: "is not a valid deployment ID (expected a CUID)" |
| Both `--value` and `--exclude` | cobra rejects mutual exclusion |
| Linking a deployment-scoped var | rejected: "linking commands require --workspace-id" |
| Linking a var that doesn't exist | "environment object not found: <key>" |
| `link create` on already-linked deployment | upsert (no error) |
| `link create --exclude` on already-excluded deployment | idempotent no-op (no error) |
| Linking a deployment that's already excluded | "has deployment <id> in its exclude list; remove the exclude first" |
| Excluding a deployment that's already explicitly linked | "is explicitly linked to deployment <id>; delete the link first to exclude" |
| `link delete` / `link delete --exclude` on not-present | rejected (typo protection) |
| `--variable-id ""` / `--variable-key ""` | rejected pre-API: "--variable-id or --variable-key cannot be empty" |
| value-only `update` on var with links + auto-link | links, overrides, and auto-link flag all preserved |
| multiline / 100KB override in `link list` table | clamped to one bounded row, layout intact |
| Cross-workspace mismatch | platform "workspace not found" |
| Nonexistent deployment | platform "deployment with id <id> not found" |

## Related upstream issues

- [AINF-1808](https://linear.app/astronomer/issue/AINF-1808) — link API returns 500 instead of 4xx on invalid `scopeEntityId`. Worked around with CLI pre-validation.
- [AINF-1809](https://linear.app/astronomer/issue/AINF-1809) — PATCH should support clearing per-link overrides. Worked around with the delete+create two-step.
- [AINF-2073](https://linear.app/astronomer/issue/AINF-2073) — concurrent link PATCHes can 500 and leave a link unresolved at deployment scope. Found during adversarial smoke testing; rare interleaving, repaired by unlink+relink.
- [AINF-1792](https://linear.app/astronomer/issue/AINF-1792) — `autoLinkDeployments` cleared when omitted from PATCH. Mitigated by echoing the field on every PATCH.

## Verified end-to-end

Real workspace + multiple real deployments, every operation cleaned up after.

| Path | Result |
|------|--------|
| `link create` without override | ✅ |
| `link create --value <override>`, override visible at deployment scope | ✅ |
| Re-run `link create` with new `--value` (upsert replaces override) | ✅ |
| Re-run `link create` without `--value` (no-op for existing override) | ✅ |
| `link create --value ""` (empty-string override) | ✅ |
| Multi-line override (newlines preserved) | ✅ |
| Unicode override | ✅ |
| 5000-char override | ✅ |
| Override with `$`, quotes, backslashes | ✅ exact bytes preserved |
| `link create` to two deployments, distinct overrides, **isolation between them** | ✅ |
| Unlink one deployment of a multi-link, the other survives | ✅ |
| `link create --exclude` on already-excluded deployment | ✅ idempotent |
| Auto-link + exclude → deployment doesn't see var | ✅ |
| Auto-link + explicit `link --value` → override wins at deployment | ✅ |
| Secret var override hidden, `--include-secrets` reveals it | ✅ |
| **AINF-1792**: every PATCH path preserves `autoLinkDeployments=true` | ✅ stress-tested across 6 mixed ops |
| Clear-override workflow: `link delete` + `link create` without `--value` | ✅ |

## What is not in this PR

- `airflow-variable / connection / metrics-export link` — follow-up; each has its own override-flag set.
- Bulk linking via `--from-file` — could chain naturally later (sibling PR #2119 adds bulk import for create/update; same pattern would extend here).
- An explicit `update-link` verb — `link create` is upsert, so this is unnecessary.
- `astro env variable link list` without a variable identifier to audit every workspace var at once.

## Test plan

- [x] `go test ./...` green
- [x] `go vet ./...` clean
- [x] `golangci-lint run ./...` clean
- [x] 11 unit tests in `cloud/env/link_test.go` (link/unlink/exclude/unexclude/list/upsert-override/upsert-preserves-other-links/idempotent-exclude/scope-validation/deployment-id-validation)
- [x] cmd-level tests in `cmd/cloud/env_var_link_test.go` (create/list happy paths, flag validation)
- [x] Five rounds of adversarial binary smoke testing against a real workspace + multiple real deployments
- [x] regression tests: value-only updates round-trip links and auto-link across var / airflow-variable / connection / metrics-export
- [x] adversarial smoke campaign (~40 cases): flag misuse, identity confusion, full link/exclude state machine, hostile payloads (unicode, shell metachars, 100KB/1MB, invalid UTF-8) round-tripping byte-exact, concurrency races, output formats
- [x] AINF-1792 stress test: 6 mixed PATCH operations on an `autoLinkDeployments=true` var, flag survives all 6
- [x] AINF-1808 and AINF-1809 filed for the platform-side gaps surfaced during smoke testing


cloud/env/link.go
```diff
@@ -0,0 +1,380 @@
...
+)
+
+// validateDeploymentID rejects empty or obviously-malformed deployment IDs at
+// the CLI boundary so we surface a clean error instead of "Internal server
+// error" from the platform.
+func validateDeploymentID(depID string) error {
+	if depID == "" {
+		return errors.New("--deployment-id cannot be empty")
+	}
+	if !organization.IsCUID(depID) {
```
neel-astro: Should we move `IsCUID` to utils instead of referencing cross-border from the organization package

cloud/env/link.go
```diff
@@ -0,0 +1,380 @@
...
+		report.ObjectID = *current.Id
+	}
+	if current.AutoLinkDeployments != nil {
+		report.AutoLinkDeployments = *current.AutoLinkDeployments
+	}
+	if current.EnvironmentVariable != nil {
+		report.WorkspaceValue = current.EnvironmentVariable.Value
+		report.IsSecret = current.EnvironmentVariable.IsSecret
+	}
+	if current.Links != nil {
```
neel-astro: I still do not understand why codegen creates pointers to an array :), not a comment for the logic, just a frustration :)

cmd/cloud/env_var_link.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	cmd.Flags().StringVar(&envLinkDeploymentID, "deployment-id", "", "ID of the deployment to unlink (required)")
+	cmd.Flags().BoolVar(&envLinkExclude, "exclude", false, "Remove from excludeLinks instead of links")
+	_ = cmd.MarkFlagRequired("deployment-id")
+	return cmd
+}
+
+func newEnvVarLinksCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "links <id-or-key>",
```
neel-astro: We have been using the format of `<resource> <action>`, keeping the resource as singular throughout the CLI commands with parent resource ID passed as flags instead of args (ex: `astro deployment team list --deployment-id` or `astro deployment token organization-token list --deployment-id`), so I think a more apt version would be `astro env var link list --variable-id` or `astro env var link list --variable-key`

neel-astro: similar comment for link and unlink command, where the variable id or key would be a parent resource identifier, so pass that on as a flags instead of args

neel-astro (commented):
Mostly looks good to me, apart from a comment around command structure

cmd/cloud/env_var_link.go
jlaneve: makes sense, restructured to `astro env variable link create|delete|list` with the variable passed via `--variable-id` or `--variable-key` (exactly one required). same flag treatment for create and delete

cloud/env/link.go
jlaneve: good call, moved it to `pkg/util` (along with its tests) and updated the callers

cloud/env/link.go
jlaneve: haha agreed, the pointer-to-slice ergonomics are rough. comes from oapi-codegen treating optional array fields as nilable

jlaneve:
@neel-astro addressed your feedback, mind taking another look?

- commands restructured to `astro env variable link create|delete|list` with the variable passed via `--variable-id` or `--variable-key`
- `IsCUID` moved to `pkg/util`
- also migrated the new code to `astro-client-v1` after merging main, and consolidated the GET-to-PATCH conversion helpers

cloud/env/link.go
```diff
@@ -0,0 +1,367 @@
...
+		body.EnvironmentVariable = &astrov1.UpdateEnvironmentObjectEnvironmentVariableRequest{
+			Value: &v,
+		}
+	}
+	if links != nil {
+		body.Links = links
+	} else {
+		// Caller didn't touch Links; round-trip the existing ones so the server
+		// doesn't drop them on a partial PATCH.
+		existing := buildUpdateLinks(current.Links)
```
jlaneve: this comment is the tell: the platform drops links omitted from a PATCH. `UpdateVar` in `cloud/env/var.go` doesn't do this round-trip, so `astro env variable update` wipes every deployment link on the var (lost a freshly rotated prod override to this today). the fix belongs there too, ideally as a shared body builder so the next update-shaped function can't repeat it

jlaneve (commented):
**treat this as request-changes** (github won't let me formally request changes on my own PR)

found this the hard way today while rotating tokens in the circus workspace: `astro env variable update` silently wipes deployment links. I updated the workspace value of a var that had a deployment link with a freshly rotated override, and the link (and its override) vanished. the deployment lost the secret entirely

root cause is in `cloud/env/var.go` (pre-dates this PR but this PR is where the evidence lives): `UpdateVar` sends an `UpdateEnvironmentObject` body with only `EnvironmentVariable.Value` + `AutoLinkDeployments` and omits `Links`/`ExcludeLinks`. the platform drops omitted links on a partial PATCH, which `patchVarLinks` in this PR already knows about

changes needed:

1. make `UpdateVar` (and the airflow-variable/connection update siblings) fetch the current object and round-trip `Links`/`ExcludeLinks` exactly like `patchVarLinks` does, ideally by extracting the body-building into a shared helper. a regression test asserting the update body echoes existing links would have caught this
2. `LinkVar`'s doc comment says "the platform's PATCH preserves omitted fields", which contradicts the `patchVarLinks` round-trip comment in the same file. the round-trip comment matches observed platform behavior, so fix the `LinkVar` comment (and the `link create` help text that repeats the claim)

repro from today, stage org, circus workspace: var with a link override on the deployment, run `astro env variable update <key> --workspace-id ... --value ...`, then `astro env variable link list --variable-key <key>` shows `LINKS: (none)` and the deployment no longer resolves the var

jlaneve:
pushed d4e99898 addressing the review above: `echoLinksAndExcludes` round-trips Links/ExcludeLinks in all three update functions (var, airflow var, connection) and `patchVarLinks` now shares it. nil slices normalize to `[]` so the PATCH carries empty arrays, not null. fixed the `LinkVar` doc comment and `link create` help text with the actual nuance: the platform preserves omitted fields *within* a link entry but drops omitted Links/ExcludeLinks *arrays*. regression test asserts a value-only update echoes existing links

verified live against the stage org: created a var with a deployment link override, ran a value-only `update`, link and override both survived (they were wiped before this fix)

cloud/env/airflow_var.go
```diff
@@ -71,10 +72,16 @@ func CreateAirflowVar(scope Scope, key, value string, isSecret bool, autoLink *b
 // does not allow toggling IsSecret. autoLink, when non-nil, toggles the
 // "auto-link to all deployments" flag (workspace scope only).
 func UpdateAirflowVar(idOrKey string, scope Scope, value string, autoLink *bool, astroV1Client astrov1.APIClient) (*astrov1.EnvironmentObject, error) {
-	id, err := resolveID(idOrKey, scope, objectTypeAirflowVar, astroV1Client)
+	// Fetch the full object (not just the ID): the update body must round-trip
```
neel-astro, line 74: good catch, we also need to fix the same in `metricsExport` env obj update type

cloud/env/airflow_var.go
```diff
@@ -85,7 +92,8 @@ func UpdateAirflowVar(idOrKey string, scope Scope, value string, autoLink *bool,
 		},
 		AutoLinkDeployments: autoLink,
 	}
-	resp, err := astroV1Client.UpdateEnvironmentObjectWithResponse(httpcontext.Background(), c.Organization, id, body)
+	echoLinksAndExcludes(&body, current)
+	resp, err := astroV1Client.UpdateEnvironmentObjectWithResponse(httpcontext.Background(), c.Organization, *current.Id, body)
```
neel-astro: curious Q: If `autoLink` is not passed by the user in the command, wouldn't this logic still clear that? Same case for all other env obj type

cloud/env/airflow_var.go
```diff
@@ -85,7 +92,8 @@ func UpdateAirflowVar(idOrKey string, scope Scope, value string, autoLink *bool,
 		},
 		AutoLinkDeployments: autoLink,
 	}
-	resp, err := astroV1Client.UpdateEnvironmentObjectWithResponse(httpcontext.Background(), c.Organization, id, body)
+	echoLinksAndExcludes(&body, current)
```
neel-astro: curious Q: how would the `echoLinkAndExcludes` react to airflow var that are secret, wouldn't the override link value be redacted, since we are fetching with `includeSecrets` set to false, same for other env obj types as well. Also, I think there is a `EnvironmentSecretsShowable` org property, by which customer can turn off allowing `includeSecrets` completely.

cloud/env/format.go
```diff
@@ -276,6 +292,55 @@ func dotenvQuote(v string) string {
...
+	}
+	if override == nil {
+		return "-"
+	}
+	return *override
+}
+
+func writeVarLinksTable(report *VarLinksReport, includeSecrets bool, out io.Writer) {
+	value := report.WorkspaceValue
+	if report.IsSecret {
```
neel-astro: nit: should this also have the `!includeSecrets` check similar to the `overrideDisplay` logic?

neel-astro (commented):
Thanks for all the review changes, and squashing extra bugs. Left few comments mostly around the newly added bug fixes

cloud/env/airflow_var.go
jlaneve: done in 080b2558, `UpdateMetricsExport` now fetches the full object and round-trips links/excludes/auto-link through the same shared helper, with a regression test. verified live on stage: created an export with --auto-link, ran an endpoint-only update, flag survived

cloud/env/airflow_var.go
jlaneve: you were right, and it was live-reproducible: a value-only update flipped `autoLinkDeployments` from true to false (AINF-1792 again). fixed in 080b2558: the shared echo helper now echoes the current flag whenever the caller does not set it, for all four object types. explicit `--auto-link=false` still wins. verified live in both directions

cloud/env/airflow_var.go
jlaneve: great question, I tested this live on stage with a secret var instead of guessing. two platform behaviors make the no-secrets fetch safe here:

1. per-link overrides on secret objects come back *omitted* from the GET (not blanked to ""), so the echoed link entries carry no override field and the platform's per-entry merge preserves the real ones. confirmed: re-ran `link create` without `--value` on a secret var that had an override, the override survived
2. the workspace value comes back as `""` and the platform treats an empty value in a PATCH on a secret as no change, the stored secret survived every link/update op

and since the mutation paths never request secrets at all, `EnvironmentSecretsShowable` being off does not affect them (that org setting only gates reads with `showSecrets=true`). both behaviors are now documented in the code where the echo happens

cloud/env/format.go
jlaneve: yep, fixed: the workspace value now masks only when `--include-secrets` is off, matching the override logic, with a table test covering both directions

jlaneve:
pushed 080b2558 + 1668a6c7 addressing the latest review:

- **auto-link clearing (confirmed bug)**: a value-only `update` was wiping `autoLinkDeployments` because the PATCH omitted it (AINF-1792). the shared echo helper (now `echoPreservedFields`) echoes the current flag whenever the caller does not set it, across var / airflow-variable / connection / metrics-export. explicit `--auto-link=false` still wins
- **metricsExport**: `UpdateMetricsExport` now does the same full-object fetch and round-trips links/excludes/auto-link like the other three types
- **secrets + echo**: verified live on stage with a secret var, see the inline reply for details. tldr: redacted overrides come back omitted (per-entry merge preserves them) and an echoed `""` value on a secret is a no-op, so the no-secrets fetch in mutation paths is safe regardless of the org secrets policy. documented in code
- **format nit**: `link list` workspace value now respects `--include-secrets`

also ran a cleanup pass (1668a6c7): a small `derefSlice` helper replaces the repeated `*[]T` nil-check dance, the build* converters uniformly return non-nil slices so empty lists marshal as `[]` not `null`, `echoPreservedFields` only fills fields the caller did not set (no more build-then-overwrite in `patchVarLinks`), and the nil-ID guard is shared via `objectID`

everything re-verified live against stage (circus workspace, throwaway vars, all cleaned up): auto-link survives value-only updates and every link/exclude/unlink op, secret override + workspace value survive all paths, metrics export auto-link survives endpoint updates. `go test ./...`, `go vet`, `golangci-lint` all green

jlaneve:
ran an adversarial smoke campaign against stage (circus workspace, throwaway vars, everything cleaned up). ~40 cases across flag validation, identity edge cases, link/exclude state transitions, hostile payloads, races, and output formats

**what held up**
- every flag misuse gets a clean error: missing/both identifiers, missing deployment-id, `--value`+`--exclude`, empty/garbage/whitespace deployment IDs
- wrong-entity IDs give clean platform 4xxs: deployment ID as `--variable-id`, workspace ID as `--deployment-id`, well-formed nonexistent CUIDs
- full state machine is correct: upsert, exclude-vs-link conflicts both directions, idempotent re-exclude, reject-on-not-present deletes, delete var with links then recreate same key starts fresh
- override payloads round-trip byte-exact: multiline, tabs, unicode/emoji/RTL, shell metachars, flag-like values (`--exclude` as a value), padded whitespace, JSON-shaped injection attempts, 100KB and even 1MB values
- keys are case-sensitive, aliases (`links ls`/`cr`/`rm`) work, dotenv format correctly rejected for links

**found and fixed (340e5601)**
- `link list` table did not clamp values: a multiline override shredded the table and a 100KB override dumped 100KB to the terminal. now clamped like the var table
- `links`/`excludeLinks` marshaled as `null` when empty, breaking `jq .links[]` (which the PR description itself recommends). now `[]`
- `--variable-id ""` satisfies cobra's one-required group and produced `environment object not found: `. now a clear error

**platform observations (not CLI-fixable)**
- two concurrent `link create` upserts: one got a raw 500, and the surviving state was desynced once: the link existed on the object but the deployment did not resolve the var until I deleted and recreated the link. could not reproduce in 6 follow-up rounds, so it needs the unlucky interleaving. worth filing alongside AINF-1808 (GET-modify-PATCH has no optimistic locking)
- platform accepts 1MB override values without complaint
- invalid UTF-8 in `--value` is silently replaced with U+FFFD by Go json encoding (expected, but means non-UTF-8 secrets get mangled rather than rejected)
- stage rejects deployment-scoped env var creation entirely, so the CLI's "only workspace-scoped objects can be linked" guard is defense in depth

jlaneve:
filed [AINF-2073](https://linear.app/astronomer/issue/AINF-2073) for the concurrent-PATCH desync, linked it in the description alongside AINF-1808/1792

neel-astro (approved):

neel-astro (approved):
Thanks for fixing bunch of new edge cases

PR #2125, opened 2026-05-06
fix(deploy): restore .dockerignore trailing newline after deploy image build

m1racoli:
## Summary

- When building an image without dags, the CLI temporarily appends `dags/` to `.dockerignore` and removes it afterwards
- The removal used `bytes.Trim(..., "\n")` which stripped trailing newlines, leaving the file without its original trailing newline
- This caused conflicts with tools like pre-commit's `end-of-file-fixer`, which expects files to end with a newline

## Fix

`buildImageWithoutDags` now uses a snapshot-and-restore pattern: read the original `.dockerignore` bytes up front, mutate the file for the build, then restore the original bytes byte-for-byte in a single defer. If the file didn't exist before, the defer removes the one we created.

This drops the "detect trailing newline state and reconstruct it" logic entirely — there is nothing to deduce because we have the original bytes in memory.

Side benefits:
- Cross-platform: previously a Windows user's CRLF `.dockerignore` got silently normalized to LF on every dag-deploy build (the bufio scanner stripped `\r\n` and the rewrite emitted `\n`). Snapshot/restore preserves CRLF byte-for-byte.
- `removeDagsFromDockerIgnore` is no longer called from `buildImageWithoutDags`. It is kept as-is (with the PR's trailing-newline-preservation fix applied) for its other caller in the non-dag-deploy build path.

## Test plan

- [x] `TestBuildImageWithoutDagsPreservesDockerignore` — round-trip test verifying `.dockerignore` is byte-for-byte identical after a dag-deploy build. Cases: LF with trailing newline, LF without, **CRLF (Windows)**, file that already contains `dags/`, empty file.
- [x] `TestBuildImageWithoutDagsCleansUpCreatedDockerignore` — when no `.dockerignore` existed beforehand, the one we created is removed after the build.
- [x] `TestBuildImageWithoutDagsAppendsDagsDuringBuild` — uses a mock callback during `Build` to assert the file contains `dags/` *during* the build (not just after).
- [x] `TestRemoveDagsFromDockerIgnore` — retained for the other caller of this function.
- [x] `go test ./cloud/deploy/...` passes.
- [x] `golangci-lint run ./cloud/deploy/...` clean.


jlaneve:
@claude review once

jlaneve (approved):

PR #2132, opened 2026-05-12
Migrate release process to GitHub Actions with RC support and post-release  automation

pritt20:
## Description

 Replaces the hybrid CircleCI + GitHub Actions release pipeline with a fully GitHub Actions-based workflow.

  Release types:
  - Nightly (vX.Y.Z-nightly.YYYYMMDD) — built from main for testing latest development changes, no version needed
  - RC (vX.Y.Z-rc.N) — built from release-X.Y branches with
  auto-incrementing, branch validation, and approval gate
  - GA (vX.Y.Z) — requires at least one RC before release

  Automation:
  - Automates all downstream distribution updates on GA release: godownloader script PR, Homebrew tap PR (via GoReleaser built-in), winget manifest PR, and optional Slack notification
  - Uses Octo STS for short-lived tokens (Homebrew tap, winget fork) instead of long-lived PATs
  - Adds Makefile targets (nightly-tag, rc-tag, release-tag) used by both CI and local workflows

  Other changes:
  - Switches Windows archive format from raw binary to .zip containing astro.exe (required for winget compatibility)
  - Removes CircleCI configuration (.circleci/config.yml, .circleci/next_version.py)

Notion doc:

https://www.notion.so/astronomerio/Astro-CLI-Release-Process-36840290af6c80f9a9e6f5b4c6c60b97#36840290af6c80ea934bea3e1b2aa087

## 🎟 Issue(s)

Related #XXX

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+on:
+  push:
+    branches:
+      - main
+      - 'release-*'
+  pull_request:
+    branches:
+      - main
+      - 'release-*'
```
neel-astro: I think we can open this up to run against all PRs, not necessarily only against main, and release branch, I think there is no reason to restrict it.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+      - 'release-*'
```
neel-astro: I think we can drop this, now that the release process is a separate workflow

.github/workflows/release.yaml
```diff
@@ -44,9 +76,116 @@ jobs:
...
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b "${BRANCH}"
+          git add godownloader.sh
+          git commit -m "chore: update godownloader to ${TAG}"
+          git push origin "${BRANCH}"
+          PR_BODY="Automated PR to update \`LATEST_VERSION\` in \`godownloader.sh\` to \`${TAG}\`."
+          PR_BODY+=$'\n\n'"Triggered by release [${TAG}](https://github.com/${{ github.repository }}/releases/tag/${TAG})."
+          gh pr create \
+            --base main \
+            --title "chore: update godownloader to ${TAG}" \
```
neel-astro:
```suggestion
            --title "CLI Release: update godownloader to ${TAG}" \
```

.github/workflows/release.yaml
```diff
@@ -4,20 +4,52 @@ on:
...
+          fi
+
+          echo "Parsed — Tag: $TAG | Type: $([[ "$TAG" =~ ^nightly- ]] && echo nightly || ([[ "$TAG" =~ -rc\. ]] && echo rc || echo ga))"
+
   release:
+    needs: detect
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/checkout@v4
```
neel-astro: Lets stick to using the SHA instead of version tag everywhere

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
neel-astro: let use SHA instead of version tags everywhere

.github/workflows/release.yaml
```diff
@@ -44,9 +76,116 @@ jobs:
...
+      - uses: goreleaser/goreleaser-action@v6
         with:
           version: v2.15.2
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # For GA releases, use the short-lived Octo STS token scoped to astronomer/homebrew-tap.
+          # For RC/nightly builds, GoReleaser skips homebrew upload via skip_upload: auto
+          # but still evaluates the template, so we pass GITHUB_TOKEN as a safe fallback.
+          HOMEBREW_TAP_TOKEN: ${{ needs.detect.outputs.type == 'ga' && steps.homebrew-token.outputs.token || secrets.GITHUB_TOKEN }}
```
neel-astro: let's drop the secrets.GITHUB_TOKEN completely, don't think there is a need to fallback, otto sts should either work or not, and if not, we should fix it.

.github/workflows/release.yaml
```diff
@@ -44,9 +76,116 @@ jobs:
...
+    steps:
+      - name: Get Winget token via Octo STS
+        uses: octo-sts/action@f603d3be9d8dd9871a265776e625a27b00effe05 # v1.1.1
+        id: winget-token
+        with:
+          scope: astronomer
+          identity: astro-cli-winget
+
+      - uses: vedantmgoyal9/winget-releaser@v2
+        with:
```
neel-astro, line 149: do we also need to pass release-tag? I think we might need it, so the the action can fetch all the required info about the release, since version input strips the `v`

.github/workflows/release.yaml
```diff
@@ -44,9 +76,116 @@ jobs:
...
+          identifier: Astronomer.Astro
+          version: ${{ needs.detect.outputs.version }}
+          installers-regex: '\.zip$'
+          fork: astronomer/winget-pkgs
```
neel-astro: there is `fork-user` field, no `fork` field (https://github.com/vedantmgoyal9/winget-releaser#configuration-options-%EF%B8%8F). I think given we have the fork in the same org, we do not need to specify the field

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+          echo "TAG=$TAG" >> $GITHUB_ENV
+          echo "Computed GA tag: $TAG"
+
+      - name: Create and push tag
+        run: |
+          set -e
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag -a "$TAG" -m "Release $TAG"
+          git push origin "$TAG"
```
neel-astro, line 104: I think we would need to use the otto sts token here to push the tag, if Github default token is used to create PR or tags, that will not trigger the corresponding Github CI runs

Makefile
```diff
@@ -25,3 +25,49 @@ mock:
...
+
+validate-version:
+ifndef VERSION
+	$(error VERSION is required. Usage: make rc-tag VERSION=1.43.0)
+endif
+
+nightly-tag:
+	@set -e; \
+	DATE=$$(date -u +%Y%m%d); \
+	EXISTING=$$(git tag -l "nightly-$$DATE" "nightly-$$DATE.*" 2>/dev/null | wc -l | tr -d ' '); \
```
neel-astro: GoReleaser requires the tag to be a valid semver, so we would probably need to adjust this to be of the form latest minor release version + 1 as the version and then nightly as the pre-release version, i.e. `1.43.0-nightly.YYYY.MM.DD`
Reference - https://goreleaser.com/resources/limitations/semver/

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+          echo "Pushed tag: $TAG"
+          echo "## Release Tag Created" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_STEP_SUMMARY
+          echo "**Tag:** \`$TAG\`" >> $GITHUB_STEP_SUMMARY
+          echo "**Type:** ${{ inputs.type }}" >> $GITHUB_STEP_SUMMARY
+          echo "**Branch:** \`$GITHUB_REF_NAME\`" >> $GITHUB_STEP_SUMMARY
+          echo "**Commit:** \`$(git rev-parse --short HEAD)\`" >> $GITHUB_STEP_SUMMARY
+          echo "**Actor:** @$GITHUB_ACTOR" >> $GITHUB_STEP_SUMMARY
```
neel-astro, line 112: Goreleaser has a pro feature to manage nightly, as part of that it can deletes older nightlies, should we inherit that behaviour as future improvement, so that we do not clog up the github release or tag page with all the nightlies

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Run unit tests
+        run: go test -count=1 -cover -covermode=atomic -test.v ./...
+
+  integration-test:
+    if: github.event_name == 'workflow_dispatch'
```
neel-astro: This workflow can does not support dispatch mode, maybe what you are looking for is an approve step before this one, or simply drop this if condition, both are ok

Makefile
```diff
@@ -25,3 +25,49 @@ mock:
...
+	if [ -z "$$HIGHEST_RC" ]; then \
+		NEXT_RC=1; \
+	else \
+		NEXT_RC=$$((HIGHEST_RC + 1)); \
+	fi; \
+	echo "v$(VERSION)-rc.$$NEXT_RC"
+
+release-tag: validate-version
+	@set -e; \
+	RC_COUNT=$$(git tag -l "v$(VERSION)-rc.*" 2>/dev/null | wc -l | tr -d ' '); \
```
neel-astro, line 75: should we also be validating that the latest rc tag commit is same as from where the release is being cut, otherwise slightly beats the purpose of having this check. Also, maybe in certain condition we would want to bypass this rule, if so we should document on how to do that in our notion docs.

.github/workflows/release.yaml
```diff
@@ -4,20 +4,52 @@ on:
...
+            echo "type=rc" >> $GITHUB_OUTPUT
+            VERSION=$(echo "$TAG" | sed 's/^v//; s/-rc\.[0-9]*$//')
+            echo "version=$VERSION" >> $GITHUB_OUTPUT
+          else
+            echo "type=ga" >> $GITHUB_OUTPUT
+            VERSION=$(echo "$TAG" | sed 's/^v//')
+            echo "version=$VERSION" >> $GITHUB_OUTPUT
+          fi
+
+          echo "Parsed — Tag: $TAG | Type: $([[ "$TAG" =~ ^nightly- ]] && echo nightly || ([[ "$TAG" =~ -rc\. ]] && echo rc || echo ga))"
```
neel-astro: `type` can be figure out in the if else block above and simply be logged here like `tag` to avoid the duplicate regex evaluation

.goreleaser.yml
```diff
@@ -27,58 +25,67 @@ builds:
...
       owner: astronomer
       name: homebrew-tap
+      token: "{{ .Env.HOMEBREW_TAP_TOKEN }}"
+      pull_request:
+        enabled: true
+        base:
+          owner: astronomer
+          name: homebrew-tap
+          branch: main
     directory: Formula
```
neel-astro, line 86: I think we would have to also add a separate almost duplicate entry where it would only differ in the name of the brew to preserve the logic for maintaining both latest and versioned taps:
```
brews:
  - name: astro            # writes Formula/astro.rb
    repository: { ... }
    directory: Formula
    skip_upload: auto
    # ...

  - name: "astro@{{ .Version }}"  # writes Formula/astro@1.43.0.rb
    repository: { ... }
    directory: Formula
    skip_upload: auto
    # ...
```

neel-astro (commented):
Had a first pass and left few comments to squash

.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,112 @@
...
+  create-tag:
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.type != 'nightly' && 'release' || '' }}
+    steps:
+      - name: Get token via Octo STS
+        uses: octo-sts/action@f603d3be9d8dd9871a265776e625a27b00effe05 # v1.1.1
+        id: sts-token
+        with:
+          scope: astronomer
+          identity: astro-cli-release
```
neel-astro, line 34: you will need to add this to the chainguard - https://github.com/astronomer/.github/tree/main/.github/chainguard

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: go.mod
+
+      - name: Run unit tests
+        run: go test -count=1 -cover -covermode=atomic -test.v ./...
+
+  integration-test:
+    if: github.ref == 'refs/heads/main'
```
neel-astro: should we run this on the PR as well? maybe optional so that users can trigger if need be, or it can be on the basis of changes going in the PR, i.e. if relevant folder have been updated then run the tests

neel-astro (approved):
LGTM, left two tiny comments, but pre-approving

.github/workflows/create-release.yaml
pritt20: Thank you!

Raised a PR https://github.com/astronomer/.github/pull/69

.github/workflows/create-release.yaml
schnie: Was also thinking about this - tons of nightly tags filling up the pages. Will that happen or are we avoiding that somehow in here?

pritt20: Yes, This is already handled. The `cleanup-nightlies` job runs after every nightly release and keeps only the 5 most recent nightlies. Older ones are deleted (both the GitHub release and the git tag).

ref: https://github.com/astronomer/astro-cli/blob/454c868d6950877ffcd92a6ec4a73a7bd142e81b/.github/workflows/release.yaml#L156

schnie (approved):

PR #2158, opened 2026-05-28
feat(scaffold): ignore .astro/worktrees/ in default .gitignore

schnie:
## Summary
- Add `.astro/worktrees/` to the scaffold's `.gitignore` template (parallels the existing `.astro/standalone/` entry)
- Update scaffold test to assert the new entry is present

## Motivation
astro-desktop is moving per-branch git worktrees from a shared home-dir location (`~/.astro-desktop/worktrees/<project>/`) into the project itself at `<project>/.astro/worktrees/`. The otto CLI uses the same path so both tools see the same worktrees. Without this template change, those nested worktrees would show up as untracked in new projects.

For existing projects astro-desktop writes the entry to `.git/info/exclude` instead (untracked, per-clone), so this PR only affects projects newly scaffolded by `astro dev init`.

## Test plan
- [x] `go test ./...` in `pkg/airflowrt/`
- [ ] Manual: `astro dev init` → confirm `.gitignore` contains the new entry


akshayarora (approved):
This is great, thank you.

jlaneve (approved):

PR #2169, opened 2026-06-12
Add pkg/container and converge airflow/runtimes onto it

schnie:
## What

Introduces `github.com/astronomer/astro-cli/pkg/container` — an importable sub-module (its own go.mod, testify-only) that owns container-engine **resolution**, the **Podman astro-machine lifecycle**, Docker **auto-start**, and the **connection environment**, with config + user-feedback **injected** (no `config.CFG` singleton, no CLI spinner). Then **converges `airflow/runtimes` onto it**: that package is now a thin adapter that injects `config.CFG` + the CLI spinner and delegates to `pkg/container`.

Two commits:
1. Add `pkg/container` (the DI'd implementation + tests + mocks).
2. Converge `airflow/runtimes` → thin adapter; remove the duplicated logic.

## Why

astro-desktop needs to drive whatever container engine a project runs under (logs, container shell, deploy) without hardcoding `docker`. Rather than re-implement engine detection + the Podman machine/socket wiring (which would drift from the CLI), the desktop imports `pkg/container`. Putting the shared code in a module the **CLI also consumes** means one implementation, exercised by CI, with no duplication — instead of an orphan helper.

## Scope / safety

- CLI behavior is unchanged: `airflow/runtimes`' public surface (`ContainerRuntime`, `GetContainerRuntime`, `GetContainerRuntimeBinary`, `IsPodman`, `CreateOSChecker`) and all call sites are untouched; the adapter preserves engine-resolution order, the `astro-machine` lifecycle, the `DOCKER_HOST`/`CONTAINER_HOST` + default-connection wiring, and Docker auto-start.

`go build ./...` + container tests pass; the moved runtime tests live in `pkg/container`.


pkg/container/manager.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		if len(containers) > 0 {
+			return errors.New(machineAlreadyRunningMsg)
+		}
+		if err := m.podman.StopMachine(other); err != nil {
+			return err
+		}
+	}
+
+	state, err := m.MachineState()
```
pritt20, line 123: Hey @schnie ,

EnsureMachine brings the astro-machine up but never configures DOCKER_HOST/CONTAINER_HOST for the running process. The helper that does that configureMachineForUsage exists, but is only reachable from Configure()/ConfigureOrKill(), never from the Initialize → EnsureMachine path and hence when trying to spin up astro local project with podman as container engine fails with below error:

```
Error: error building, (re)creating or starting project containers: Error response from daemon: pull access denied for af-3_4e4d2f/airflow, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Once I configured required host manually, it started working fine:

```
export DOCKER_HOST=unix:///var/folders/1l/kzxfwckj6zlbnxvy2h75b4yr0000gn/T/podman/astro-machine-api.sock
export CONTAINER_HOST=unix:///var/folders/1l/kzxfwckj6zlbnxvy2h75b4yr0000gn/T/podman/astro-machine-api.sock
```

Previously we had similar calls under `ensureMachine` in podman_runtime.go, where we would check if machine is pre-existing or do we need to create Astro machine and set then configure the `DOCKER_HOST` for podman to make these calls. Ref:

https://github.com/astronomer/astro-cli/blob/d36818ff14c593c649aa1b9ed977454a707cd853/airflow/runtimes/podman_runtime.go#L170
https://github.com/astronomer/astro-cli/blob/d36818ff14c593c649aa1b9ed977454a707cd853/airflow/runtimes/podman_runtime.go#L180
https://github.com/astronomer/astro-cli/blob/d36818ff14c593c649aa1b9ed977454a707cd853/airflow/runtimes/podman_runtime.go#L193

pritt20 (commented):
Overall changes look good. Left one comment to address to configure `DOCKER_HOST` for podman container engine.

pritt20 (approved):
Thanks for Addressing the podman issue. Changes look good to me. 👍

PR #2179, opened 2026-06-18
feat(dev): persist build secrets via config to avoid repeating --build-secrets flags

collinmcnulty:
> Disclaimer: I do not know Go and relied heavily on Claude for this. I think it's a small enough change that it's conceivable this just works, but please treat this more like a glorified issue than a vetted solution.

## Summary

PR [astronomer/astro-runtime#2974](https://github.com/astronomer/astro-runtime/pull/2974) added support for a `netrc` Docker build secret to enable private pip installs. However, users must pass `--build-secrets id=netrc,env=NETRC_CONTENT` on every `astro dev` invocation (`start`, `restart`, `parse`, `pytest`, `build`, `upgrade-test`).

This PR adds a `dev.build_secrets` config key so the secret can be declared once and picked up automatically:

```bash
# One-time setup (project-scoped):
astro config set dev.build_secrets "id=netrc,env=NETRC_CONTENT"

# Or globally:
astro config set --global dev.build_secrets "id=netrc,env=NETRC_CONTENT"

# Then all dev commands just work without extra flags:
astro dev start
astro dev pytest
astro dev parse
```

**Fallback priority order:**
1. `--build-secrets` flag (explicit, highest priority)
2. `dev.build_secrets` in `.astro/config.yaml` (project config)
3. `dev.build_secrets` in `~/.astro/config.yaml` (global config)
4. `BUILD_SECRET_INPUT` env var (existing undocumented fallback, preserved)

Cloud deploy commands (`astro deploy`, `astro remote deploy`) intentionally do **not** read this config key — secrets for cloud builds should remain explicit.

## Changes

- `config/types.go` — add `DevBuildSecrets cfg` field to the `cfgs` struct
- `config/config.go` — register `dev.build_secrets` config key with empty default
- `pkg/util/util.go` — add variadic `fallbacks ...string` to `GetbuildSecretString` (backwards-compatible; existing callers without a fallback are unaffected)
- `cmd/airflow.go` — pass `config.CFG.DevBuildSecrets.GetString()` as fallback at all 6 `astro dev` call sites

## Test plan

- [ ] `go test ./config/... ./pkg/util/... ./cmd/...` passes (verified locally)
- [ ] `astro config set dev.build_secrets "id=netrc,env=NETRC_CONTENT"` persists to `.astro/config.yaml`
- [ ] `astro dev start` without `--build-secrets` picks up the config value
- [ ] Explicit `--build-secrets` flag overrides the config value
- [ ] `astro deploy` does not pick up the dev config value

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -810,7 +810,7 @@ func airflowUpgradeTest(cmd *cobra.Command, astroV1Client astrov1.APIClient) err
 		fmt.Printf("failed to add 'upgrade-test*' to .gitignore: %s", err.Error())
 	}
 
-	buildSecretString = util.GetbuildSecretString(buildSecrets)
+	buildSecretString = util.GetbuildSecretString(buildSecrets, config.CFG.DevBuildSecrets.GetString())
 
 	err = containerHandler.UpgradeTest(runtimeVersion, deploymentID, customImageName, buildSecretString, versionTest, dagTest, lintTest, lintDeprecations, lintFix, lintConfigFile, astroV1Client)
 	if err != nil {
@@ -845,7 +845,7 @@ func airflowStart(cmd *cobra.Command, args []string, astroV1Client astrov1.APICl
 		return err
 	}
 
-	buildSecretString = util.GetbuildSecretString(buildSecrets)
+	buildSecretString = util.GetbuildSecretString(buildSecrets, config.CFG.DevBuildSecrets.GetString())
 
 	return containerHandler.Start(&airflow.StartOptions{
 		ImageName:         customImageName,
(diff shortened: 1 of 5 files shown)


jlaneve:
hey collin, I pushed directly to this branch (force-push) to pull out the last commit ("Allow passing workspace or deployment through config.yml") so this PR is now just the `dev.build_secrets` change. that part looks good and I verified build + `go test ./cmd/... ./config/... ./pkg/util/...` all pass

context on why I pulled the workspace/deployment bit out for now:

- it read the new `dev.workspace_id` / `dev.deployment_id` keys via `GetString()`, which falls back to the global `~/.astro/config.yaml`. since a user can have multiple projects in different workspaces, a globally-set value would bleed across every project
- it also made `astro dev start`/`restart` call the Environment Manager API whenever those IDs are set, and that path requires a logged-in cloud context. so a stray global value would make `astro dev start` fail for anyone not logged in, where today it runs fine offline

if we want to bring it back, I think the shape is roughly:

- resolve workspace/deployment from *project* config only (reuse the existing `project.deployment` key, read via `GetProjectString()` so it never falls back to home) instead of new `dev.*` keys. `project.deployment` is already written by `astro deploy --save`, so it'd mostly just work
- make the connection sync best-effort: if the user isn't logged in or the API errors, warn and keep starting airflow instead of hard-failing

the `dev.build_secrets` feature is unaffected by all of that and good to go. thanks for putting this up

jlaneve (approved):
lgtm, build-secrets config reads cleanly and tests pass. thanks collin

Issue #2181, opened 2026-06-18
--workspace-id ignored on deployment commands: coalesceWorkspace() checks current-workspace context before reading the flag


jremigio-circle:
## Summary

With an organization-scoped API token (no `workspaceId` claim) used via `ASTRO_API_TOKEN` in a non-interactive/CI context, `astro deployment variable update` (and other commands routing through `coalesceWorkspace`) fail with:

```
Error: failed to find a valid workspace: failed to get current Workspace: current workspace context not set
```

**EVEN WHEN** `--workspace-id <id>` is passed. The explicit flag is effectively ignored.

## Affected versions

v1.40.1, v1.42.1 (latest stable), and v1.43.0-nightly — the relevant functions are byte-identical across these, so upgrading does not help.

## Root cause (code)

In `cmd/cloud/workspace.go`, `coalesceWorkspace()` calls `workspace.GetCurrentWorkspace()` and returns its error BEFORE it checks the `--workspace-id` flag:

```go
func coalesceWorkspace() (string, error) {
    wsFlag := workspaceID
    wsCfg, err := workspace.GetCurrentWorkspace()   // called first
    if err != nil {
        return "", errors.Wrap(err, "failed to get current Workspace")  // returns here — flag never read
    }
    if wsFlag != "" { return wsFlag, nil }           // --workspace-id only honored after
    if wsCfg != "" { return wsCfg, nil }
    return "", errors.New("no valid Workspace source found")
}
```

`GetCurrentWorkspace()` (`cloud/workspace/workspace.go`) errors whenever `c.Workspace == ""`. In the API-token path, `checkAPIToken()` (`cmd/cloud/setup.go`) derives the workspace from token claims, falling back to `c.Workspace` (`if wsID == "" { wsID = c.Workspace }`); an org-scoped token has no `workspaceId` claim, so in a fresh config home `c.Workspace` stays empty and the flag cannot rescue it.

## Reproduction

1. `ASTRO_HOME=$(mktemp -d)`, set `ASTRO_API_TOKEN` to an organization-scoped token, no `astro workspace switch` performed.
2. Run `astro deployment variable update --workspace-id <valid-ws-id> --deployment-id <valid-dep-id> --key FOO --value bar`.
3. Observe the `current workspace context not set` error despite `--workspace-id`.
4. Running `astro workspace switch <ws-id>` first makes it succeed.

## Expected

An explicit `--workspace-id` should be honored before requiring a current-workspace context.

## Suggested fix

[...]

PR #2182, opened 2026-06-18
fix(workspace): honor explicit --workspace-id before requiring current workspace context

jlaneve:
## Problem

When authenticating non-interactively with an **organization-scoped** (workspace-operator) API token via `ASTRO_API_TOKEN`, commands like `astro deployment variable update` fail with:

```
Error: failed to find a valid workspace: failed to get current Workspace: current workspace context not set,
you can switch to a workspace with
	astro workspace switch WORKSPACEID
```

Passing `--workspace-id <id>` does **not** prevent the error — the flag is silently ignored on this path.

### Root cause

`coalesceWorkspace()` called `workspace.GetCurrentWorkspace()` first and returned its error *before* ever consulting the `--workspace-id` flag:

```go
wsFlag := workspaceID
wsCfg, err := workspace.GetCurrentWorkspace()   // called first
if err != nil {
    return "", errors.Wrap(err, "failed to get current Workspace")  // returns here — flag never consulted
}
if wsFlag != "" { return wsFlag, nil }           // flag only honored after
```

An org-scoped token has no `workspaceId` claim, so in a fresh CI home `checkAPIToken` leaves `c.Workspace` empty. `GetCurrentWorkspace()` then errors, and because the flag is checked *after*, `--workspace-id` cannot rescue it. This is deterministic given an empty current-workspace context (it only looks intermittent because a prior `astro workspace switch` may have persisted a workspace).

## Fix

Return an explicit `--workspace-id` **before** calling `GetCurrentWorkspace()`, making the flag authoritative in non-interactive flows. This covers every command routed through `coalesceWorkspace` (deployment variable update, deployment objects, worker queues, deploy, dbt, env, inspect) — not just `variable update`.

## Tests

Adds `TestCoalesceWorkspace` covering:
- explicit flag with no current context (**the regression** — fails before this change)
- explicit flag precedence over an existing context
- fallback to the current workspace context when no flag is provided
- error when neither source is available

## Workaround (pre-fix)

`astro workspace switch <id>` first persists `c.Workspace`, after which the command succeeds. Confirmed still valid on affected versions.

### DIFF
--- a/cmd/cloud/workspace.go
+++ b/cmd/cloud/workspace.go
@@ -844,14 +844,17 @@ func addOrgTokenToWorkspace(cmd *cobra.Command, args []string, out io.Writer) er
 }
 
 func coalesceWorkspace() (string, error) {
-	wsFlag := workspaceID
+	// An explicit --workspace-id flag is authoritative and must win before we
+	// require a current-workspace context. Org-scoped API tokens leave that
+	// context empty (no workspaceId claim), so consulting it first would error
+	// out even when the caller supplied a valid workspace explicitly.
+	if wsFlag := workspaceID; wsFlag != "" {
+		return wsFlag, nil
+	}
+
 	wsCfg, err := workspace.GetCurrentWorkspace()
 	if err != nil {
 		return "", errors.Wrap(err, "failed to get current Workspace")
-	}
-
-	if wsFlag != "" {
-		return wsFlag, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pritt20 (approved):
Looks good to me!

sakethsomaraju:
Closes https://github.com/astronomer/astro-cli/issues/2181

PR #2198, opened 2026-07-08
Expose non-DAG bundle deploys and bundle management

jeremybeard:
## Description

Generalizes non-DAG bundle deploys beyond dbt and adds commands to manage the bundles registered on a Deployment.

- `astro deploy --non-dags` deploys any non-DAG bundle from a directory separate from the Astro project, driven by `--non-dags-mount-path`, `--non-dags-bundle-type` (free-form label, defaults to `none`), and `--non-dags-local-path`. It rides the stable v1 `BUNDLE` deploy surface — the same path `astro dbt deploy` already uses, which stays as a dedicated special case.
- `astro deploy --dag-bundle-name` targets a named DAG bundle. Because `dagBundleName` exists only on the v1alpha1 deploy API, the create call routes through the v1alpha1 client while upload and finalize stay on v1.
- `astro deployment bundle create/list/update/delete` manages bundle registrations on a Deployment. `list` pages through all results and supports `--json`, `--output`, and `--template`.
- `astro deploy --help` groups its flags into sections by deploy type.

All of the new flags and the `astro deployment bundle` command group are **hidden** while the bundle features are in preview. They work when invoked but do not appear in `--help` output, so the surface can be validated end to end before being announced. Un-hiding is a one-line change per flag/command once the features graduate from preview.

## 🧪 Functional Testing

Verified locally end to end:

- `astro deploy <id> --dags` — DAG deploy to the default `main` bundle.
- `astro deploy <id> --non-dags --non-dags-mount-path /usr/local/airflow/dbt/jaffle_shop --non-dags-bundle-type dbt` — non-DAG bundle deploy to a distinct blob.
- `astro deploy <id> --dag-bundle-name finance` — DAG deploy to a named DAG bundle.
- `astro deployment bundle create/list/update/delete` against a Deployment, including `--json` output.

Automated: unit tests cover the `--dag-bundle-name` create path (v1alpha1 type mapping and git mapping), the `--non-dags` command flow, and the bundle CRUD service functions.

## 📋 Checklist


cmd/cloud/deploy_help.go
```diff
@@ -0,0 +1,85 @@
...
+Non-DAG Bundle Flags:
+{{filterDeployFlags .LocalFlags "non-dags" | trimTrailingWhitespaces}}{{end}}{{end}}{{if .HasAvailableInheritedFlags}}
+
+Global Flags:
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}
+`
+
+// annotateDeployFlag sets a group annotation on a flag for grouped help display.
+func annotateDeployFlag(cmd *cobra.Command, name, group string) {
+	cmd.Flags().SetAnnotation(name, "group", []string{group}) //nolint:errcheck
```
neel-astro, line 84: love the annotations ❤️

cmd/cloud/deployment_bundle.go
```diff
@@ -0,0 +1,138 @@
...
+	return cmd
+}
+
+func newDeploymentBundleUpdateCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update BUNDLE-ID",
+		Short: "Update a bundle on an Astro Deployment",
+		Long:  "Update a bundle's description or, for a non-DAG bundle, the DAG bundles it is served alongside.",
+		Example: `  # Update a bundle's description
+  astro deployment bundle update <bundle-id> --deployment-id <id> --description "my bundle"
```
neel-astro, line 100: should we simplify the discoverability/use of bundle command by supporting `--bundle-name <name>` flag to make it easy to consume. As of now it seems tiny bit difficult to write a script around bundles, because bundle id is only present in output that would need to be parsed in create/list operation.

neel-astro (approved):
Left a tiny comment, but otherwise LGTM 🚀

cmd/cloud/deployment_bundle.go
jeremybeard: Good call, have added `--name` and `--mount-path` flags for this

jeremybeard:
Merging despite integration test failure which is a separate issue we will soon be addressing

PR #2202, opened 2026-07-10
Automatically cut nightly releases on a schedule

jeremybeard:
## Summary
- Adds a `schedule` trigger (daily, 07:00 UTC) to `Create Release Tag` so nightlies get cut automatically instead of requiring a manual `workflow_dispatch`.
- Scheduled runs default `type` to `nightly` throughout the workflow since the `inputs` context is unset for non-`workflow_dispatch` events (`inputs.type || 'nightly'`).
- Manual dispatch (nightly/rc/ga) is unchanged.

## Test plan
- [ ] Confirm the scheduled trigger fires from `main` and produces a valid nightly tag/release once merged
- [ ] Manually re-run `workflow_dispatch` with `type: nightly`/`rc`/`ga` to confirm no regression in the existing manual path

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,8 +1,11 @@
 name: Create Release Tag
 
-run-name: "${{ inputs.type }} release${{ inputs.version && format(' v{0}', inputs.version) || '' }} from ${{ github.ref_name }} (by @${{ github.actor }})"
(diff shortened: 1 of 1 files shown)


jlaneve (approved):
Scheduling nightlies before the release build optimization lands keeps the rollout focused. I verified the scheduled/manual type handling and will rebase the follow-up performance PR on this.

PR #2205, opened 2026-07-10
Expand `astro api`: full `describe` fields + `--json` for describe & ls

jeremybeard:
## Description

Improves `astro api <cloud|airflow|registry>` for both humans and agents/scripts.

### 1. `describe` now expands `$ref` schemas down the full stack

Previously `describe` stopped at the first `$ref`, printing only the schema name with no fields:

```
Request Body (required):
  Schema: CreateDeploymentRequest
```

The recursion was severed at the conversion layer (`convertSchemaProxy` stored a `$ref` name but never resolved it), so the (already cycle-safe) printer never received the referenced fields.

This resolves refs lazily via a component-schema registry:

- `pkg/openapi`: `ExtractComponentSchemas` (v3 `components.schemas`) / `ExtractDefinitionsV2` (v2 `definitions`) build a `name -> schema` registry, exposed via `Cache.GetSchemas()`.
- `SchemaResolver` gains `NewSchemaResolverWithSchemas`; `ResolveSchema` returns the registered schema when a `SchemaRef` carries only a ref name.
- The `describe` printer resolves through the resolver, expanding nested `$ref`s and array-of-`$ref` element schemas. Genuine cycles print `(see X above)`.

After:

```
Request Body (required):
  Schema: CreateDeploymentRequest
  executor  Executor
      type  string
  isDagDeployEnabled  boolean
  name*  string
      Deployment name
  workerQueues  array of WorkerQueue
      name  string
```

### 2. `--json` output for `describe` and `ls`

For programmatic use by agents and scripts, both subcommands gain a `--json` flag (all three API families).

- `describe --json` emits the endpoint with every `$ref` resolved inline: ordered `properties`, `readOnly` fields included and flagged, real cycles marked `"circular": true` (siblings still expand), and a depth backstop marks `"truncated": true` on pathologically deep schemas. A single match is a JSON object; a path matching multiple methods is an array.
- `ls --json` emits a lean array of endpoint summaries (`method`, `path`, `operationId`, `summary`, `tags`, `deprecated`, `pathParameters`) — always an array.
- No human-readable text (banners, trailers, verbose spec-URL, the Airflow version line) leaks into any `--json` output.

### 3. Review-driven correctness fixes

An adversarial review of the change surfaced (and this PR fixes):

- **`allOf` + sibling `properties`**: a schema that extends a base *and* adds its own fields now prints both in text output, instead of stopping after `allOf`.
- **Unresolved `$ref` property**: a property pointing at a schema absent from the registry now still shows its name and referenced schema name, instead of being silently dropped.
- **JSON depth backstop**: bounds recursion for deep/wide acyclic schemas (cycles were already cut).

## Issue(s)

Related #XXX

## Functional Testing

```
# Human-readable, now expands referenced schemas down the stack
astro api cloud describe CreateDeployment

# Machine-readable, refs resolved inline
astro api cloud describe CreateDeployment --json | jq '.requestBody.schema.properties'

# List endpoints as JSON
astro api cloud ls deployments --json | jq '.[].operationId'
```

## Checklist

### DIFF
--- a/cmd/api/airflow.go
+++ b/cmd/api/airflow.go
@@ -578,6 +578,7 @@ func NewAirflowListCmd(out io.Writer, parentOpts *AirflowOptions) *cobra.Command
 	var filter string
 	var verbose bool
 	var refresh bool
+	var jsonOut bool
 
 	cmd := &cobra.Command{
 		Use:     "ls [filter]",
@@ -615,7 +616,9 @@ The filter matches against endpoint paths, methods, operation IDs, summaries, an
 				return err
 			}
 
-			fmt.Fprintf(out, "Airflow version: %s\n\n", parentOpts.detectedVersion)
+			if !jsonOut {
+				fmt.Fprintf(out, "Airflow version: %s\n\n", parentOpts.detectedVersion)
+			}
 
 			// Create list options
 			listOpts := &ListOptions{
@@ -624,6 +627,7 @@ The filter matches against endpoint paths, methods, operation IDs, summaries, an
 				Filter:    filter,
 				Verbose:   verbose,
 				Refresh:   refresh,
+				JSON:      jsonOut,
 			}
 
 			return runList(listOpts)
@@ -632,6 +636,7 @@ The filter matches against endpoint paths, methods, operation IDs, summaries, an
 
 	cmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "Show additional details like summaries and tags")
 	cmd.Flags().BoolVar(&refresh, "refresh", false, "Force refresh of the OpenAPI specification cache")
+	cmd.Flags().BoolVar(&jsonOut, "json", false, "Output the endpoint list as JSON for programmatic use")
 
 	return cmd
 }
--- a/cmd/api/airflow.go
+++ b/cmd/api/airflow.go
@@ -641,6 +641,7 @@ func NewAirflowDescribeCmd(out io.Writer, parentOpts *AirflowOptions) *cobra.Com
 	var method string
 	var refresh bool
 	var verbose bool
+	var jsonOut bool
 
 	cmd := &cobra.Command{
 		Use:   "describe <endpoint>",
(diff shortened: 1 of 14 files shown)


akshayarora (approved):
This looks good, thanks!

jlaneve (approved):
Thanks Jeremy — I validated the schema expansion against the live Astro Cloud, Registry, and representative Airflow specs.

I pushed a small follow-up that:
- expands top-level array response schemas in text output; this affects three current Cloud endpoints;
- treats visited refs as ancestry, so reused sibling schemas expand normally while genuine cycles remain bounded;
- consolidates nested traversal, preventing duplicate output for schemas with properties and composition;
- makes describe --json consistently return an array, avoiding a cardinality-dependent API contract.

Added focused regression tests for each case. Local package tests and lint pass, and CI is green on commit 57e5d77c.

The remaining concerns—component aliases, external refs, and pathological expansion budgets—are not exercised by current first-party specs and can be follow-up hardening.

jlaneve:
Follow-up with concrete before/after examples from the shipped specs:

### Top-level array responses

Before, Cloud endpoints such as `ListPermissionGroups` stopped at the response status:

```text
Responses:
  200: OK
```

After, array item fields are expanded:

```text
Responses:
  200: OK
    description*  string
    name*  string
    permissions*  array of PermissionEntry
    scope*  string
```

### Nullable schemas

Before, nullable Airflow and Registry fields looked like objects with empty branches:

```text
next_cursor  object
    Any of the following:

    Option 1:

    Option 2:
```

After, primitive unions are compact:

```text
next_cursor  string or null
duration*  number or null
state*  TaskInstanceState or null
```

Nullable object branches remain expanded, with every option labeled:

```text
dag_version*  DagVersionResponse or null
    Any of the following:

    Option 1: DagVersionResponse
      id*  string (uuid)
      version_number*  integer

    Option 2: null
```

### Stable JSON shape

Before, a single `describe --json` match was an object while ambiguous paths returned arrays. After, both use a stable array:

```json
[
  {
    "method": "GET",
    "operationId": "get_task_instances"
  }
]
```

These changes are in `57e5d77c` and `1a054820`. Focused regressions, package tests, race tests, lint, and the 641-endpoint shipped-spec sweep pass locally.

PR #2210, opened 2026-07-17
fix(cli): paginate org/workspace/deployment lists past the first page

AliceMur:
## Description

The CLI only fetched the first page of organizations, workspaces, and deployments, so `astro organization switch` and `astro workspace switch` failed with "could not be found" when the target wasn't on the first page.

`ListOrganizations`, `GetWorkspaces`, and `ListDeployments` now page through all results; switch-by-name returns as soon as the match is found. Follows the existing `cloud/env` pagination pattern (page-count safety cap, offset advanced by rows returned, stop once the total is reached).

## 🧪 Functional Testing

With more orgs/workspaces than a single API page, run `astro organization switch <name/id>` and `astro workspace switch <name/id>` against a target that isn't on the first page — the switch now succeeds instead of erroring. Covered by unit tests for multi-page retrieval, an end-to-end switch to a later-page target, and an HTTP-level test driving a real client against a stub API to confirm correct `offset`/`limit` query params and multi-page aggregation.

## 📸 Screenshots

N/A

## 📋 Checklist

### DIFF
--- a/cloud/deployment/deployment.go
+++ b/cloud/deployment/deployment.go
@@ -17,6 +17,7 @@ import (
 	airflowversions "github.com/astronomer/astro-cli/airflow_versions"
 	"github.com/astronomer/astro-cli/astro-client-v1"
 	"github.com/astronomer/astro-cli/cloud/organization"
+	"github.com/astronomer/astro-cli/cloud/pagination"
 	"github.com/astronomer/astro-cli/cloud/workspace"
 	"github.com/astronomer/astro-cli/config"
 	"github.com/astronomer/astro-cli/pkg/ansi"
@@ -70,10 +71,6 @@ var (
 	listLimit        = 1000
 	dagDeployEnabled bool
 )
-
-// maxListPages caps a single paginated list call at maxListPages pages as a safety
-// bound against a server that mis-reports TotalCount (mirrors cloud/env).
-const maxListPages = 100
 
 // defaultWorkerMachineName is the machine type UpdateDeployment falls back to when
 // an executor change forces it to synthesize a replacement worker queue.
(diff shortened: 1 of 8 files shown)


cloud/deployment/deployment.go
```diff
@@ -71,6 +71,10 @@ var (
 	dagDeployEnabled bool
 )
 
+// maxListPages caps a single paginated list call at maxListPages pages as a safety
+// bound against a server that mis-reports TotalCount (mirrors cloud/env).
+const maxListPages = 100
```
neel-astro: nit: should we just scope this constant to `cloud` package itself, so that we do not need to redeclare it in all sub packages

cloud/organization/organization.go
```diff
@@ -75,11 +90,13 @@ func GetOrganization(orgID string, astroV1Client astrov1.APIClient) (*astrov1.Or
 	return resp.JSON200, nil
 }
 
+// findOrganizationByName pages through Organizations and returns as soon as it
+// finds a name match, so a match on an early page avoids fetching later pages.
 func findOrganizationByName(name string, astroV1Client astrov1.APIClient) (*astrov1.Organization, error) {
```
neel-astro, line 86: nit: the `findOrganizationByName` and `ListOrganizations` method logic seems almost identical, with the only difference being the early exit in `findOrganizationByName`, should we DRY them up?

cloud/workspace/workspace_test.go
```diff
@@ -70,6 +70,38 @@ func (s *Suite) TestList() {
...
+			Workspaces: []astrov1.Workspace{{Name: "ws-page2", Id: "ws-2"}},
+			TotalCount: 2,
+			Limit:      1000,
+			Offset:     1000,
+		},
+	}
+	mockClient := new(astrov1_mocks.ClientWithResponsesInterface)
+	// First call returns page 1 (offset 0), second call returns page 2 (offset 1000).
+	mockClient.On("ListWorkspacesWithResponse", mock.Anything, mock.Anything, mock.Anything).Return(&page1, nil).Once()
+	mockClient.On("ListWorkspacesWithResponse", mock.Anything, mock.Anything, mock.Anything).Return(&page2, nil).Once()
```
neel-astro: should we assert the offset change behavior in the API calls instead of mock.Anything, otherwise we are leaving a gap in our tests for the actual code to diverge

cloud/workspace/workspace.go
neel-astro, line 157: isolated from your existing change, but can we drop this errors.Wrap, because wrapping a nil `err` variable will result in nil.
```suggestion
if wsID == "" {
return errors.New("workspace id/name could not be found")
}
```

cloud/workspace/workspace.go
```diff
@@ -375,29 +375,45 @@ func selectWorkspace(workspaces []astrov1.Workspace) (astrov1.Workspace, error)
 	return selected, nil
 }
 
+// maxListPages caps a single paginated list call at maxListPages pages as a safety
+// bound against a server that mis-reports TotalCount (mirrors cloud/env).
+const maxListPages = 100
```
neel-astro: can we move this to the top of the packages, to keep all the constants and global variable declared at the same place

cloud/organization/organization.go
```diff
@@ -45,22 +45,37 @@ var organizationTableConfig = output.BuildTableConfig(
 	output.WithPadding([]int{44, 50}),
 )
 
+// maxListPages caps a single paginated list call at maxListPages pages as a safety
+// bound against a server that mis-reports TotalCount (mirrors cloud/env).
+const maxListPages = 100
```
neel-astro: nit: same here, group it with the constant section above it

cloud/organization/organization.go
```diff
@@ -45,22 +45,37 @@ var organizationTableConfig = output.BuildTableConfig(
...
-	if err != nil {
-		return nil, err
-	}
-	err = astrov1.NormalizeAPIError(resp.HTTPResponse, resp.Body)
-	if err != nil {
-		return nil, err
+	pageSize := 100
+	offset := 0
+	var orgs []astrov1.Organization
+	for page := 0; page < maxListPages; page++ {
```
neel-astro: nit: we are missing unit tests for this newly added pagination logic :)

neel-astro (commented):
LGTM mostly, left few tiny comments to squash

neel-astro (approved):
LGTM 🚀

PR #2212, opened 2026-07-21
Multiple build secrets

### LINKED ISSUE #1720: Fix --build-secrets behavior for multiple secrets
**Describe the bug**

the astro cli --build-secrets flag generates an incorrect docker command when passed multiple secrets.
`astro dev parse --build-secrets "id=secret1,env=ENV_SECRET" --build-secrets "id=secret2,env=ENV_SECRET2"`-> `build -t ul_44f240/airflow:latest -f Dockerfile . --secret id=secret1,env=ENV_SECRET,id=secret2,env=ENV_SECRET2`

The command should include multiple --secrets flags. `build -t ul_44f240/airflow:latest -f Dockerfile . --secret id=secret1,env=ENV_SECRET --secret id=secret2,env=ENV_SECRET2`

1.28.1

- [x] Astro
- [ ] Software
- [ ] None/Unknown

Affects both Mac and Windows

**🪜 Steps To Reproduce**

run commands above.

nickstenning:
## Description

Allow specifying multiple build secrets, using the CLI (`--build-secret`), dev config, or the `BUILD_SECRET_INPUT` env var.

The existing `--build-secrets` flag, despite its name, didn't actually allow passing multiple build secrets to Docker. It was parsed with `StringSliceVar`, meaning that literal build secret strings such as `id=aws,env=AWS_SECRET_ACCESS_KEY` would be split into two strings (`id=aws`, `env=AWS_SECRET_ACCESS_KEY`) and then rejoined into a single string that was ultimately passed as one `docker build --secret` argument.

Docker parses that as a single secret spec in which later fields override earlier ones, so providing more than one secret silently mounted only the last one.

As it was presumably originally intended that you could provide multiple secrets, this commit makes the following changes:

- Mark `--build-secrets` as deprecated. It can still be provided but will warn on use.
- Add `--build-secret`, which allows specifying a single secret, e.g. `id=aws,env=AWS_SECRET_ACCESS_KEY`, but can be provided multiple times.
- Pass all provided `--build-secret` values through to Docker.
- Dev config overrides as well as the `BUILD_SECRET_INPUT` env var can now contain newline-delimited strings to carry multiple build secrets. This is backwards compatible: a newline can't occur inside a valid secret specification, and a multiline value previously produced a single `--secret` argument that docker rejected, so no working configuration changes meaning.

Note: Comma-delimiting multiple secrets inside a single flag value (e.g. `id=a,src=x,id=b,env=Y`) is intentionally not supported: the comma already separates fields within one docker secret spec, so it cannot unambiguously separate specs too.

## 🎟 Issue(s)

Fixes #1720.

## 🧪 Functional Testing

- [x] Run `astro dev build` locally with multiple `--build-secret` args and verify all are passed to Docker.

## 📸 Screenshots

image

## 📋 Checklist

- ~[ ] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary)~.
- ~[ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary)~.
- [x] ~Communicated to/tagged owners of respective clients potentially impacted by these changes.~ (N/A: backwards-compatible change.)
- [x] Updated any related [documentation](https://github.com/astronomer/docs/): https://github.com/astronomer/docs/pull/7000 (to merge after this is released)

### DIFF
--- a/cmd/airflow.go
+++ b/cmd/airflow.go
@@ -52,7 +52,6 @@ var (
 	pytestFile             string
 	workspaceID            string
 	deploymentID           string
-	buildSecretString      string
 	followLogs             bool
 	schedulerLogs          bool
 	webserverLogs          bool
@@ -287,7 +286,9 @@ func newAirflowUpgradeTestCmd(astroV1Client astrov1.APIClient) *cobra.Command {
 	cmd.Flags().StringVarP(&lintConfigFile, "lint-config-file", "", "", "Relative path within project to a custom ruff config file. If not specified, a default config will be used.")
 	cmd.Flags().StringVarP(&deploymentID, "deployment-id", "i", "", "ID of the Deployment you want run dependency tests against.")
 	cmd.Flags().StringVarP(&customImageName, "image-name", "n", "", "Name of the upgraded image. Updates the FROM line in your Dockerfile to pull this image for the upgrade.")
-	cmd.Flags().StringSliceVar(&buildSecrets, "build-secrets", []string{}, "Expose a secret to containers. Equivalent to 'docker build --secret'. Example input id=mysecret,src=secrets.txt")
+	cmd.Flags().StringArrayVar(&buildSecrets, "build-secret", []string{}, "Expose a secret to containers. Equivalent to 'docker build --secret'. Example input id=mysecret,src=secrets.txt. Can be used multiple times to expose multiple secrets")
+	cmd.Flags().StringArrayVar(&buildSecrets, "build-secrets", []string{}, "Deprecated: use --build-secret instead")
+	cmd.Flags().MarkDeprecated("build-secrets", "use --build-secret instead") //nolint:errcheck
 	var err error
 	var avoidACFlag bool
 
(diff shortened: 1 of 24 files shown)


cmd/airflow.go
```diff
@@ -845,20 +856,20 @@ func airflowStart(cmd *cobra.Command, args []string, astroV1Client astrov1.APICl
 		return err
 	}
 
-	buildSecretString = util.GetbuildSecretString(buildSecrets, config.CFG.DevBuildSecrets.GetString())
+	resolvedBuildSecrets := util.ResolveBuildSecrets(buildSecrets, config.CFG.DevBuildSecrets.GetString(), os.Getenv("BUILD_SECRET_INPUT"))
```
jlaneve, line 847: mixing `--build-secret` and `--build-secrets` silently drops values: both bind to the same slice, and each pflag wrapper's first `Set` *replaces* it, so whichever flag appears later on the command line wins. since silently dropping a secret is the bug this PR fixes, probably worth erroring:

```suggestion
	if cmd.Flags().Changed("build-secret") && cmd.Flags().Changed("build-secrets") {
		return errors.New("cannot use --build-secret and --build-secrets together; use --build-secret")
	}

	resolvedBuildSecrets := util.ResolveBuildSecrets(buildSecrets, config.CFG.DevBuildSecrets.GetString(), os.Getenv("BUILD_SECRET_INPUT"))
```

same applies to the other commands that register both flags — might be worth a small shared helper

jlaneve (approved):
lgtm — threading the slice through instead of the joined string is the right shape, precedence order is preserved exactly, and the newline handling for config/env values is genuinely backwards compatible

left one non-blocking inline comment about mixing `--build-secret` and `--build-secrets` during the deprecation window — worth a look but I wouldn't hold the merge on it

pgvishnuram (approved):

cmd/airflow.go
nickstenning: Fixed this but went for the other option, where `--build-secrets` and `--build-secret` both add to the same slice but don't overwrite each other. Moved definition of these flags into a helper in `cmd/utils/flags.go` and wrote a unit test to ensure the behaviour is as expected.

PR #2236, opened 2026-08-03
feat(opt-in): Add dbt projects hash pre-deployment

tatiana:
### What

Adds an opt-in pre-deploy step to the Astro CLI. When it is enabled, the Cosmos Boost pre-deploy step runs over the deploy payload before that payload leaves the machine, so work that Cosmos would otherwise repeat on every DAG parse is done once, at deploy time.

Off by default. With `cosmos_boost.pre_deploy` unset or false, a deploy is a strict no-op for Cosmos Boost: nothing is walked, written, or removed. Artifacts left by earlier enabled deploys stay in place until removed with `astro dbt cleanup` (the command help says so).

The step runs in-process, no external tool involved: it discovers every dbt project (`dbt_project.yml`) and standalone dbt manifest in the payload and writes a `.astro/dbt_metadata.json` file beside each, carrying a deterministic content hash (`sha256-tree-v2` for project trees, `sha256-manifest-v2` for standalone manifests; `.git` is excluded from the tree hash, matching what actually deploys) that the Cosmos Boost plugin uses as its cache version key at parse time instead of hashing the project tree itself. A golden test pins the exact hash output.

### User-facing surface

| Item | Behavior |
|---|---|
| `astro config set cosmos_boost.pre_deploy true` | Enables the pre-deploy step. Default `false`. |
| `astro dbt cleanup [path ...]` | Removes the Cosmos Boost artifacts under each path (default `.`). |

### Where it hooks

| Deploy variant | Seam |
|---|---|
| `astro deploy --dags`, and the dags half of a default deploy | `UploadBundle`, before tar/gzip |
| `astro dbt deploy` | same seam, reached via `DeployBundle` |
| `astro deploy`, `--image` | `buildImage`, before `docker build` |
| `--image-name` (prebuilt image) | not covered, see Known gaps |

Both seams run before the payload is sealed, so what the step writes travels inside the artifact rather than being applied afterwards.

### What the CLI guarantees

- **Stamping is best effort.** A failing pre-deploy step produces a warning and the deploy or build continues: consumers fall back to today's behavior, hashing at parse time, so writing nothing is always safe.
- **An enabled deploy never ships a stale artifact.** Cleanup runs before every pre-deploy step, and consumers cannot tell fresh output from stale, so absence is the safety mechanism: an enabled deploy that cannot remove a leftover artifact fails with instructions rather than shipping it. (A disabled deploy deliberately does not touch the tree - see Known gaps.)
- **Cleanup only removes its own files.** A `dbt_metadata.json` this feature did not write (checked via its `generated_by` field) is never deleted, and nothing else under `.astro/` is ever touched.

### Known gaps

- A prebuilt image passed with `--image-name` is deployed as is; the CLI does not modify images it did not build.
- Sidecars written beside standalone or compiled manifests are not consumed yet: the plugin currently resolves the sidecar only next to the dbt project directory. The manifest-relative lookup on the plugin side is tracked in [BOSS-587](https://linear.app/astronomer/issue/BOSS-587); until it ships, those files are inert.
- Disabling the feature leaves previously written artifacts in place, and they ship with subsequent (disabled) deploys until removed with `astro dbt cleanup` - the price of a disabled deploy being a strict no-op.
- On enabled deploys, artifact removal happens in place in the deploy directory, so a leftover artifact that was committed to version control is deleted from the working copy.

## 🎟 Issue(s)

Closes: BOSS-540

## 🧪 Functional Testing

1. `make build`
2. `mkdir -p /tmp/shop/models && printf 'name: shop\n' > /tmp/shop/dbt_project.yml && echo 'select 1' > /tmp/shop/models/a.sql`
3. `./astro config set -g cosmos_boost.pre_deploy true`
4. Run `./astro dbt deploy <deployment-id>` from `/tmp/shop` (or `./astro deploy` on a project containing it) and confirm `Cosmos Boost pre-deploy step complete` is printed and `/tmp/shop/.astro/dbt_metadata.json` exists with `schema: 1` and a `version.hash`.
5. `./astro dbt cleanup /tmp/shop` and confirm the file is removed while any `dbt_metadata.json` not written by the CLI is left in place.
6. `./astro config set -g cosmos_boost.pre_deploy false`, deploy again, and confirm the tree is untouched (a leftover artifact stays until `astro dbt cleanup` removes it).

## 📸 Screenshots

Not applicable (CLI change); sample artifact written by step 4:

```json
{
  "schema": 1,
  "version": {
    "algo": "sha256-tree-v2",
    "hash": "358d67472294311e4e9574ce191488e94b94f71e9e9fb4b662f937e9e68837ee"
  },
  "generated_by": {
    "application": "astro",
    "version": "SNAPSHOT-<commit>"
  }
}
```

## 📋 Checklist

Note on `make test`: all packages pass except `TestAirflow/TestHasProxyConfigured` (airflow package), which fails identically on the base commit in this environment - pre-existing and unrelated to this change. `make lint` equivalent run (golangci-lint v2) over all touched packages: 0 issues.


cloud/deploy/bundle.go
```diff
@@ -216,6 +217,16 @@ func UploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool,
 		}
 	}()
 
+	// Cosmos Boost pre-deploy step. Removing leftover artifacts is mandatory —
+	// a stale one must not ship inside the bundle — while stamping fresh ones
+	// is opt-in and best-effort.
+	if err := cosmosboost.EnsureClean(bundlePath); err != nil {
```
jeremybeard: Should we also gate this (and in `deploy.go`) behind the config? As it is here if you're not using the feature you now have to do a tree walk, and if it fails then the deploy is blocked (and `astro dbt cleanup` will fail for the same reason).

tatiana: Agreed and done in 15472c87

pkg/cosmosboost/precompute/cleanup.go
```diff
@@ -0,0 +1,155 @@
...
+func removeSidecar(path string) CleanupResult {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return CleanupResult{Path: path, Err: err}
+	}
+	var meta Metadata
+	if json.Unmarshal(data, &meta) != nil ||
+		(meta.GeneratedBy.Application != application && meta.GeneratedBy.Application != legacyApplication) {
+		return CleanupResult{Path: path, Kept: true}
+	}
```
neel-astro: curious q: should we log an error or have some guideline on how to fix this? I think as of now, an unmarshal error would mean we ignore that as "not cosmosboost" owned, but if something did go wrong on our end, then this cleanup bit would ignore that, and push it on subsequent deploy, correct?

pkg/cosmosboost/precompute/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+	// algoProjectTree hashes a whole dbt project directory (source files).
+	// v2 excludes .git, which never ships in a deploy payload, so VCS activity
+	// (commits, fetches, gc) cannot change the hash of unchanged content.
+	algoProjectTree = "sha256-tree-v2"
+	// algoManifestJSON hashes a dbt manifest.json by content, ignoring volatile
+	// metadata. Used for manifest-only deployments.
+	algoManifestJSON = "sha256-manifest-v1"
+
+	sidecarDir  = ".astro"
+	sidecarName = "dbt_metadata.json"
```
neel-astro, line 29: probably worth adding the json file to `pkg/airflowrt/include/gitignore`?

pkg/cosmosboost/precompute/cleanup.go
```diff
@@ -0,0 +1,155 @@
...
+// such as a root that cannot be walked.
+func Cleanup(roots []string) (CleanupSummary, error) {
+	start := time.Now()
+
+	seen := map[string]bool{}
+	var results []CleanupResult
+	for _, root := range roots {
+		err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
```
kaxil, line 58: Returning the error here means one unreadable directory fails the whole cleanup, and EnsureClean turns that into a failed deploy for anyone with the flag on. The case I'd worry about is a root-owned `logs/` subdir left behind by a Linux bind mount. Docker never reads it since `.dockerignore` excludes it, so that project deploys fine today, but this walk will hit it and abort, and the fix suggested in the error message (`astro dbt cleanup`) dies on the same directory. Run never writes sidecars under `logs/` or `dbt_packages/` anyway, both discovery walks skip them by name. Should the cleanup walk skip them too?

cloud/deploy/deploy.go
```diff
@@ -712,6 +713,19 @@ func buildImage(path, currentVersion, deployImage, imageName, organizationID str
...
 		fmt.Println(composeImageBuildingPromptMsg)
 
+		// Cosmos Boost pre-deploy step, opt-in via cosmos_boost.pre_deploy;
+		// with the setting off the build does not touch the tree at all.
+		// Cleanup runs first and is fatal on failure (a stale artifact must
+		// not ship inside the image), while stamping is best-effort (a
+		// missing artifact is safe). Artifacts left by earlier enabled
+		// deploys are removed with `astro dbt cleanup`.
+		if config.CFG.CosmosBoostPreDeploy.GetBool() {
+			if err := cosmosboost.EnsureClean(path); err != nil {
```
kaxil, line 723: With DAG deploy enabled this stamps everything under `path`, including dbt projects under `dags/` that the image build excludes anyway. The dags half of the deploy later runs EnsureClean over `dagsPath` via UploadBundle, deletes those fresh sidecars, and re-stamps them with the same content. The result is correct since the hash is deterministic, but the dags tree gets hashed twice per deploy. Worth skipping `dagsPath` here when the dags half runs its own pass?

cloud/deploy/deploy.go
```diff
@@ -712,6 +713,19 @@ func buildImage(path, currentVersion, deployImage, imageName, organizationID str
 		// Build our image
 		fmt.Println(composeImageBuildingPromptMsg)
 
+		// Cosmos Boost pre-deploy step, opt-in via cosmos_boost.pre_deploy;
+		// with the setting off the build does not touch the tree at all.
+		// Cleanup runs first and is fatal on failure (a stale artifact must
+		// not ship inside the image), while stamping is best-effort (a
+		// missing artifact is safe). Artifacts left by earlier enabled
+		// deploys are removed with `astro dbt cleanup`.
+		if config.CFG.CosmosBoostPreDeploy.GetBool() {
```
pankajastro, line 722: Is it intentional that a dbt project under `dags/` gets hashed and stamped twice per deploy — once here over the full `path` before `buildImageWithoutDags` excludes `dags/` via `.dockerignore`, and again in `deployDags`→`UploadBundle(dagsPath,...)` right after?

pkg/cosmosboost/precompute/dbtproject.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	cfg := dbtConfig{
+		targetPath: cleanRelDir(raw.TargetPath),
+		logPath:    cleanRelDir(raw.LogPath),
+	}
+	switch {
+	case raw.PackagesInstallPath == "":
+		// unset → the caller excludes the default dbt_packages/
+	case strings.Contains(raw.PackagesInstallPath, "{{"):
```
pankajastro: This guards `packages-install-path` against Jinja templates (`{{...}}`) but `target-path`/`log-path` aren't checked the same way — should those get the same guard/warning?

pkg/cosmosboost/precompute/hash.go
```diff
@@ -0,0 +1,217 @@
...
+			// depth. A dbt manifest.json in a subdirectory (e.g. docs/manifest.json)
+			// gets its own sidecar written into the walked tree (docs/.astro/), and
+			// because manifest and project units run concurrently, hashing it would
+			// make the project hash depend on that sidecar and vary between runs.
+			// .astro is always our output, never dbt source, so skipping it by name
+			// everywhere is safe (unlike target/logs, which are matched root-relative
+			// to avoid over-excluding a same-named source dir).
+			if d.Name() == sidecarDir || d.Name() == gitDir {
+				return filepath.SkipDir
+			}
```
neel-astro, line 109: isn't these two dirs already present as part of `excludedRel` and covered in the check below this one?

neel-astro (approved):
Overall looking good, left few curious comments

pkg/cosmosboost/precompute/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+	"path/filepath"
+)
+
+// application is the producer name recorded in the sidecar. (The version is owned
+// by the caller and passed in to Run.)
+const application = "astro"
+
+// legacyApplication is the producer name written by the retired standalone
+// astro-cosmos-boost helper; its sidecars are still ours to remove.
+const legacyApplication = "astro-cosmos-boost"
```
pankajkoti: Do we need to have this considered?

pkg/cosmosboost/precompute/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	// schemaVersion lets the read-side cope with future format changes.
+	schemaVersion = 1
+
+	// algoProjectTree hashes a whole dbt project directory (source files).
+	// v2 excludes .git, which never ships in a deploy payload, so VCS activity
+	// (commits, fetches, gc) cannot change the hash of unchanged content.
+	algoProjectTree = "sha256-tree-v2"
```
pankajkoti, line 20: And a quick follow-up question on whether this needs to be v2?

pankajkoti (approved):
Thanks for working on this @tatiana ! 

Wish to understand on why we're having this as an opt-in and not enabled by default? Are there some reservations?

tatiana:
Clarification on manifest sidecar placement, from a review question by @pankajkoti (raising it here for visibility):

You're right - today the plugin only resolves the sidecar next to the dbt project folder, so the ones written beside manifests (e.g. `target/.astro/`) have no reader yet; nobody should point their project dir at `target/`.

Keeping this in the CLI does not break the plugin - it validates `schema` and `version.hash` and ignores files it does not consume - so the plan is to keep the CLI PR as is and fix the lookup on the plugin side to read next to the manifest - thanks for filing [BOSS-587](https://linear.app/astronomer/issue/BOSS-587) for exactly that.

pkg/cosmosboost/precompute/cleanup.go
tatiana: Fixed in 910e7df8: the cleanup walk now skips `logs/` and `dbt_packages/` by name (plus `.git`) - discovery never writes sidecars under them, so an unreadable bind-mount leftover no longer aborts a deploy or `astro dbt cleanup`. `target/` deliberately stays visited: a compiled manifest's sidecar lives at `target/.astro/`. An unreadable directory with any other name still fails an enabled deploy, intentionally - a sidecar could genuinely sit inside, and skipping it would reopen the stale-artifact hole.

cloud/deploy/deploy.go
tatiana: Confirmed (and @pankajastro reported the same in https://github.com/astronomer/astro-cli/pull/2236#discussion_r3711116427) - the dags subtree is hashed twice on image deploys with DAG deploy enabled. Correct output, wasted work. Tracking the skip as an explicit follow-up so it doesn't hold this PR: https://linear.app/astronomer/issue/BOSS-588 (image hook skips `dagsPath` whenever the bundle hook owns it).

cloud/deploy/deploy.go
tatiana: Not intentional as a permanent state - consolidated with Kaxil's identical finding in https://github.com/astronomer/astro-cli/pull/2236#discussion_r3710907772 and tracked as https://linear.app/astronomer/issue/BOSS-588: the image hook will skip `dagsPath` whenever the dags half of the deploy owns it. Performance only (the hash is deterministic, so the double stamp produces identical artifacts), hence the follow-up rather than blocking this PR.

pkg/cosmosboost/precompute/cleanup.go
tatiana: Good question - it exposed a real gap, closed in 910e7df8. The semantics now: a file that fails to parse or carries an unrecognized `generated_by` is never deleted (we don't own it), but an ENABLED deploy now refuses to ship it - EnsureClean fails naming the files and asking for manual removal, because the plugin would consume its `version.hash` without being able to prove freshness. `astro dbt cleanup` still preserves such files and reports them as "unrecognized producer; left in place". Malformed JSON on the plugin side is safe either way (treated as absent, falls back to live hashing); the dangerous case was a well-formed sidecar we didn't write, and that can no longer ship through an enabled deploy.

pkg/cosmosboost/precompute/dbtproject.go
tatiana: Fixed in e477a91d - `target-path` and `log-path` now get the same guard as `packages-install-path`: an unresolved Jinja value is treated as unset (the dbt default stays excluded) instead of becoming a meaningless literal exclusion, and the stamped unit's warning names every templated setting. Table test covers all three.

pkg/cosmosboost/precompute/metadata.go
tatiana: Done in c89bf3e7 - added `**/.astro/dbt_metadata.json` to the scaffold gitignore (`**/` because dbt projects can be nested) and extended the scaffold test.

pkg/cosmosboost/precompute/hash.go
tatiana: No functional overlap, though it does look like it: this check skips `.astro` and `.git` BY NAME at every depth (a sidecar directory can appear anywhere a manifest was stamped, and worktree/submodule `.git`s can be nested), while `excludedRel` below matches configured directories only at their root-relative paths - deliberately, so a source directory that happens to be called `target` deeper in the tree is not over-excluded. Also `.git` is not in `excludedRel` at all. Happy to expand the comment if it would help.

pkg/cosmosboost/precompute/metadata.go
tatiana: Fair question. No public release ever wrote that producer value - it exists so machines that tested the pre-release standalone binary internally still get their leftovers cleaned. It costs one string comparison, but if you'd rather not carry Astronomer-internal history in the public codebase I'm happy to drop the constant (those few internal machines can remove leftovers manually).

pkg/cosmosboost/precompute/metadata.go
tatiana: The algo name identifies the computation, not the release: a v1 of this computation (which included .git in the tree walk) did exist and produced artifacts during internal validation, so giving a different computation the same name felt riskier than starting the public history at v2. The plugin ignores the algo field entirely (it gates on schema and reads version.hash opaquely), so the only cost is cosmetic. If you'd prefer the public repo to start at v1 I can collapse it - nothing consumed v1 outside Astronomer.

tatiana:
Review round addressed in c309aae7..c89bf3e7 (all threads replied inline; two items are summarized here because their original threads are no longer live):

- **Manifest hashes are now stable across full parses** (c309aae7): dbt-owned `created_at` on every resource (including the `disabled` collection's list entries) and `metadata.run_started_at` are stripped; user `meta.created_at` inside a resource is preserved; algo renamed `sha256-manifest-v2`; regression test pins two full-parse-shaped manifests to one hash.
- **Enabled deploys refuse to ship unrecognized sidecars** (910e7df8): cleanup still never deletes a file it does not own, but `EnsureClean` now fails naming any kept file and asking for manual removal, so a foreign `version.hash` cannot ship and be consumed; `astro dbt cleanup` keeps preserving such files ("unrecognized producer; left in place").
- **Cleanup no longer traverses `logs/` / `dbt_packages/`** (910e7df8) - discovery never writes there; `target/` stays visited (manifest sidecars live at `target/.astro/`).
- **Jinja guard generalized** to `target-path`/`log-path` (e477a91d) and **scaffold gitignore** gained `**/.astro/dbt_metadata.json` (c89bf3e7).
- **Double hashing under `dags/`**: confirmed, consolidated across both reports, tracked as an explicit follow-up in [BOSS-588](https://linear.app/astronomer/issue/BOSS-588) (performance, not correctness).

Patch coverage remains 100%; all tests pass with `-race`; lint 0 issues.

tatiana:
> Thanks for working on this @tatiana !
> 
> Wish to understand on why we're having this as an opt-in and not enabled by default? Are there some reservations?

@pankajkoti  This is opt-in while in labs - and potentially while in preview. I believe it should be changed to default when we move to GA, though.

pkg/cosmosboost/precompute/metadata.go
tatiana: Removed in 725e0fd7 per discussion - the constant referenced a pre-release standalone binary that never shipped publicly and never made it into the Astro CLI itself, so there was nothing real for it to recognize. Cleanup now accepts exactly one producer: `astro`.

tatiana:
@jeremybeard @kaxil @pankajastro @neel-astro @pankajkoti Thanks a lot for the reviews and valuable feedback. I believe I addressed all the feedback.

Please let me know if there is anything - otherwise, I'll ask someone with the right super powers to merge this PR :D

PR #2238, opened 2026-08-05
Attach the Astro token to CLI telemetry so events can be attributed

jlaneve:
## Description

The CLI's telemetry request is unauthenticated, so events arrive with no way to associate them with an organization.

This attaches the caller's existing Astro bearer token to the telemetry POST when one is available. Requests without credentials are unchanged and still send — that path has to keep working.

- `SendWithToken()` sets `Authorization: Bearer` when the token is non-empty. `Send()` keeps its current signature and delegates with an empty token, so callers of `pkg/telemetry` outside this repo are unaffected.
- `currentToken()` reads `ASTRO_API_TOKEN`, then the active context. Local reads only — no network call and no token refresh, since the sender runs in a detached subprocess and refreshing would race the main process for `~/.astro/config.yaml`. A stale token costs one unattributed event and never blocks the user's command.
- Contexts persist the token with a `Bearer ` prefix already, while `ASTRO_API_TOKEN` holds it raw. Both are normalized to the raw token and the scheme is added when the header is set, matching `cmd/auth.go`.
- The token is passed to the sender subprocess on stdin rather than argv, and is sent as a header. It is not written into the event body.
- `SenderPayload` is exported from `pkg/telemetry` and used directly by the caller, rather than redeclared per module with hand-matched JSON tags.

The notice was also printed before the filter that skips untracked commands, so `astro telemetry disable` announced what we collect to someone in the act of opting out. It now prints only on commands that actually send telemetry — so the first time it appears is the first time it is true.

The first-run notice previously described collection as anonymous. It is reworded, and versioned so users who saw the earlier wording are shown the update once rather than inheriting the change.

No behavior change until the telemetry endpoint reads the header.

Verified against the live endpoint with a real token: a valid token returns `200`, an expired one
returns `401 Jwt is expired`, and no header returns `200`. The change is inert until the server also
stamps org/user onto the event — safe to merge before then, just not yet visible in the data.

## 🎟 Issue(s)

Server-side counterpart tracked separately.

## 🧪 Functional Testing

Unit tests cover `SendWithToken` and `currentToken`. The subprocess hand-off is not covered by unit tests, so it was checked end to end against a local listener:

| Case | Expected | Result |
|---|---|---|
| Logged in | `Authorization` header present | pass |
| Logged out | no header, event still sends | pass |
| `ASTRO_API_TOKEN` set | `Authorization` header present | pass |

Notice behavior was checked the same way, since `TrackCommand` returns early in test binaries and cannot be unit-tested:

| Command | Expected | Result |
|---|---|---|
| `astro version` | notice shown | pass |
| `astro telemetry disable` | no notice | pass |
| `astro telemetry enable` | no notice | pass |

Token absent from the event body in all three.

To reproduce: point `ASTRO_TELEMETRY_API_URL` at a local listener, set `HOME` to a temporary directory containing `.astro/config.yaml`, run any command, and inspect the received request.

## 📸 Screenshots

n/a — no user-visible output beyond the reworded notice.

## 📋 Checklist

- [ ] Tested against [Astro-API](https://github.com/astronomer/astro/) (if necessary)
- [ ] Tested against [Houston-API](https://github.com/astronomer/houston-api/) and [Astronomer](https://github.com/astronomer/astronomer/) (if necessary)
- [ ] Communicated to/tagged owners of respective clients potentially impacted by these changes
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/) — telemetry docs need the reworded notice

### DIFF
--- a/cmd/telemetry.go
+++ b/cmd/telemetry.go
@@ -13,15 +13,18 @@ import (
 func newTelemetryCmd(out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "telemetry",
-		Short: "Manage anonymous telemetry settings",
-		Long: `Manage anonymous telemetry settings for the Astro CLI.
+		Short: "Manage telemetry settings",
+		Long: `Manage telemetry settings for the Astro CLI.
 
 Telemetry helps us understand how the CLI is used and improve it.
-We collect anonymous usage data including:
+We collect usage data including:
 - Commands used (not arguments or values)
 - CLI version
 - Operating system
 - Invocation context (CI, interactive, etc.)
+
+While you are logged in, events are linked to your Astro organization so we can
+see how the CLI is used across accounts. Logged-out usage stays anonymous.
 
 No personally identifiable information is collected.
 You can opt out at any time using 'astro telemetry disable' or by setting
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -55,6 +55,27 @@ func IsInteractive() bool {
 	return term.IsTerminal(int(os.Stdin.Fd()))
 }
 
+// currentToken returns the Astro bearer token for the active context, or "" when
+// the user has not logged in and no API token is configured.
+//
+// It reads local state only — no network call and no token refresh. The sender
+// runs in a detached subprocess alongside the main command, so refreshing here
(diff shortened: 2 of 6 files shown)


schnie (approved):
Looks good. Second notice version codified so users should see an updated notification.

PR #2242, opened 2026-08-12
feat: Generate slim field-filtered dbt manifest at deploy time

pankajastro:
## Summary

Cosmos reads only a handful of fields from `manifest.json` when building DAGs (`LoadMode.DBT_MANIFEST`), so on large projects the full manifest inflates dag-processor parse time and memory proportionally (~7x file size / ~5x peak memory in internal testing).

The Cosmos Boost pre-deploy step now writes a slim, field-filtered copy of each discovered `manifest.json` to its manifest-relative `.astro/manifest.slim.json`, next to the existing `dbt_metadata.json` hash sidecar. The original `manifest.json` is never modified.

Linear: [BOSS-598](https://linear.app/astronomer/issue/BOSS-598/boost-plugin-astro-cli-generate-a-slim-field-filtered-dbt-manifest-at)

## Gating

One master switch, this optimization on by default underneath it, disabled per-feature by env var — the shape `astro-cosmos-boost-plugin` uses.

| Switch | Default | Effect |
| --- | --- | --- |
| `cosmos_boost.pre_deploy` (config) | off | the step as a whole; writes the hash sidecars |
| `ASTRO_COSMOS_BOOST_SLIM_MANIFEST_DISABLED` (env) | unset → enabled | set truthy to skip the slim manifest only |

Named for the *disable* so the repo's existing `util.CheckEnvBool` does the parsing: it counts only explicit truthy values, so unset or misspelled input leaves the optimization on.

## What's written

The sidecar gains an optional pointer, making it the consumer's single entry point — it discovers the slim manifest there and falls back to the full one when the section is absent:

```json
"filtered_manifest": {
  "schema": 1,
  "path": "manifest.slim.json",
  "version": { "algo": "sha256-filtered-manifest-v1", "hash": "..." }
}
```

Top-level `version.hash` still hashes the source manifest; this one hashes the slim file's bytes, so a consumer can confirm the two are a matched pair without re-hashing.

Kept per resource: `original_file_path`, `package_name`, `resource_type`, `tags`, `config` (whole), `fqn`, `depends_on.nodes`, `freshness` (sources only), `database`, `schema`, `alias`, `name`. Kept top-level: `nodes`, `sources`, `exposures`, `selectors`, `metadata.project_name`. Everything else is dropped.

Verified field-by-field against `astronomer-cosmos` 1.15.1 at both places it reads a manifest — `graph.py::_build_dbt_node_from_manifest_resource` for the `DbtNode`, and `dataset.py::compute_model_outlet_uris`, which builds outlet URIs as `database.schema.alias` from this file under `ExecutionMode.WATCHER` on Kubernetes. Absent keys are omitted rather than emitted as `null`.

## Cleanup

`astro dbt cleanup` / `EnsureClean` handle both artifacts, each judged by its own producer marker — so a foreign `manifest.slim.json` survives beside our sidecar, ours is still removed beside a foreign sidecar, and a failure names the file that actually failed. A slim manifest from an unrecognized producer now fails `EnsureClean`, same as a foreign sidecar.

## Scope

astro-cli producer side only. The plugin does not read `manifest.slim.json` yet — separate change in `astro-cosmos-boost-plugin`, mirroring how BOSS-587/591 were separate producer/consumer tickets.

Two naming calls worth a second opinion, both public interface: `filtered_manifest` is snake_case to match `generated_by` in the same file, and the env var is `..._DISABLED` rather than `..._ENABLED` for the reason above. Happy to flip either.

## Testing

Verified in a `golang:1.26` container (no Go toolchain on the machine used): `go build ./...`, `go vet`, `gofumpt -l` clean, `golangci-lint run` (pinned `v2.11.4`) `0 issues`, `go test ./...` passing except pre-existing environment-only failures in `airflow`/`cmd`/`pkg/git` that reproduce identically at the pre-change commit. Also checked end to end through the exported API: both artifacts written, sidecar carrying both hashes, `manifest.json` byte-identical afterwards.

Unit tests cover the field allowlist, freshness sources-only, no-nulls, a size guard (slim under a third of full on a proportioned fixture), one test per gating state, `manifest.json` unchanged after a run, and per-artifact cleanup provenance in both directions.

### DIFF
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -227,7 +227,7 @@ func UploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool,
 		if err := cosmosboost.EnsureClean(bundlePath); err != nil {
 			return "", err
 		}
-		cosmosboost.BestEffortPreDeploy(bundlePath, config.CFG.CosmosBoostSlimManifest.GetBool())
+		cosmosboost.BestEffortPreDeploy(bundlePath)
 	}
 
 	// Generate the bundle tar
--- a/cloud/deploy/bundle.go
+++ b/cloud/deploy/bundle.go
@@ -227,7 +227,7 @@ func UploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool,
 		if err := cosmosboost.EnsureClean(bundlePath); err != nil {
 			return "", err
 		}
-		cosmosboost.BestEffortPreDeploy(bundlePath)
+		cosmosboost.BestEffortPreDeploy(bundlePath, config.CFG.CosmosBoostSlimManifest.GetBool())
 	}
 
 	// Generate the bundle tar
--- a/cloud/deploy/cosmosboost_hook_test.go
+++ b/cloud/deploy/cosmosboost_hook_test.go
@@ -58,9 +58,32 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
 
 // TestUploadBundleWritesSlimManifestWhenEnabled pins that a discovered
 // manifest.json gets a slim, field-filtered copy alongside its hash sidecar
-// whenever cosmos_boost.pre_deploy is enabled - there is no separate opt-in
-// for this.
+// once cosmos_boost.slim_manifest is enabled on top of the step itself.
 func TestUploadBundleWritesSlimManifestWhenEnabled(t *testing.T) {
+	setupCosmosBoostEnv(t)
+	require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
+	require.NoError(t, config.CFG.CosmosBoostSlimManifest.SetHomeString("true"))
+
+	bundleDir := writeDbtBundle(t)
+	manifest := `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`
+	require.NoError(t, os.MkdirAll(filepath.Join(bundleDir, "target"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "target", "manifest.json"), []byte(manifest), 0o644))
+
+	azureUploader = func(sasLink string, file io.Reader) (string, error) {
+		return "version-id", nil
+	}
+
+	_, err := UploadBundle(t.TempDir(), bundleDir, "http://upload-url", false, "0.0.0")
+	require.NoError(t, err)
+
+	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "dbt_metadata.json"), "the manifest must still be stamped")
+	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
+}
+
+// TestUploadBundleSkipsSlimManifestByDefault: cosmos_boost.slim_manifest is a
+// switch of its own, defaulting off while no released plugin reads the file.
+// Enabling the step alone must still stamp the hash sidecar and nothing more.
+func TestUploadBundleSkipsSlimManifestByDefault(t *testing.T) {
 	setupCosmosBoostEnv(t)
 	require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
 
@@ -77,7 +100,7 @@ func TestUploadBundleWritesSlimManifestWhenEnabled(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "dbt_metadata.json"), "the manifest must still be stamped")
-	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
+	assert.NoFileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
 }
 
 func TestUploadBundleSkipsPreDeployByDefault(t *testing.T) {
--- a/cloud/deploy/cosmosboost_hook_test.go
+++ b/cloud/deploy/cosmosboost_hook_test.go
@@ -56,6 +56,30 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
 		"the pre-deploy step must stamp the bundle before it is tarred")
 }
 
+// TestUploadBundleWritesSlimManifestWhenEnabled pins that a discovered
+// manifest.json gets a slim, field-filtered copy alongside its hash sidecar
+// whenever cosmos_boost.pre_deploy is enabled - there is no separate opt-in
+// for this.
+func TestUploadBundleWritesSlimManifestWhenEnabled(t *testing.T) {
+	setupCosmosBoostEnv(t)
+	require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
+
+	bundleDir := writeDbtBundle(t)
+	manifest := `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`
+	require.NoError(t, os.MkdirAll(filepath.Join(bundleDir, "target"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "target", "manifest.json"), []byte(manifest), 0o644))
+
+	azureUploader = func(sasLink string, file io.Reader) (string, error) {
+		return "version-id", nil
+	}
+
+	_, err := UploadBundle(t.TempDir(), bundleDir, "http://upload-url", false, "0.0.0")
+	require.NoError(t, err)
+
+	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "dbt_metadata.json"), "the manifest must still be stamped")
+	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
+}
+
 func TestUploadBundleSkipsPreDeployByDefault(t *testing.T) {
 	setupCosmosBoostEnv(t)
 	// cosmos_boost.pre_deploy defaults to false, so nothing may be produced.
--- a/cloud/deploy/cosmosboost_hook_test.go
+++ b/cloud/deploy/cosmosboost_hook_test.go
@@ -56,51 +56,48 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
 		"the pre-deploy step must stamp the bundle before it is tarred")
 }
 
-// TestUploadBundleWritesSlimManifestWhenEnabled pins that a discovered
-// manifest.json gets a slim, field-filtered copy alongside its hash sidecar
-// once cosmos_boost.slim_manifest is enabled on top of the step itself.
-func TestUploadBundleWritesSlimManifestWhenEnabled(t *testing.T) {
-	setupCosmosBoostEnv(t)
-	require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
-	require.NoError(t, config.CFG.CosmosBoostSlimManifest.SetHomeString("true"))
+// TestUploadBundleSlimManifest pins the slim manifest's gating end to end: on
+// by default once the step is enabled, and off only when the per-feature env
+// var disables it. Either way the hash sidecar still lands - the two
+// optimizations are independent.
+func TestUploadBundleSlimManifest(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		disable  string // value for the disable env var; "" leaves it unset
+		wantSlim bool
+	}{
+		{name: "on by default", wantSlim: true},
+		{name: "env var disables it", disable: "true", wantSlim: false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			setupCosmosBoostEnv(t)
+			require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
+			if tc.disable != "" {
+				t.Setenv("ASTRO_COSMOS_BOOST_SLIM_MANIFEST_DISABLED", tc.disable)
+			}
 
-	bundleDir := writeDbtBundle(t)
-	manifest := `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`
-	require.NoError(t, os.MkdirAll(filepath.Join(bundleDir, "target"), 0o755))
-	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "target", "manifest.json"), []byte(manifest), 0o644))
+			bundleDir := writeDbtBundle(t)
+			manifest := `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`
+			require.NoError(t, os.MkdirAll(filepath.Join(bundleDir, "target"), 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "target", "manifest.json"), []byte(manifest), 0o644))
 
-	azureUploader = func(sasLink string, file io.Reader) (string, error) {
-		return "version-id", nil
+			azureUploader = func(sasLink string, file io.Reader) (string, error) {
+				return "version-id", nil
+			}
+
+			_, err := UploadBundle(t.TempDir(), bundleDir, "http://upload-url", false, "0.0.0")
+			require.NoError(t, err)
+
+			astroDir := filepath.Join(bundleDir, "target", ".astro")
+			assert.FileExists(t, filepath.Join(astroDir, "dbt_metadata.json"), "the manifest must still be stamped")
+			slim := filepath.Join(astroDir, "manifest.slim.json")
+			if tc.wantSlim {
+				assert.FileExists(t, slim)
+			} else {
+				assert.NoFileExists(t, slim)
+			}
+		})
 	}
-
-	_, err := UploadBundle(t.TempDir(), bundleDir, "http://upload-url", false, "0.0.0")
-	require.NoError(t, err)
-
-	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "dbt_metadata.json"), "the manifest must still be stamped")
-	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
-}
-
-// TestUploadBundleSkipsSlimManifestByDefault: cosmos_boost.slim_manifest is a
-// switch of its own, defaulting off while no released plugin reads the file.
-// Enabling the step alone must still stamp the hash sidecar and nothing more.
-func TestUploadBundleSkipsSlimManifestByDefault(t *testing.T) {
-	setupCosmosBoostEnv(t)
-	require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
-
-	bundleDir := writeDbtBundle(t)
-	manifest := `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`
-	require.NoError(t, os.MkdirAll(filepath.Join(bundleDir, "target"), 0o755))
-	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "target", "manifest.json"), []byte(manifest), 0o644))
-
-	azureUploader = func(sasLink string, file io.Reader) (string, error) {
-		return "version-id", nil
-	}
-
-	_, err := UploadBundle(t.TempDir(), bundleDir, "http://upload-url", false, "0.0.0")
-	require.NoError(t, err)
-
-	assert.FileExists(t, filepath.Join(bundleDir, "target", ".astro", "dbt_metadata.json"), "the manifest must still be stamped")
-	assert.NoFileExists(t, filepath.Join(bundleDir, "target", ".astro", "manifest.slim.json"))
 }
 
 func TestUploadBundleSkipsPreDeployByDefault(t *testing.T) {
--- a/config/config.go
+++ b/config/config.go
@@ -102,6 +102,12 @@ var (
 		ProxyPort:               newCfg("proxy.port", "6563"),
 		OttoAutoUpdate:          newCfg("otto.auto_update", "true"),
 		CosmosBoostPreDeploy:    newCfg("cosmos_boost.pre_deploy", "false"),
+		// Gates the slim, field-filtered manifest copy only; the hash sidecar
+		// is written either way. Off by default because no released Cosmos
+		// Boost plugin reads manifest.slim.json yet - until one does, writing
+		// it only costs deploy time and payload size. Flip to "true" once the
+		// plugin ships the read-with-fallback path.
+		CosmosBoostSlimManifest: newCfg("cosmos_boost.slim_manifest", "false"),
 	}
 
 	// viperHome is the viper object in the users home directory
--- a/config/config.go
+++ b/config/config.go
@@ -102,12 +102,6 @@ var (
 		ProxyPort:               newCfg("proxy.port", "6563"),
 		OttoAutoUpdate:          newCfg("otto.auto_update", "true"),
 		CosmosBoostPreDeploy:    newCfg("cosmos_boost.pre_deploy", "false"),
-		// Gates the slim, field-filtered manifest copy only; the hash sidecar
-		// is written either way. Off by default because no released Cosmos
-		// Boost plugin reads manifest.slim.json yet - until one does, writing
-		// it only costs deploy time and payload size. Flip to "true" once the
-		// plugin ships the read-with-fallback path.
-		CosmosBoostSlimManifest: newCfg("cosmos_boost.slim_manifest", "false"),
 	}
 
 	// viperHome is the viper object in the users home directory
--- a/pkg/cosmosboost/precompute/cleanup.go
+++ b/pkg/cosmosboost/precompute/cleanup.go
@@ -26,8 +26,6 @@ var cleanupSkipDirs = map[string]bool{
 }
 
 // CleanupResult records what happened to one artifact found during Cleanup.
-// Err always describes Path: each artifact is removed on its own terms, so a
-// failure is never reported against a neighboring file's name.
 type CleanupResult struct {
 	Path string // artifact file path
 	Kept bool   // left in place: the file was not written by this tool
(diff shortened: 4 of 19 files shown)


pkg/cosmosboost/precompute/slim.go
```diff
@@ -0,0 +1,88 @@
...
+const slimManifestName = "manifest.slim.json"
+
+// slimSections are the only top-level manifest collections Cosmos loads nodes
+// from when parsing a dbt manifest (LoadMode.DBT_MANIFEST); everything else
+// (macros, disabled, docs, parent_map, child_map, ...) is unused.
+var slimSections = []string{"nodes", "sources", "exposures"}
+
+// slimResourceFields are the per-resource fields Cosmos reads to build a
+// DbtNode from a manifest resource.
+var slimResourceFields = []string{"original_file_path", "package_name", "resource_type", "tags", "config", "fqn"}
```
kaxil: The allowlist covers `_load_nodes_from_manifest_data`, but `ProjectConfig.manifest_path` feeds two other Cosmos consumers and neither survives slimming. `LocalDbtBaseOperator` copies it verbatim into `<tmp_project>/target/manifest.json` via `copy_manifest_file_if_exists` (cosmos/operators/local.py), where dbt itself picks it up. And `_k8s_common.py` hands it to `compute_model_outlet_uris`, which reads `database`, `schema`, `alias`/`name` off each node and skips any node missing them, so per-model dataset emission silently drops to zero rather than erroring. Is the plan to keep the consumer's swap scoped to graph loading, or should those fields join the allowlist? Worth settling here, since the field set is what the plugin has to live with once it ships.

pkg/cosmosboost/precompute/cleanup.go
```diff
@@ -119,6 +121,8 @@ func removeSidecar(path string) CleanupResult {
 	if err := os.Remove(path); err != nil {
 		return CleanupResult{Path: path, Err: err}
 	}
+	// Best-effort: a slim manifest is optional, so its absence is not an error.
+	_ = os.Remove(filepath.Join(filepath.Dir(path), slimManifestName))
```
kaxil: Discarding the error here can leave an orphan nothing can find later. `Cleanup` only walks for `dbt_metadata.json`, so once the sidecar is gone (line 121) a `manifest.slim.json` that failed to delete for any reason other than ENOENT is invisible to every future `astro dbt cleanup` and to `EnsureClean`, which then reports clean and lets the stale file ship. I checked it: remove the sidecar by hand, leave the slim file, and `Cleanup` returns 0 results with the file still sitting there. Removing the slim manifest before the sidecar and propagating non-`IsNotExist` errors would keep the sidecar as the anchor.

pkg/cosmosboost/precompute/slim.go
```diff
@@ -0,0 +1,88 @@
...
+
+// buildSlimManifest returns a manifest document containing only the fields
+// Cosmos reads when building DAGs from a dbt manifest.json, so the file the
+// plugin loads at parse time is a fraction of the size of the original. doc
+// is not mutated.
+func buildSlimManifest(doc map[string]any) map[string]any {
+	slim := map[string]any{
```
kaxil, line 29: Nothing in the file identifies the format. The allowlist is pinned to one Cosmos version's read path, so when it changes, an older CLI's output is indistinguishable from a current one and the plugin has no way to decide it should fall back to the full manifest. The sidecar next to it carries `schema` and `generated_by`; a matching key here is a lot cheaper to add now than once slim manifests are sitting in customer bundles.

pkg/cosmosboost/precompute/precompute.go
```diff
@@ -48,6 +48,10 @@ type Summary struct {
 // and does not stop the others. Run only returns a non-nil error for a top-level
 // problem, such as a root that cannot be walked. version is recorded in each
 // sidecar's generated_by.
+//
+// Every discovered manifest.json also gets a slim, field-filtered copy (see
```
kaxil: "Every discovered manifest.json" has a gap worth calling out. `findManifests` skips a manifest whose parent is a discovered project dir, reasoning that the project's folder hash already covers it. That reasoning is hash-specific and doesn't carry over to slimming. I checked with a project holding both `manifest.json` and `target/manifest.json`: only the `target/` one gets a slim copy. Fine if intentional (the plugin just falls back to the full manifest), but it's a silent no-op for anyone pointing `ProjectConfig.manifest_path` at the project root.

pkg/cosmosboost/precompute/precompute.go
```diff
@@ -129,20 +133,30 @@ func processProject(dir, version string) Result {
...
+// along with a slim, field-filtered copy of the manifest (see
+// buildSlimManifest). A file that isn't a dbt manifest is skipped (neither is
+// written) so unrelated manifest.json files in the project aren't stamped.
 func processManifest(path, version string) Result {
 	start := time.Now()
-	hash, bytes, isDbt, err := hashManifest(path)
+	doc, bytes, isDbt, err := readManifestDoc(path)
+	var hash string
+	if err == nil && isDbt {
+		hash = hashDocument(doc)
```
kaxil, line 163: `hashDocument` mutates `doc` in place, so `buildSlimManifest(doc)` on line 158 sees the stripped document rather than what was read off disk. Benign today, since nothing in `volatileManifestKeys` or `stripEntityCreatedAt` overlaps the allowlist, but `buildSlimManifest`'s comment says "doc is not mutated", which reads as a guarantee its only caller doesn't give it. Building the slim manifest before hashing would drop the ordering dependency.

pkg/cosmosboost/precompute/cleanup.go
pankajastro: Good catch, confirmed. Fixed by removing the slim manifest before the sidecar and propagating any non-ENOENT error, so the sidecar survives a failed slim removal and stays discoverable to future runs. Added a test (`TestCleanupKeepsSidecarWhenSlimManifestUnremovable`). Pushed in 7c6143bd.

pkg/cosmosboost/precompute/slim.go
pankajastro: Verified against cosmos source — `compute_model_outlet_uris` does silently zero out dataset emission without `database`/`schema`/`alias`, confirmed. `copy_manifest_file_if_exists` copies it verbatim, though dbt's own full reparse overwrites it before executing anything, so it's less severe than it looks. Plan is to keep the plugin-side swap scoped strictly to the graph-loading read, not widen the allowlist. Added a scope note on `buildSlimManifest` so this doesn't get lost before the plugin PR. Pushed in 7c6143bd.

pkg/cosmosboost/precompute/slim.go
pankajastro: Added — the slim manifest now carries `_schema`/`_generated_by` (underscore-prefixed so they can't collide with real dbt manifest keys; confirmed against dbt-core's `Manifest` dataclass, which has neither name). Pushed in 7c6143bd.

pkg/cosmosboost/precompute/precompute.go
pankajastro: Confirmed, doc was wrong. Updated the comment to call out that a project-root manifest.json is excluded. Pushed in 7c6143bd.

pkg/cosmosboost/precompute/precompute.go
pankajastro: Fixed — slim manifest is now built before `hashDocument` mutates `doc`, so there's no ordering dependency left. Pushed in 7c6143bd.

kaxil: The reorder helps for the fields `slimResource` re-keys, but it doesn't fully decouple the two. `slim` is a shallow copy: `selectors` is literally the same map object as `doc["selectors"]`, and each resource's `config`/`tags`/`fqn`/`depends_on.nodes` values are shared as well. `writeSlimManifest` marshals `slim` after `hashDocument` has already run, so a mutation to any shared nested value still lands in the written file.

It's observable today. `stripEntityCreatedAt` walks every top-level map, `selectors` included, so a selector entry carrying `created_at` loses it in the slim output even though slim was built first. Marshaling the same `slim` map before and after `hashDocument`:

```
before: "selectors":{"my_sel":{"created_at":111,"definition":{"method":"tag"}}}
after : "selectors":{"my_sel":{"definition":{"method":"tag"}}}
```

Harmless right now (Cosmos only reads `definition` off selectors, and `stripEntityCreatedAt` stops at depth 2), but `// before hashDocument mutates doc` reads as a guarantee that only holds for the top-level and per-resource keys. Either soften the comment to say what it actually buys, or marshal slim to bytes at build time and hand `writeSlimManifest` the bytes, which makes the decoupling real.

pkg/cosmosboost/precompute/cleanup.go
kaxil: The ordering change does what I asked and closes the failed-removal route. The wider class is still open through any other way a slim manifest ends up without its sidecar, and the likeliest one is this PR's own error path: `EnsureClean` tells users to "remove them manually and retry" when it finds a foreign sidecar, and `astro dbt cleanup` from a CLI older than this PR deletes the sidecar without knowing the slim file exists. Either way the slim manifest is orphaned and no future run can see it.

I checked: stamp a project, delete `dbt_metadata.json` by hand, and `EnsureClean` returns nil with `manifest.slim.json` still sitting in `.astro/`. That's the same stale-artifact-ships outcome the anchor was meant to prevent, just reached differently.

The `_generated_by` marker you added in this push makes a direct fix cheap: `Cleanup`'s walk can match `manifest.slim.json` under `.astro/` too and apply the same ownership check it applies to the sidecar, so the file anchors itself instead of depending on its neighbour. One wrinkle on the reporting side, you'd want to emit a separate `CleanupResult` for it only when there's no sidecar beside it, or the normal case double-counts every removal.

Not blocking for merge given nothing reads the slim manifest yet, but it wants closing before the plugin-side consumer ships.

kaxil:
Re-reviewed at 7c6143bd. Three of the five round-1 items are cleanly resolved:

- The `_schema`/`_generated_by` marker is there and serializes as expected.
- The `findManifests` doc comment now calls out the project-root exclusion.
- The allowlist scope decision holds up. I re-checked against cosmos: manifest field reads appear in exactly four places, three in `graph.py` (`nodes`/`sources`/`exposures`, `metadata.project_name`, `selectors`) which the allowlist covers field for field against `_build_dbt_node_from_manifest_resource`, plus `dataset.py:312`, which is the outlet-URI reader you documented as out of scope. Nothing needed by the graph-load path is missing.

Left replies on the other two threads. The `hashDocument` ordering is only partly decoupled (slim shallow-aliases doc, and it's marshaled after the mutation), and the cleanup anchor still leaves a route to an orphaned slim manifest. Neither blocks merge.

One small thing that didn't need its own thread: `hashManifest` (`pkg/cosmosboost/precompute/hash.go:236`) has no production callers left now that `processManifest` calls `readManifestDoc` and `hashDocument` directly. Only `hash_test.go` uses it, which is why `unused` stays quiet. Worth either dropping it and pointing those tests at the two functions, or keeping it with a note on why.

I'd approve this. The two open points are producer-side invariants that matter once the plugin actually prefers `manifest.slim.json`, so they're fine to land in the consumer PR as long as they don't get lost.

pkg/cosmosboost/precompute/precompute.go
pankajastro: Fixed for real this time. `buildSlimManifest`'s result is now marshaled to bytes immediately after building it, before `hashDocument` gets a chance to mutate any value it shares with `doc` (`selectors`, and each resource's `config`/`tags`/`fqn`/`depends_on.nodes`). `writeSlimManifest` now takes the marshaled `[]byte` instead of the map, so there's no window left where a shared reference could still be mutated before it's written. Added `TestRunSlimManifestUnaffectedByHashDocumentMutation` — a selector's `created_at` fails to survive on the old code and passes now. Pushed in 1faed1e0.

pkg/cosmosboost/precompute/cleanup.go
pankajastro: Agreed, and fixed the wider class: `Cleanup`'s walk now also matches `manifest.slim.json` under `.astro/`, and when there's no sidecar beside it, checks it against its own `_generated_by` marker (the same provenance check `removeSidecar` applies) and removes or keeps it on that basis — so it's self-anchored rather than depending on a sidecar to find it. When a sidecar *is* present, its own removal still owns the slim file next to it, so normal removals aren't double-counted (added an assertion for that in `TestCleanupRemovesSlimManifestAlongsideSidecar`). Added `TestCleanupRemovesOrphanedSlimManifest` and `TestCleanupKeepsForeignOrphanedSlimManifest` for the new path. Pushed in 1faed1e0.

pankajastro:
Both open points from the last review are closed in 1faed1e0 — the slim manifest is now marshaled before `hashDocument` can mutate anything it shares with `doc`, and `Cleanup` self-anchors an orphaned `manifest.slim.json` via its own `_generated_by` marker instead of relying on a sidecar to find it. Replied inline on each thread with specifics.

kaxil (approved):

pkg/cosmosboost/precompute/slim.go
```diff
@@ -0,0 +1,109 @@
+package precompute
+
+const slimManifestName = "manifest.slim.json"
```
kaxil, line 3: There's no `.gitignore` line for this one. #2236 added `**/.astro/dbt_metadata.json` to `pkg/airflowrt/include/gitignore` when the sidecar shipped, and the slim manifest lands in the same kind of nested `.astro/` but isn't covered. I checked with the scaffold's gitignore verbatim: after a stamping run `git status` shows `?? target/.astro/manifest.slim.json` while the sidecar stays ignored. Artifacts outlive the deploy (only the next `EnsureClean` or `astro dbt cleanup` clears them), so someone commits it and the deploy after that deletes it under them. Worth adding `**/.astro/manifest.slim.json` next to the existing line.

pkg/cosmosboost/precompute/metadata.go
```diff
@@ -69,9 +69,44 @@ func writeSidecar(dir, algo, hash, version string) error {
...
+// writeArtifact writes data to dir/.astro/name, creating the directory if
+// needed. The write is atomic — a temp file in the same directory, renamed
+// over the destination — so a failed or interrupted write leaves the previous
+// contents (or nothing) rather than a truncated file. That matters most for
+// the slim manifest, which is large enough for a partial write to be real:
+// truncated JSON is unusable to the plugin AND indistinguishable from another
+// producer's file to Cleanup, which would then report it as an unrecognized
+// artifact and have EnsureClean block every later deploy over a file we wrote
+// ourselves. A hard kill between create and rename can still strand a temp
+// file in .astro, which only keeps the directory from being pruned.
```
kaxil: I'd push back on "only". `Cleanup`'s switch matches the two exact names, so a stranded `manifest.slim.json.tmp-*` is invisible to it: I stamped a project, dropped a temp file beside the artifacts, and `astro dbt cleanup` reported 2 removed / 0 kept / 0 failed with the temp file still sitting there. It ships too, since the bundle tar excludes only `.git/`. Matching the temp prefix in that switch and removing it without a provenance check (a half-written one has no marker to read) would make this self-healing rather than permanent.

pkg/cosmosboost/precompute/precompute.go
kaxil: Verified, this holds now. Ran it end to end through the exported API: a manifest carrying `selectors.my_selector.created_at`, and the field is present in the written `manifest.slim.json`. Marshaling on line 157 before `hashDocument` on 158 is what does it, and taking `[]byte` in `writeSlimManifest` removes the window entirely. `objectOrEmpty` is a good catch on top of that, cosmos does `manifest.get("selectors", {})` at graph.py:1340 so the old `null` would have come back as None. Closing this from my side.

pkg/cosmosboost/precompute/cleanup.go
kaxil: Verified. Stamped a project, deleted `dbt_metadata.json` by hand, and `Cleanup` now finds the orphan on its own `_generated_by`, removes it, and prunes `.astro`. A foreign one is kept instead, so `EnsureClean` blocks rather than shipping it. Both directions of the mixed state behave, and the per-artifact failure attribution is a nice side effect. Closing this from my side.

pkg/cosmosboost/precompute/slim.go
pankajastro: Good catch, and the failure mode you describe is the bad part — not the untracked file itself but that committing it puts the next deploy's `EnsureClean` in a position to delete it under whoever committed it.

Fixed in `dd27266e`: `**/.astro/manifest.slim.json` added next to the existing line in `pkg/airflowrt/include/gitignore`. Verified with the scaffold gitignore verbatim in a throwaway repo — `target/.astro/manifest.slim.json` is ignored alongside the sidecar, and `.astro/config.yaml` stays tracked, so the pattern isn't over-matching real project files.

pkg/cosmosboost/precompute/metadata.go
pankajastro: You're right on every point here — invisible to `Cleanup`'s two-name switch, and it ships, since the tar excludes only `.git/`. My "only" was wrong.

The concern is moot now rather than fixed, though, and not because of anything you flagged: the atomic write is gone as of `0173deea`. On review of the branch against BOSS-598 the temp-file-plus-rename was scope creep — it changed how the pre-existing hash sidecar is written to guard a partial-write scenario the ticket never raises — so it's reverted to a plain `os.WriteFile`. There's no `CreateTemp` left anywhere in `pkg/cosmosboost`, so no `.tmp-*` is ever created and there's nothing for the switch to match.

Worth saying: if we ever do want atomic writes here, your prefix-matching-without-a-provenance-check is the right shape for it — a half-written file has no marker to read — and that reasoning should go in with the change rather than be rediscovered. Sorry you verified against a version that was already on its way out.

pkg/cosmosboost/precompute/slim.go
```diff
@@ -0,0 +1,125 @@
...
+//
+// Only a substitute for the graph-loading read - Cosmos also reads
+// manifest_path directly for dbt's own subprocess copy and per-model
+// dataset/Asset outlet URIs (database/schema/alias), which this drops.
+func buildSlimManifest(doc map[string]any, version string) map[string]any {
+	slim := map[string]any{
+		"_schema":       slimSchemaVersion,
+		"_generated_by": GeneratedBy{Application: application, Version: version},
+		"metadata":      slimMetadata(doc),
+		"selectors":     objectOrEmpty(doc["selectors"]),
```
neel-astro, line 33: noob question: Do we need something like a `source hash` in this slim manifest file, so that if the model to generate the slim manifest changes it is captured over there? Or do you think schema is meant for the same purpose?

cloud/deploy/cosmosboost_hook_test.go
```diff
@@ -56,6 +56,53 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
 		"the pre-deploy step must stamp the bundle before it is tarred")
 }
 
+// TestUploadBundleWritesSlimManifestWhenEnabled pins that a discovered
+// manifest.json gets a slim, field-filtered copy alongside its hash sidecar
+// once cosmos_boost.slim_manifest is enabled on top of the step itself.
+func TestUploadBundleWritesSlimManifestWhenEnabled(t *testing.T) {
```
neel-astro: nit: These could be table driven tests to avoid the duplication

pkg/cosmosboost/precompute/cleanup_test.go
```diff
@@ -47,6 +47,237 @@ func TestCleanupRemovesSidecarAndPrunesDir(t *testing.T) {
...
+// meaningful alongside a fresh hash sidecar, so cleanup must remove a stale
+// one too - otherwise a later deploy (with a changed manifest, or with
+// cosmos_boost.pre_deploy disabled) could still ship an outdated
+// manifest.slim.json.
+func TestCleanupRemovesSlimManifestAlongsideSidecar(t *testing.T) {
+	dir := t.TempDir()
+	writeFiles(t, dir, map[string]string{
+		"manifest.json": `{"metadata":{"dbt_schema_version":"https://schemas.getdbt.com/dbt/manifest/v12.json"},"nodes":{}}`,
+	})
+	summary, err := Run([]string{dir}, "test", Options{SlimManifest: true})
```
neel-astro: can we reuse the existing `stampProject` declared above and used in other tests already

neel-astro (commented):
Left few nits and curious question, but otherwise looking good

tatiana (commented):
@pankajastro Thanks a lot for working on this. Some feedback:

1. The `version.hash` still hashes the full manifest. I believe we should have a dedicated hash for the filtered manifest, an optional section similar to:
```
     "filteredManifest": {
       "schema": 1,
       "path": "manifest.slim.json",
       "version": {
         "algo": "sha256-filtered-manifest-v1",
         "hash": "..."
       }
     }
```

2. If there is a `manifest.json` in the root of a dbt project, alongside `dbt_project.yml`, it gets ignored.

3. it seems no of the tests confirm if the `manifest.json` is unchanged - please, could you add some check on this.

tatiana (commented):
@pankajastro Please, could you replace the`separate cosmos_boost.slim_manifest` config key by an environment variable?

I believe that in Astro CLI we should switch all optimisations when `cosmos_boost.pre_deploy` is enabled, and allow individual opt-out via an environment variable instead - following the pattern of `astro-cosmos-boost-plugin` (one master switch, then individual optimisations on by default and disabled per-feature via env var).

Something like `ASTRO_COSMOS_BOOST_SLIM_MANIFEST_ENABLED` would keep both repos consistent and avoid a second flag users have to discover.

config/config.go
```diff
@@ -102,6 +102,12 @@ var (
 		ProxyPort:               newCfg("proxy.port", "6563"),
 		OttoAutoUpdate:          newCfg("otto.auto_update", "true"),
 		CosmosBoostPreDeploy:    newCfg("cosmos_boost.pre_deploy", "false"),
+		// Gates the slim, field-filtered manifest copy only; the hash sidecar
+		// is written either way. Off by default because no released Cosmos
+		// Boost plugin reads manifest.slim.json yet - until one does, writing
+		// it only costs deploy time and payload size. Flip to "true" once the
+		// plugin ships the read-with-fallback path.
+		CosmosBoostSlimManifest: newCfg("cosmos_boost.slim_manifest", "false"),
```
tatiana: As mentioned before, I don't think we want to add a specific Astro CLI config for each optimisation. We want to be consistent with the plugin - having env vars to control fine control optimisations.

cloud/deploy/bundle.go
```diff
@@ -227,7 +227,7 @@ func UploadBundle(tarDirPath, bundlePath, uploadURL string, prependBaseDir bool,
 		if err := cosmosboost.EnsureClean(bundlePath); err != nil {
 			return "", err
 		}
-		cosmosboost.BestEffortPreDeploy(bundlePath)
+		cosmosboost.BestEffortPreDeploy(bundlePath, config.CFG.CosmosBoostSlimManifest.GetBool())
```
tatiana: Once the control for disabling the slim manifest is an envvar, we should revert this change.  I don't think we want to change this method signature for each optimisation that is introduced.

`slimManifestEnabled()` can become a small helper in `pkg/cosmosboost`, reading `ASTRO_COSMOS_BOOST_SLIM_MANIFEST_ENABLED` directly with `os.Getenv`

pkg/cosmosboost/precompute/cleanup.go
```diff
@@ -38,11 +40,20 @@ type CleanupSummary struct {
...
+// file is judged by its own producer marker: one whose marker is not ours — or
+// that isn't valid JSON — is left in place and reported as kept, so cleanup
+// never deletes a file some other tool owns.
+//
+// Checking the two independently is what makes the mixed states come out
+// right, and neither is rare. A slim manifest outlives its sidecar whenever an
+// older astro-cli's cleanup deleted only the sidecar it knew about, or
+// EnsureClean left a foreign one in place; a sidecar we wrote can equally sit
+// next to a manifest.slim.json we did not. Inferring either file's provenance
+// from the other would delete a file we do not own in one direction and strand
```
tatiana: Could we express the same in fewer lines? AI tends to be very verbose.

pkg/cosmosboost/precompute/slim.go
pankajastro: `_schema` versions the allowlist, not the input, so it doesn't cover this. Added a `filtered_manifest` section to the sidecar in `45b196da` (Tatiana asked for the same thing) — it now carries both hashes: `version.hash` for the source manifest, `filtered_manifest.version.hash` for the slim file. A consumer can check either.

cloud/deploy/cosmosboost_hook_test.go
pankajastro: Done in `71d14a9a` — the two slim-manifest cases are one table now.

pkg/cosmosboost/precompute/cleanup_test.go
pankajastro: `stampProject` writes `dbt_project.yml`, which produces a project sidecar and no manifest, so it can't stand in here. Added `stampManifest` as its sibling instead — replaces five copies of the preamble. `38166e76`.

config/config.go
pankajastro: Agreed — `cosmos_boost.slim_manifest` is gone, env var instead. `71d14a9a`.

cloud/deploy/bundle.go
pankajastro: Reverted; `slimManifestEnabled()` now lives in `pkg/cosmosboost`. `71d14a9a`.

pkg/cosmosboost/precompute/cleanup.go
pankajastro: trimmed this and other code code comments a bit

pankajastro:
@tatiana All four addressed:

1. **Filtered-manifest hash** — added as a `filtered_manifest` section on the sidecar (`45b196da`): `schema`, `path`, and a `sha256-filtered-manifest-v1` hash of the slim file's bytes. Top-level `version.hash` still covers the source manifest, so a consumer can confirm the two are a matched pair without re-hashing. Snake_case to match `generated_by` in the same file — happy to flip to `filteredManifest`.
2. **Root-level `manifest.json`** — fixed in #2244, stacked on this branch. It can't be its own discovery unit (its `.astro/` is the project's), so `processProject` writes the slim copy and carries the pointer.
3. **Unchanged-manifest test** — added (`45b196da`), byte-identical after a run.
4. **Env var** — done (`71d14a9a`), config key and signature change both gone. Named `..._DISABLED` rather than `_ENABLED` so the repo's existing `util.CheckEnvBool` does the parsing — through a disable flag its truthy-only rule leaves the optimization on for unset or misspelled input, matching the plugin's fail-toward-on bias with no new parser. Happy to switch if plugin symmetry matters more.

cloud/deploy/cosmosboost_hook_test.go
```diff
@@ -56,6 +56,50 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
...
+		wantSlim bool
+	}{
+		{name: "on by default", wantSlim: true},
+		{name: "env var disables it", disable: "true", wantSlim: false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			setupCosmosBoostEnv(t)
+			require.NoError(t, config.CFG.CosmosBoostPreDeploy.SetHomeString("true"))
+			if tc.disable != "" {
+				t.Setenv("ASTRO_COSMOS_BOOST_SLIM_MANIFEST_DISABLED", tc.disable)
```
tatiana, line 76: The plugin's convention is `ASTRO_COSMOS_BOOST_SLIM_MANIFEST_ENABLED` - please change this so we are consistent.

tatiana (approved):
Thanks a lot, @pankajastro , for addressing all the feedback.
I'm happy for the PR to be merged after this feedback is addressed: https://github.com/astronomer/astro-cli/pull/2242/changes#r3774690159

cloud/deploy/cosmosboost_hook_test.go
pankajastro: Done in #2244 (`748c0c04`) — renamed to `ASTRO_COSMOS_BOOST_SLIM_MANIFEST_ENABLED`.

Still on by default under `cosmos_boost.pre_deploy`, and still parsed by this repo's existing `util.CheckEnvBool` rather than a second parser: unset means on, otherwise only a value that helper reads as true (`true`/`1`/`yes`/`y`/`on`, case-insensitive) keeps it on. One divergence from the plugin worth knowing: it keeps an *unrecognized* value on, this turns the optimization off.

pankajastro:
This got merged before I addressed it, so it's in #2244.

kaxil:
Oh that was because I had GitHub's "Enabled Auto merge" feature -- once there is an approval from the CLI maintainers

PR #2243, opened 2026-08-12
fix(deploy): don't require Astro project dir for --dags with --dags-path

jeremybeard:
## Summary
- `astro deploy --dags --dags-path <path>` no longer requires being run from an Astro project root, since DAG-only deploys sourcing DAGs via `--dags-path` don't read anything else from the working directory.
- The check still runs for plain `--dags` (DAGs read from the working directory's `dags/` folder), for `--dags-path` without `--dags` (full image+dags deploy, which needs the project root as the Docker build context), and whenever `--pytest`/`--parse` is also set (these build and run a local image from the project's Dockerfile to test-parse the DAGs).
- `--dags-path` is no longer a hidden flag.

Context: astronomer/astro-cli#2107 established the precedent of bypassing this check for `--image-name` for the same underlying reason (nothing is read from the working directory in that case).

## Test plan
- [x] `go build ./...`
- [x] `go test ./cmd/cloud/...`
- [x] Added `TestDeploySkipsEnsureProjectDirWhenDagsPathSet` covering: bypass with `--dags --dags-path`; check still enforced for `--dags` alone, `--dags-path` alone, and `--dags --dags-path` combined with `--pytest`/`--parse`
- [x] `astro deploy --help` shows `--dags-path` under DAG Flags

### DIFF
--- a/cmd/cloud/deploy.go
+++ b/cmd/cloud/deploy.go
@@ -75,7 +75,9 @@ func NewDeployCmd() *cobra.Command {
 			// A DAG-only deploy sourcing its DAGs from --dags-path doesn't read anything else
 			// from the working directory, unless --pytest/--parse is also used, which builds
 			// and runs a local image to test-parse the DAGs and so still needs the project root.
-			if cmd.Flags().Changed("dags") && cmd.Flags().Changed(dagsPathFlag) && !cmd.Flags().Changed("pytest") && !cmd.Flags().Changed("parse") {
+			// Check the bound values (not cmd.Flags().Changed) so --dags=false or an explicit
+			// empty --dags-path can't be misread as enabling the bypass.
+			if dags && dagsPath != "" && !pytest && !parse {
 				return nil
 			}
 			return EnsureProjectDir(cmd, args)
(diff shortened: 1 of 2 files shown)


neel-astro (approved):
Looks like a straight forward change aligning with the PR description

cmd/cloud/deploy.go
```diff
@@ -71,6 +72,12 @@ func NewDeployCmd() *cobra.Command {
 			if cmd.Flags().Changed(imageNameFlag) || cmd.Flags().Changed(nonDagsFlag) {
 				return nil
 			}
+			// A DAG-only deploy sourcing its DAGs from --dags-path doesn't read anything else
+			// from the working directory, unless --pytest/--parse is also used, which builds
+			// and runs a local image to test-parse the DAGs and so still needs the project root.
+			if cmd.Flags().Changed("dags") && cmd.Flags().Changed(dagsPathFlag) && !cmd.Flags().Changed("pytest") && !cmd.Flags().Changed("parse") {
```
neel-astro: this would evaluate to true if someone passes `--dags=false` flag explicitly

neel-astro (approved):

PR #2244, opened 2026-08-13
feat: Create a slim dbt `manifest.json` on project root

pankajastro:
Two follow-ups to review feedback on #2242, which merged before they were addressed.

Linear: [BOSS-598](https://linear.app/astronomer/issue/BOSS-598/boost-plugin-astro-cli-generate-a-slim-field-filtered-dbt-manifest-at)

## 1. Slim a `manifest.json` in a dbt project's root

A `manifest.json` beside `dbt_project.yml` got no slim copy, so the projects most likely to keep one there gained nothing from BOSS-598.

It can't be a discovery unit of its own: its `.astro/` is the project's own, so a manifest sidecar there would collide with the project sidecar, which is why `findManifests` skips it. `processProject` handles it instead, writing the slim copy and carrying the `filtered_manifest` pointer in the sidecar it already writes.

- The project hash can't move: `hashProject` skips `.astro` at any depth, pinned by a test since a moved hash would invalidate every cache key once.
- The root manifest is read twice per deploy: streamed for the folder hash, parsed for slimming.
- Project sidecars can now carry `filtered_manifest`, so the plugin sees it from both sidecar-resolution paths.

## 2. Rename the env var to `..._SLIM_MANIFEST_ENABLED`

Was `..._SLIM_MANIFEST_DISABLED`; renamed to match the plugin's convention. Still on by default under `cosmos_boost.pre_deploy`, and still parsed by `util.CheckEnvBool`: unset means on, otherwise only a value that helper reads as true keeps it on.

## Testing

`go build`, `go vet`, `gofumpt`, `golangci-lint` (v2.11.4) and `go test ./...` all clean in a `golang:1.26` container, except the pre-existing `airflow`/`cmd`/`pkg/git` failures that also fail on `main`.

Also ran end to end on a 200-node / 567 KiB fixture: both a root and a `target/` manifest slimmed to 61 KiB, each sidecar carrying its own algo plus the pointer, `manifest.json` byte-identical, and nothing written with the env var off.

### DIFF
--- a/cloud/deploy/cosmosboost_hook_test.go
+++ b/cloud/deploy/cosmosboost_hook_test.go
@@ -57,23 +57,23 @@ func TestUploadBundleRunsPreDeployWhenEnabled(t *testing.T) {
 }
 
 // TestUploadBundleSlimManifest pins the slim manifest's gating end to end: on
-// by default once the step is enabled, and off only when the per-feature env
-// var disables it. Either way the hash sidecar still lands - the two
-// optimizations are independent.
+// by default once the step is enabled, and off when the per-feature env var
+// says so. Either way the hash sidecar still lands - the two optimizations are
+// independent.
 func TestUploadBundleSlimManifest(t *testing.T) {
 	for _, tc := range []struct {
 		name     string
-		disable  string // value for the disable env var; "" leaves it unset
+		envVar   string // value for the per-feature env var; "" leaves it unset
 		wantSlim bool
 	}{
 		{name: "on by default", wantSlim: true},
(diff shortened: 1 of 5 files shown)


tatiana (commented):
@pankajastro since @kaxil merged before https://github.com/astronomer/astro-cli/pull/2242#pullrequestreview-4926198171 was addressed, please, address it in this PR

pankajastro:
Already pushed this in `748c0c04`, just after your review.

tatiana (approved):
Thanks for addressing the feedback, @pankajastro !  Looks great!

PR #2255, opened 2026-09-09
Bump local Postgres to 15 and pin existing projects to their version

jeremybeard:
## Description

`astro dev start` has run Postgres 12.6 since the CLI V1 convergence in 2022. Postgres 12 reached end of life in November 2024, and Airflow 3.2's migrations call `gen_random_uuid()`, which is only built in from Postgres 13 — so pointing an existing local project at a 3.2 runtime fails during `db migrate` with `function gen_random_uuid() does not exist`. New projects now start on Postgres 15, which is the oldest major any Astro data plane provisions.

Raising the default alone would break every existing local project, because a Postgres server refuses to read a data directory written by an older major version. So the version a project runs is derived from its own data volume rather than from config: a project with an existing data directory keeps the major that wrote it, and only a project without one follows `postgres.tag`. Moving an existing project across is `astro dev kill` then `astro dev start`, which discards the local database along with the volume.

Defaults are also no longer handed to viper. viper serializes its defaults into the config file it creates, so the default shipped on the day a file appeared was frozen onto disk, and no later change to any default ever reached that user — `~/.astro/config.yaml` has carried `postgres.tag: "12.6"` on every install since 2022. Defaults are resolved in the `cfg` getters instead: an unset value follows the code default, and a value present in the file is the user's own choice.

A config file written before that change is migrated once, on the next startup. Entries still matching a default the CLI has shipped are removed and the file is stamped `config_version: 1`, so it is never scanned again. Removing them changes nothing about what the CLI does that day — the same value comes back from the default — but it lets those settings follow the default again, for this bump and every future one. Only keys with a non-empty default are considered, which is exactly the set viper could have written; contexts and tokens are never touched. A value the user chose is kept, and after the stamp anything they set stays put, `12.6` included.

**Files to review (8, +827 / -67):**

| File | Why |
|---|---|
| `airflow/postgres_version.go` *(start here)* | New. Finds the project's data volume, reads `PG_VERSION` from it, decides the tag. |
| `airflow/docker.go` | Calls the above in `Start` and `ComposeExport`; `PortOverrides` renamed. |
| `airflow/container.go` | `ComposeOverrides` gains `PostgresTag`, applied when rendering the compose template. |
| `config/types.go` | Getters resolve defaults, including the new bool/int fallbacks. |
| `config/config.go` | The `SetDefault` loop is gone, replaced by the one-time file migration; `postgres.tag` default is 15. |

```diff
  astro dev start                            cmd/airflow.go
  └─ DockerCompose.Start()                   airflow/docker.go
     ├─ imageHandler.Build()                 # unchanged
+    ├─ resolvePostgresTag()                 airflow/postgres_version.go
+    │  ├─ findPostgresDataVolume()          # by compose labels, then by name
+    │  └─ readVolumePGVersion()             # create container, copy out, remove
+    │     ├─ removeStaleProbes()            # clear strays from an interrupted run
+    │     └─ probeImage()                   # project's own image; postgres image otherwise
     ├─ createDockerProjectWithOverrides()   # was proxy-mode only, now every generated file
     └─ composeService.Up()                  # unchanged
```

Unchanged: every other `astro dev` subcommand, the compose template itself, and the containers a project runs — only the postgres image tag can differ.

**Reviewer notes**

- **The probe container is created but never started.** Docker populates volume mounts at create time, so `PG_VERSION` can be copied straight back out without paying for a container start. Verified against Docker (moby 29.4) and Podman's docker-compat API.
- **Every failure to read the volume falls through to the configured tag** rather than failing the start. A transient Docker error therefore reintroduces the stranding this guards against, which is the deliberate trade: a version check should not be able to stop a project starting.
- **A custom `postgres.repository` disables the check.** Pinning means asking for `<repository>:<major>`, which only the official image is known to publish — `postgis/postgis:12` does not exist. Users pointing elsewhere own their own version management.
- **`postgres.tag: latest` is not covered.** The tag cannot be parsed to a major, so the check is skipped and a `latest` server can still land on an older data directory. Same as today's behaviour; worth a follow-up.
- **The migration is the only thing that can reach existing installs.** Without it the bump is inert for anyone who has run the CLI before, because their config already pins `12.6` and that is indistinguishable from a deliberate choice. The cost is that someone who genuinely chose a value equal to a shipped default loses that pin, once.
- **Focus area:** `findPostgresDataVolume`. It has to derive the same project name compose did — `normalizeName`, not just lowercasing — or it silently finds nothing and the project is stranded rather than pinned.

## 🎟 Issue(s)

None.

## 🧪 Functional Testing

Run against Docker (OrbStack, moby 29.4.0, arm64), Airflow 3 runtime 3.3-7 and Airflow 2 runtime 12.9.0.

| Case | Result |
|---|---|
| Fresh install, new project | `postgres:15.19`, all components healthy, `example_astronauts` ran to success |
| Existing pg12 project, `postgres.tag` 15 | starts on `postgres:12`, no extra output, variables and connections intact |
| Same, then `dev kill` + `dev start` | volume dropped, `postgres:15.19`, empty database |
| Volume newer than config (pg17, tag 15) | stays on `postgres:17` |
| Airflow 2 runtime, pg12 volume, tag 15 | `postgres:12`, healthy |
| `dev restart` on a pinned project | stays on `postgres:12`, data preserved |
| `dev object export --compose` | exports `postgres:12`, matching what the project runs |
| Start from that exported file via `--compose-file` | `postgres:12`, healthy |
| `postgres.tag: 14` on a new project | `postgres:14` |
| `postgres.repository: postgis/postgis`, tag `15-3.4` | honoured, never overridden |
| `duplicate_volumes: false` | check skipped, starts on the configured tag |
| Legacy install: existing project after upgrade | config migrated, project stays on `postgres:12`, its data intact |
| Legacy install: new project after upgrade | `postgres:15`, healthy |
| Legacy config carrying contexts and tokens | migrated, credentials preserved byte for byte |
| Deliberate `postgres.tag: 13.4` and `page_size: 50` | both survive the migration |
| `astro config set -g postgres.tag 12.6` after migrating | sticks across invocations |
| Read-only `~/.astro` | nothing printed, shell completion unaffected |

<details>
<summary>Cost, cleanup and the config file</summary>

A warm start on a project with an existing volume takes 5.8s end to end; the probe itself is a create/copy/remove round trip of roughly 0.3s, and a project with no volume pays two volume lookups (~40ms) and no container at all. No probe containers were left behind across the run.

A fresh config file is written stamped and otherwise empty, and `astro config set -g page_size 30` writes only `page_size` — no defaults are materialised. Migrating a legacy file is idempotent: a second run leaves it byte-identical.

A future default change needs no new schema version. The migration removes the frozen entry, so the setting follows the code default from then on, and `supersededDefaults` never grows — it records only what the old viper behaviour could have written, and that behaviour ends at version 1.

</details>

`make test` passes. `make lint` needs `prek`, which was not available; `golangci-lint` was run directly against the repo's `.golangci.yml` instead and reports the same 50 pre-existing `goconst` issues as `main`, none in the new code. Cross-compiles for windows/amd64, windows/arm64, linux/amd64 and linux/arm64.

## 📸 Screenshots

None — no visual output changes.

## 📋 Checklist


airflow/postgres_version.go
```diff
@@ -0,0 +1,275 @@
...
+		logger.Debugf("unreadable %s in volume %s: %s", pgVersionFile, volName, err)
+		return 0
+	}
+	return major
+}
+
+// removeStaleProbes deletes probe containers left behind by an earlier run. They are
+// found by label: naming them instead would mean a leftover collided with the next
+// create, which would silently switch the version check off.
+func (d *DockerCompose) removeStaleProbes(ctx context.Context) {
```
jlaneve, line 205: this only runs on start, so a leftover probe container breaks `dev kill`: the volume stays, compose says "still in use", exit code is 0. next start cleans up the probe container and pins to 12 again. can we call this from `Kill` too, before the volume is removed? I hit this by leaving a labeled container on the volume and running `dev kill`

airflow/postgres_version.go
```diff
@@ -0,0 +1,275 @@
...
+
+	rc, _, err := d.cliClient.CopyFromContainer(ctx, created.ID, pgDataDir+"/"+pgVersionFile)
+	if err != nil {
+		// Either way the configured tag is used, so a docker failure here can still
+		// put a new postgres on an old data directory. The two are logged apart
+		// because only the second is worth investigating.
+		if errdefs.IsNotFound(err) {
+			logger.Debugf("no %s in volume %s", pgVersionFile, volName)
+		} else {
+			logger.Debugf("could not read %s from volume %s: %s", pgVersionFile, volName, err)
```
jlaneve: when this fails, the user only sees the health check time out. the real reason is debug-only. can we warn here at normal verbosity? the not-found case can stay quiet

airflow/postgres_version.go
```diff
@@ -0,0 +1,275 @@
...
+	// Pinning means asking for <repository>:<major>, which only the official postgres
+	// image is guaranteed to publish. A custom repository is the user's to version.
+	if repo := config.CFG.PostgresRepository.GetString(); repo != config.PostgresRepositoryDefault {
+		logger.Debugf("skipping postgres version check: postgres.repository is %s", repo)
+		return "", nil
+	}
+
+	configuredMajor, err := majorVersion(config.CFG.PostgresTag.GetString())
+	if err != nil {
+		// An unparseable tag is the user's own pin — a digest, "latest", an internal
```
jlaneve, line 70: `postgres.tag: latest` now puts postgres 18 on a 15 volume and the db crash-loops. same as before, but the bump makes it show up. worth an issue

config/config.go
```diff
@@ -159,6 +168,80 @@ func initHome(fs afero.Fs) {
...
+			continue
+		}
+		current := viperHome.GetString(cfg.Path)
+		if current == cfg.Default || slices.Contains(supersededDefaults[cfg.Path], current) {
+			deleteNested(settings, cfg.Path)
+		}
+	}
+	settings[configVersionKey] = currentConfigVersion
+
+	stamped := viper.New()
```
jlaneve, line 211: nit: this drops any comments in the yaml file. fine for a one-time pass, just noting

jlaneve (approved):
tested against docker: old pg12 project, project name with a dot and a space, unlabeled volume, bad PG_VERSION, leftover probe container, and a dozen hand-made config files. pinning, migration and kill/start all work. one small ask on `dev kill`, rest is non-blocking

PR #2256, opened 2026-09-10
chore: bump golangci-lint to v2.13.2 for Go 1.27 support

jeremybeard:
## Description

`make lint` fails on a clean tree for anyone running Go 1.27, because golangci-lint v2.11.4 cannot read Go 1.27 stdlib export data — it reports typecheck errors on imports of `fmt`, `os`, `slices` and `strings` (`export data version 4 is greater than maximum supported version 2`). v2.13.0 added go1.27 support, so this bumps the pin in `prek.toml` to v2.13.2. v2.12.2 is not a way out: it panics outright on Go 1.27.

CI is unaffected today — it builds with `go-version-file: go.mod` (1.26.1) — so this is about unblocking local runs, and about not being stuck on 1.26 when CI's Go version next moves.

The bump surfaces 54 findings v2.11.4 did not report. None were real defects; 50 came from one new goconst behaviour.

| Finding | Count | Resolution |
|---|---|---|
| `goconst` on cobra `Use`/`Short` fields | 50 | `exclude-types` in `.golangci.yml` |
| `govet` `inline` analyzer | 1 | exclusion rule |
| `staticcheck` SA1019 `http.Transport.Dial` | 1 | fixed — now `DialContext` |
| `staticcheck` SA9010 in `airflow/standalone.go` | 1 | false positive — `//nolint` |
| `unparam` unused test closure param | 1 | fixed — renamed to `_` |

**Reviewer notes:**

- **The goconst 50 are one upstream change, not 50 problems.** The bundled goconst gained a `CompositeLit` context type, so string literals inside composite literals are now counted — which means every `&cobra.Command{Use: "list", ...}` in `cmd/cloud/`. `exclude-types` restores the previous behaviour repo-wide rather than carving out a path. `Call` is listed alongside `CompositeLit` because it is goconst's own default, and naming the key replaces the default rather than adding to it.
- **SA9010 is wrong here.** `settings.SetExecAirflowCommand` (`settings/settings.go:775`) swaps the command and returns the previous one, so `defer settings.SetExecAirflowCommand(origExec)` *is* the restore. SA9010 targets the `defer trace.Start()` shape, where the returned function is the cleanup.
- **Focus area:** `houston/houston.go` is the only behavioural change. `Dial` and `DialContext` on `net.Dialer` have the same timeout semantics and `DialContext` takes priority when both are set, so this should be inert — worth a second look if Houston request cancellation matters to you.

## 🎟 Issue(s)

None.

## 🧪 Functional Testing

1. On a machine with Go 1.27.x installed, `make lint` on `main` fails with four typecheck errors in `scripts/patch-v1-gen/main.go`.
2. On this branch, `make lint` passes.
3. `make fmt` produces no diff.

```
$ make lint          # main, Go 1.27.1
  4 issues:
  * typecheck: 4
make: *** [lint] Error 1

$ make lint          # this branch, Go 1.27.1
golangci-lint............................................................Passed
```

Verified against Go 1.27.1 (darwin/arm64). `HOUSTON_HOST=houston make test` passes with no failures. Not verified on Go 1.26.1, which is what CI uses — the linter is backward compatible with older Go, but CI on this PR is the real check for that.

## 📸 Screenshots

N/A — terminal output above.

## 📋 Checklist


jlaneve (approved):

Simpcyclassy (approved):
On the Houston update Dial→DialContext swap in houston.go looks fine. I tested a simple authentication path with this branch and LGTM!

